PR #33, opened 2020-05-20
Reduce number peaks

jspaaks:
Originally here https://github.com/matchms/matchms-backup/pull/209


matchms/filtering/reduce_to_number_of_peaks.py
```diff
@@ -0,0 +1,53 @@
...
+    n_required: int
+        Number of minimum required peaks. Spectra with fewer peaks will be set
+        to 'None'. Default is 1.
+    n_max: int
+        Maximum number of peaks. Remove peaks if more peaks are found.
+    ratio_desired: float, optional
+        Set desired ratio between maximum number of peaks and parent mass.
+        Default is None.
+    """
+    def _set_maximum_number_of_peaks_to_keep(spectrum, ratio_desired):
```
jspaaks: the pythonic way is to not have these arguments, just have 
```python
def _set_maximum_number_of_peaks_to_keep():
```

matchms/filtering/reduce_to_number_of_peaks.py
```diff
@@ -0,0 +1,53 @@
...
+        parent_mass = spectrum.get("parent_mass", None)
+        if parent_mass and ratio_desired:
+            n_desired_by_mass = int(ceil(ratio_desired * parent_mass))
+            threshold = max(n_required, n_desired_by_mass)
```
jspaaks: ``threshold`` shadows a name from outer scope, can be avoided by just ``return``ing without assignment first

matchms/filtering/reduce_to_number_of_peaks.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            threshold = n_max
+        return threshold
+
+    def _remove_lowest_intensity_peaks(spectrum):
```
jspaaks: the pythonic way is to not have this argument, just have
```python
def _remove_lowest_intensity_peaks():
```

matchms/filtering/reduce_to_number_of_peaks.py
```diff
@@ -0,0 +1,53 @@
...
+        mz, intensities = spectrum.peaks
+        idx = intensities.argsort()[-threshold:]
+        idx_sort_by_mz = mz[idx].argsort()
+        spectrum.peaks = Spikes(mz=mz[idx][idx_sort_by_mz],
```
jspaaks, line 34: I think we should look into adding ``getitem`` and ``setitem`` to ``Spikes``, so we can avoid these ugly indices (but let's do that later).

matchms/filtering/reduce_to_number_of_peaks.py
florian-huber: Done.

matchms/filtering/reduce_to_number_of_peaks.py
florian-huber: Done.

matchms/filtering/reduce_to_number_of_peaks.py
florian-huber: Agreed, that looks better.

matchms/filtering/reduce_to_number_of_peaks.py
florian-huber: Yes, a setter for Spikes would be nice!
(to be done later)

florian-huber:
Thanks for the comments @jspaaks .  I changed the code accordingly. Let me know if that looks OK for you then I can merge it.

matchms/filtering/reduce_to_number_of_peaks.py
jspaaks: https://github.com/matchms/matchms/issues/43

jspaaks:
@florian-huber, this is good to merge. GitHub won't let me approve at the moment because I transferred this issue from matchms-backup repo, which made me the author.

florian-huber:
Thanks Juriaan. I'll merge it.

PR #34, opened 2020-05-20
229 clean compound name

### LINKED ISSUE #23: Handle different compound name input types and fields
When handling both data from GNPS json files (see matchms/matchms-backup#207), a number of new issues arise.
One of the is that GNPS json files do not have ``name`` field, but already provide a (poorly) cleaned ``compound_name`` and ``adduct`` instead. 

The current ``add_adduct`` function does not work on the new format. And later lookup operations rely on one specific field to retrieve the compound name (preferably in a cleaned version).

jspaaks:
Originally here https://github.com/matchms/matchms-backup/pull/230

Addressing #23 .
It grew a bit bigger, because cleaning up the compound names turned out to be a very important for subsequent lookup-based  metadata cleaning.

- Changed ``add_adducts`` to ``derive_adduct_from_name``. It now checks potential adducts based on regular expression and removes it from name if found.
- Added analogous ``derive_formula_from_name`` filter.
- Added ``clean_compound_name`` function to remove known parts that do not belong to the compound name.
- Added tests.
- Added below order of filters to ``default_filters``.

The workflow for me would then look like (though I also added it to ``default_filters``):
```python
def name_cleaning(s):
    s = add_compound_name(s)
    s = derive_adducts_from_name(s)
    s = derive_formula_from_name(s)
    s = clean_compound_name(s)
    return s

spectrums_cleaned_name = [name_cleaning(s) for s in spectrums]
```


matchms/filtering/add_compound_name.py
```diff
@@ -0,0 +1,23 @@
...
+
+def add_compound_name(spectrum_in: SpectrumType) -> SpectrumType:
+    """Add compound_name to correct field: "compound_name" in metadata."""
+    if spectrum_in is None:
+        return None
+
+    spectrum = spectrum_in.clone()
+
+    if spectrum.get("compound_name", None) is None:
+        # No compound_name entry found
```
cwmeijer: To me this comment is redundant as the code is already very clear. Removing the comment would make it cleaner in my opinion. But you can leave it if you like. Just a small thing.

matchms/filtering/clean_compound_name.py
```diff
@@ -0,0 +1,63 @@
...
+def clean_compound_name(spectrum_in: SpectrumType) -> SpectrumType:
+    """Clean compound name.
+
+    A list of frequently seen name additions that do not belong to the compound
+    name will be removed."""
+
+    def remove_non_compound_name_parts(name):
+        """Clean "name string by removing known parts that don't belong there."""
+        name = name.strip()
+        # remove type NCGC00180417-03_C31H40O16_
```
cwmeijer, line 14: good use of comments here :-)

matchms/filtering/clean_compound_name.py
```diff
@@ -0,0 +1,63 @@
...
+        name = re.split(r"ReSpect:[A-Z]{2,}[0-9]{6}.*\|", name)[-1]
+        # ReSpect:PS013405 option1
+        name = re.split(r"[A-Z]{2,}[0-9]{6}( )", name)[-1]
+        # remove type 0072_2-Mercaptobenzothiaz
+        name = re.split(r"^[0-9]{4}_", name)[-1]
+        # remove type nameofcompound_CID20_170920 or Spiraeoside_HCD30_170919
+        name = re.split(r"_((HCD)|(CID))[0-9]{2}_[0-9]{5,6}$", name)[0]
+
+        # Remove further non compound-name parts
+        parts_remove = ["Spectral Match to",
```
cwmeijer, line 38: If you like you could group these lines (36-40) in a function and call it 'remove_futher_non_compount_parts' for example. I think it would be nice, although it is a bit of a fuzzy name. If you can't think of a better name that describes what you do in these lines, then maybe leave it like this, or remove the blank line above because it implies that these lines are somehow doing something conceptually different than the ones before line 36.

matchms/filtering/clean_compound_name.py
```diff
@@ -0,0 +1,63 @@
...
+        for part in parts_remove:
+            name = name.replace(part, "")
+        return name.strip()
+
+    if spectrum_in is None:
+        return None
+
+    spectrum = spectrum_in.clone()
+
+    # Get compound name
```
cwmeijer: redundant and confusing comment

matchms/filtering/clean_compound_name.py
```diff
@@ -0,0 +1,63 @@
...
+    if spectrum.get("compound_name", None) is not None:
+        name = spectrum.get("compound_name")
+    else:
+        assert spectrum.get("name", None) in [None, ""], ("Found 'name' but not 'compound_name' in metadata",
+                                                          "Apply 'add_compound_name' filter first.")
+        return spectrum
+
+    # Clean compound name
+    name_cleaned = remove_non_compound_name_parts(name)
+    name_cleaned = name_cleaned.strip("; ")
```
cwmeijer: should this line not be part of the cleaning?

matchms/filtering/derive_adduct_from_name.py
```diff
@@ -0,0 +1,48 @@
...
+        Input spectrum.
+    remove_adduct_from_name: bool
+        Remove found adducts from compound name if set to True. Default is True.
+    """
+    if spectrum_in is None:
+        return None
+
+    spectrum = spectrum_in.clone()
+
+    # Get compound name
```
cwmeijer: You have duplicated this piece of code quite a lot. Could you extract it as a function and place it in some utils file or find some other way to get rid of the duplication?

matchms/filtering/derive_adduct_from_name.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Detect adduct in compound name
+    adduct_from_name = None
+    name_split = name.split(" ")
+    for name_part in name_split[::-1][:2]:
+        if looks_like_adduct(name_part):
+            adduct_from_name = name_part
+            break
+
+    # Remove found adduct from compound name (if remove_adduct_from_name=True)
```
cwmeijer: The code below is just as clear as the comment. I'd say the comment is unnecessary. If you feel these line need a 'header' or introduction somehow, I'd suggest making it a function (with all its benefits like clear scope and reusability and what not)

matchms/filtering/derive_formula_from_name.py
```diff
@@ -0,0 +1,51 @@
...
+        Input spectrum.
+    remove_formula_from_name: bool
+        Remove found formula from compound name if set to True. Default is True.
+    """
+    if spectrum_in is None:
+        return None
+
+    spectrum = spectrum_in.clone()
+
+    # Get compound name
```
cwmeijer: Confusing comment. Removing it increases readibility to me. And code duplication.

matchms/filtering/derive_formula_from_name.py
```diff
@@ -0,0 +1,51 @@
...
+    if spectrum.get("compound_name", None) is not None:
+        name = spectrum.get("compound_name")
+    else:
+        assert spectrum.get("name", None) in [None, ""], ("Found 'name' but not 'compound_name' in metadata",
+                                                          "Apply 'add_compound_name' filter first.")
+        return spectrum
+
+    # Detect formula at end of compound name
```
cwmeijer, line 28: Consider the shorter:
end_of_name = name.split(" ")[-1]
formula_from_name = end_of_name if looks_like_formula(end_of_name) else None

cwmeijer (changes requested):
Hey Florian,
I had a look at the PR. I didn't have time to completely finish it right now but saw most of it except for tests, The tests also look fine, but I lack the domain knowledge to look into them anyway.

Great way of cleaning names and looks systematic. There is some code duplication though. It would be a big improvement to find someway to get rid of it, in terms of quality, maintenance, readibility etc.
I also saw quite some comments that were really just stating the next line, which was already clear on its own. I also saw some long functions with options to split into functions that would probably get rid of any need to have comments at all. 

Good job so far :-)

matchms/filtering/add_compound_name.py
florian-huber: Done.

matchms/filtering/clean_compound_name.py
florian-huber: thanks :)

matchms/filtering/clean_compound_name.py
florian-huber: I removed the comment.

matchms/filtering/clean_compound_name.py
florian-huber: True. I moved it to the cleaning function.

matchms/filtering/derive_adduct_from_name.py
florian-huber: Comment was removed.

matchms/filtering/derive_formula_from_name.py
florian-huber: Removed the comment.

matchms/filtering/clean_compound_name.py
florian-huber: I broke it down into two functions as you suggested. 
``remove_parts_by_regular_expression`` and ``remove_known_non_compound_parts``

matchms/filtering/derive_adduct_from_name.py
florian-huber: It still felt a bit too "small" to me to add it to utils. I removed the unnecessary comments and kept it unchanged otherwise. Would it be OK for you to keep it at that for now?

matchms/filtering/derive_formula_from_name.py
florian-huber: Yes, thanks. I changed it to your suggestion.

florian-huber:
@sverhoeven I am in doubt how to fix this, so it would be great to have your opinion first.

Right now, the tests fail because of the readthedocs ``index.rst`` example does give a different output. That is because I changed the ``default_filters`` with this PR and it now contains a more more extensive compound-name cleaning. And (which is what makes the test fail): it prints all changes it does to the name...

I would see several options to fix this:
1. Suppress the print statements in ``derive_adduct_from_name``, ``derive_formula_from_name``. (maybe make it optional in default_filters by adding a flag?)
2. Change the index example, maybe by breaking it into two parts: i) importing & filtering and ii) calculating cosine scores and include the output from the name cleaning steps.
3. Use pre-cleaned data.

Any preference, or better alternative?

sverhoeven:
> @sverhoeven I am in doubt how to fix this, so it would be great to have your opinion first.
> 
> Right now, the tests fail because of the readthedocs `index.rst` example does give a different output. That is because I changed the `default_filters` with this PR and it now contains a more more extensive compound-name cleaning. And (which is what makes the test fail): it prints all changes it does to the name...
> 
> I would see several options to fix this:
> 
> 1. Suppress the print statements in `derive_adduct_from_name`, `derive_formula_from_name`. (maybe make it optional in default_filters by adding a flag?)
> 2. Change the index example, maybe by breaking it into two parts: i) importing & filtering and ii) calculating cosine scores and include the output from the name cleaning steps.
> 3. Use pre-cleaned data.
> 
> Any preference, or better alternative?

4. Use logging library instead of print

    ```python
    import logging
    logger = logging.getLogger(__name__)
    ...
    logger.info(msg)

    # in user code
    import logging
    logging.basicConfig(level=logging.INFO)
    ```

5. Adjust test and use `...` to ignore most of the filter prints

        ```
        Removed adduct M-H from compound name.
        Added adduct M-H to metadata.
        ...
        Reference scan id: 675
        Query scan id: 2833
        Score: 0.0293
        Number of matching peaks: 20
        ----------------------------
        Reference scan id: 1320
        Query scan id: 2833
        Score: 0.0137
        Number of matching peaks: 24
        ...
        ```

I would go for logging (option 4) see table at https://docs.python.org/3/howto/logging.html#when-to-use-logging, that suggest for these kinds of prints we should use info or debug. Also see https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

florian-huber:
@sverhoeven I would agree that logging the made metadata changes as "info" seems better! For now, however, I decided to leave out the logging and only changed the ``index.rst``. 
I made an issue (#71) to add the logging in another PR, mostly because I feel it should be implemented in other filter functions as well. And it would then also be nice to have a flag for saving the calls to a file.

sverhoeven (approved):

PR #35, opened 2020-05-20
Add correct charge test

jspaaks:
Originally here https://github.com/matchms/matchms-backup/pull/232

- added unit tests for ``correct_charge``
- added asserts to ``correct_charge`` to make sure users execute filters in the right order.


florian-huber:
This one is ready to go and not to big.. Anyone up for reviewing it?

tests/test_correct_charge.py
```diff
@@ -0,0 +1,56 @@
...
+    spectrum_in = Spectrum(mz=numpy.array([], dtype='float'),
+                           intensities=numpy.array([], dtype='float'),
+                           metadata={"ionmode": "positive"})
+
+    spectrum = correct_charge(spectrum_in)
+
+    assert spectrum.get("charge") == 1, "Expected different charge value."
+
+
+def test_correct_charge_correct_charge_sign_plus_to_min():
```
fdiblen:
```suggestion
def test_correct_charge_sign_plus_to_min():
```

tests/test_correct_charge.py
```diff
@@ -0,0 +1,56 @@
...
+                           intensities=numpy.array([], dtype='float'),
+                           metadata={"ionmode": "negative",
+                                     "charge": 2})
+
+    spectrum = correct_charge(spectrum_in)
+
+    assert spectrum.get("charge") == -2, "Expected different charge value."
+
+
+def test_correct_charge_correct_charge_sign_min_to_plus():
```
fdiblen:
```suggestion
def test_correct_charge_sign_min_to_plus():
```

fdiblen (changes requested):
suggested very minor changes

florian-huber:
Thanks for the review @fdiblen ! Let me know if I can merge it.

matchms/filtering/correct_charge.py
```diff
@@ -14,17 +15,21 @@ def correct_charge(spectrum_in: SpectrumType) -> SpectrumType:
     spectrum = spectrum_in.clone()
 
     ionmode = spectrum.get("ionmode", None)
+    if ionmode:
+        assert ionmode == ionmode.lower(), ("Ionmode field not harmonized.",
+                                            "Apply 'make_ionmode_lowercase' filter first.")
 
     charge = spectrum.get("charge", None)
+    assert not isinstance(charge, str), ("Charge is given as string.",
```
fdiblen, line 23: would be nicer to use numpy.isscalar here

https://numpy.org/doc/stable/reference/generated/numpy.isscalar.html

florian-huber: Not sure. I wanted to specifically return a message if a string is given, which happens when importing spectra from GNPS json files without running the ``make_charge_scalar`` filter first.

fdiblen: To me this is easier to read and understand.
`assert numpy.isscalar(charge)`

florian-huber: ``numpy.iscalar`` won't work. It only checks if it is a scalar and not an array. 
``numpy.iscalar("text")`` will also return ``True``

florian-huber: @fdiblen are you fine with leaving it as it is and merging it?

cwmeijer: Numpy isscalar indeed does not seem to do what its name implies. It is not suited here, as Florian explained and can be read here (https://numpy.org/devdocs/reference/generated/numpy.isscalar.html).
The documentation states:
'If you need a stricter way to identify a numerical scalar, use isinstance(x, numbers.Number), '
So that would be another option. 
In this case though, you want to help users who make a common mistake of inputting a string specifically. I think it makes sense to keep it like this.

florian-huber: Thanks @cwmeijer, to me that resolves it. I would go ahead and merge this soon then.

PR #36, opened 2020-05-20
226 fix loss intensity order

### LINKED ISSUE #20: Intensities of losses are in wrong order
The ``add losses`` filter creates losses by subtracting peak mz from the precursor-m/z. 
When using it on actual data I ran into two issues:
- loss intensities are in the wrong order (inverse to the loss mz)
- people using losses restrict them to a much smaller range than mz values, so the function should have a ``from_mz`` and ``to_mz`` parameters.
There also should be a default ``from_mz=0``, because some spectra have peaks with m/z larger than the precursor-m/z.

jspaaks:
Originally here https://github.com/matchms/matchms-backup/pull/233

-   fix intensity order in ``add_losses``
-   add unit tests
-   add ``loss_mz_from`` and ``loss_mz_to`` parameters to ``add_losses``


florian-huber:
Got too far behind, I replaced it by PR #48

PR #44, opened 2020-05-25
smaller logo on readme

cwmeijer:
Also removed matchms header, as the logo IS the header. 
Because image height is ignored by github while rendering rst (https://github.com/github/markup/issues/295 from 2014), I'm using raw html to get the logo the size that I want.


fdiblen (approved):
We (@jspaaks and I) liked it.
:100:

sverhoeven (approved):
Looks nice but I am a bit sad that HTML was needed.

Issue #22, opened 2020-05-14
run unit tests and integration test when building the conda package


fdiblen:
- [x] run the test before building the conda package 
- [x] remove the tests from the package tarball

PR #47, opened 2020-05-26
22 remove test folder

fdiblen:
### **Patch release: 0.3.3**

Refs: #22 

### Added
- Build workflow runs the tests after installing the package

### Changed
- tests were removed from the package (see setup.py)


.github/workflows/conda_build.yml
```diff
@@ -201,3 +201,18 @@ jobs:
...
+          conda list
+      - name: Run tests after installing the package
+        shell: bash -l {0}
+        run: |
+          export TESTDIR=$RUNNER_TEMP/matchms/_test
+          [ "$RUNNING_OS" = "Windows" ] && export TESTDIR=$RUNNER_TEMP\\matchms\\_test\\
+          mkdir $TESTDIR
+          cp -iRv tests $TESTDIR
+          cd $TESTDIR/tests
+          pytest --ignore-glob="test_version_string_consistency.py"
```
sverhoeven: Integration test should also be run and don't use glob when you can be exact.

```suggestion
          cp -iRv tests integration-tests $TESTDIR
          cd $TESTDIR
          pytest --ignore="tests/test_version_string_consistency.py"
```

.github/workflows/conda_publish.yml
```diff
@@ -74,3 +74,22 @@ jobs:
...
+          conda list
+      - name: Run tests after installing the package
+        shell: bash -l {0}
+        run: |
+          export TESTDIR=$RUNNER_TEMP/matchms/_test
+          [ "$RUNNING_OS" = "Windows" ] && export TESTDIR=$RUNNER_TEMP\\matchms\\_test\\
+          mkdir $TESTDIR
+          cp -iRv tests $TESTDIR
+          cd $TESTDIR/tests
+          pytest --ignore-glob="test_version_string_consistency.py"
```
sverhoeven: Integration test should also be run and don't use glob when you can be exact.

```suggestion
          cp -iRv tests integration-tests $TESTDIR
          cd $TESTDIR
          pytest --ignore="tests/test_version_string_consistency.py"
```

CHANGELOG.md
```diff
@@ -21,8 +21,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+## [0.3.3] - 2020-05-26
```
sverhoeven:
```suggestion
## [0.3.3] - 2020-05-27
```

setup.py
```diff
@@ -21,7 +21,7 @@
     author="Netherlands eScience Center",
     author_email="generalization@esciencecenter.nl",
     url="https://github.com/matchms/matchms",
-    packages=find_packages(),
+    packages=find_packages(exclude=['*tests*']),
```
sverhoeven, line 24: This works, but instead of excluding tests dir you could also remove `tests/__init__.py` to get the same effect.

sverhoeven (changes requested):

setup.py
fdiblen: I prefer to do this in setup.py as this is related to how we package and distribute it. Do you have a preference?

sverhoeven: I like less code/config instead of more, but once we start importing test methods in other test module we might need the __init__.py again. So let's go with your suggestion.

.github/workflows/conda_build.yml
```diff
@@ -201,3 +201,18 @@ jobs:
...
+          conda list
+      - name: Run tests after installing the package
+        shell: bash -l {0}
+        run: |
+          export TESTDIR=$RUNNER_TEMP/matchms/_test
+          [ "$RUNNING_OS" = "Windows" ] && export TESTDIR=$RUNNER_TEMP\\matchms\\_test\\
+          mkdir $TESTDIR
+          cp -iRv tests integration-tests $TESTDIR
+          cd $TESTDIR
+          pytest --ignore="test_version_string_consistency.py"
```
sverhoeven:
```suggestion
          pytest --ignore="tests/test_version_string_consistency.py"
```

.github/workflows/conda_publish.yml
```diff
@@ -74,3 +74,22 @@ jobs:
...
+          conda list
+      - name: Run tests after installing the package
+        shell: bash -l {0}
+        run: |
+          export TESTDIR=$RUNNER_TEMP/matchms/_test
+          [ "$RUNNING_OS" = "Windows" ] && export TESTDIR=$RUNNER_TEMP\\matchms\\_test\\
+          mkdir $TESTDIR
+          cp -iRv tests integration-tests $TESTDIR
+          cd $TESTDIR
+          pytest --ignore="test_version_string_consistency.py"
```
sverhoeven:
```suggestion
          pytest --ignore="tests/test_version_string_consistency.py"
```

sverhoeven (changes requested):

sverhoeven (approved):

PR #51, opened 2020-05-27
Fix conda publish workflow

fdiblen:
### **Patch release: 0.3.4**

Fixes the issues of Anaconda publish workflow.

### Changed
- Fix verify step in conda publish workflow


.github/workflows/conda_publish.yml
```diff
@@ -74,6 +74,13 @@ jobs:
         shell: bash -l {0}
         run: |
           conda install --channel bioconda --channel conda-forge --channel nlesc matchms
+      - name: Remove the package from Anaconda Cloud due to installation failure
+        if: ${{ failure() }}
```
sverhoeven: Normally you would create a new release with a fix. You are not rolling back the release on GH and Zenodo so I would either remove the release from Anaconda, Zenodo and GH or leave them all.

Normally the rest of the steps are not performed when the prev step fails. Is there a build which actually removed the anaconda package?

fdiblen: We need to have a short chat.

sverhoeven (approved):
Easy to review now, looks ok to me.

.github/workflows/conda_publish.yml
fdiblen: done

sverhoeven:
:clap:  Congrats with successful [publish job](https://github.com/matchms/matchms/actions/runs/118814803).

PR #53, opened 2020-05-28
Add and improve docstrings

### LINKED ISSUE #49: Docstrings missing
In my code example (#41) I would like to link to documentation of methods. Sadly the docs do not show the methods because they do not have doc strings.

For

* [x] matchms/similarity/CosineGreedy.py:CosineGreedy
* [x] matchms/filtering/default_filters.py:default_filters, has docstring but does not explain what it does.
* [x] matchms/calculate_scores.py:calculate_scores

### LINKED ISSUE #50: CosineGreedy has incorrect method signature
The signature of CosineGreedy.__call__ is

```python
def __call__(self, spectrum: SpectrumType, reference_spectrum: SpectrumType) -> float
```

But it should be

```python
def __call__(self, spectrum: SpectrumType, reference_spectrum: SpectrumType) -> Tuple[float, int]
```

sverhoeven:
Fixes #49 and fixes #50 
And solves many sphinx build warnings

### DIFF
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 from typing import List
 from typing import Tuple
 import numpy
-from matchms.typing import SimilarityFunction, ReferencesType, QueriesType
+from matchms.typing import QueriesType
+from matchms.typing import ReferencesType
+from matchms.typing import SimilarityFunction
 
 
 class Scores:
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -29,7 +29,7 @@ class Scores:
         scores = Scores(spectrums, spectrums, CosineGreedy()).calculate()
 
         for (reference, query, score, n_matching) in scores:
-            print(f"Cosine score between {reference.id} and {query.id}" +
+            print(f"Cosine score between {reference.get('id')} and {query.get('id')}" +
                   f" is {score:.2f} with {n_matching} matched peaks")
 
     Should output
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -1,16 +1,57 @@
-from typing import Callable
+from __future__ import annotations
 from typing import List
 from typing import Tuple
 import numpy
-from matchms.typing import QueriesType
-from matchms.typing import ReferencesType
+from matchms.typing import SimilarityFunction, ReferencesType, QueriesType
 
 
 class Scores:
-    """An example docstring for a class definition."""
-    def __init__(self, references: ReferencesType, queries: QueriesType, similarity_function: Callable):
-        """An example docstring for a constructor."""
+    """Scores container
 
+    Example to calculate scores between 2 spectrums and iterate over the scores
+
+    .. testcode::
+
+        import numpy as np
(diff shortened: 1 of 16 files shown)


matchms/Scores.py
```diff
@@ -1,16 +1,59 @@
...
 from matchms.typing import QueriesType
 from matchms.typing import ReferencesType
+from matchms.typing import SimilarityFunction
 
 
 class Scores:
-    """An example docstring for a class definition."""
-    def __init__(self, references: ReferencesType, queries: QueriesType, similarity_function: Callable):
-        """An example docstring for a constructor."""
+    """Scores container
```
florian-huber: Could we add an extra explanation beyond "Scores container" , which probably is too computational/technical to make sense for many mass spec users.
For instance: "A scores instance contains reference and query spectrums and allows calculating similarity scores between those spectrums." (maybe also mention querying possibility? )

matchms/similarity/CosineGreedy.py
```diff
@@ -1,13 +1,63 @@
+from typing import Tuple
 import numpy
 from matchms.typing import SpectrumType
 
 
 class CosineGreedy:
+    """Factory to calculate 'greedy cosine score' between mass spectra.
```
florian-huber, line 7:
```suggestion
    """Factory to calculate 'greedy cosine score' between mass spectra.
    
    The cosine score aims at quantifying the similarity between two mass spectra.
    The score is calculated by finding best possible matches between peaks
    of two spectra. Two peaks are considered a potential match if their
    m/z ratios lie within the given 'tolerance'.
    The underlying peak assignment problem is here solved in a 'greedy' way.
    This can perform notably faster, but does occasionally deviate slightly from
    a fully correct solution (as with the Hungarian algorithm). In practice this
    will rarely affect similarity scores notably, in particular for smaller
    tolerances.
```

matchms/utils.py
```diff
@@ -1,32 +1,38 @@
...
-def convert_inchi_to_inchikey(inchi):
+def convert_inchi_to_inchikey(inchi: str) -> Optional[str]:
     return mol_converter(inchi, "inchi", "inchikey")
 
 
-def mol_converter(mol_input, input_type, output_type):
+def mol_converter(mol_input: str, input_type: str, output_type: str) -> Optional[str]:
     """Convert molecular representations using rdkit.
 
     Convert for from smiles or inchi to inchi, smiles, or inchikey.
```
florian-huber:
```suggestion
    Convert input smiles or inchi to inchi, smiles, or inchikey.
```

florian-huber (changes requested):
Super nice, thanks a lot Stefan!

matchms/Scores.py
sverhoeven: Rephrased

matchms/utils.py
sverhoeven: Fixed

sverhoeven:
@florian-huber could you review my changes.

florian-huber (approved):

florian-huber:
Looks very good to me!

Issue #11, opened 2020-04-21
should coverage be calculated only on unit tests?


jspaaks:
(I think so)

fdiblen:
Refs #64

florian-huber:
Just pragmatically speaking: to me it looks fine as it is now. SonarCloud coverage does what it is supposed to do (I think), namely enforcing unit test coverage.
Can we close this @sverhoeven ?

sverhoeven:
Indeed it is a nice to have coverage measured separately, but with the current high number of tests seems unnecessary

PR #64, opened 2020-06-02
only publish unit test results on sonarcloud refs #11

jspaaks:


florian-huber:
THis is outdated and I think given the current coverage it is ok to skip this.

PR #73, opened 2020-06-10
adapt unit test to also run on not compiled numba function

### LINKED ISSUE #72: Sonarcloud does not see test coverage of numba functions
Although covered by unit tests, sonarcloud does not recognize any test coverage of our numba-based functions (e.g. ``collect_peak_pairs``).

florian-huber:
Added tests on numba functions using ``.py_func`` to run them without compiling. This way they will also be covered according to Sonarcloud.

### DIFF
--- a/tests/test_collect_peak_pairs.py
+++ b/tests/test_collect_peak_pairs.py
@@ -1,56 +1,37 @@
 """Test function to collect matching peaks. Run tests both on numba compiled and
 pure Python version."""
 import numpy
+import pytest
 from matchms.similarity.collect_peak_pairs import collect_peak_pairs
 
 
-def test_cosine_hungarian_tolerance_01_compiled():
-    """Test finding expected peak matches within tolerance=0.2."""
+@pytest.mark.parametrize("tolerance, expected_pairs",
+                         [(0.2, [(2, 2, 1.0), (3, 3, 1.0)]),
+                          (5.0, [(0, 0, 1.0), (1, 1, 1.0)])])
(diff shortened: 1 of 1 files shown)


tests/test_collect_peak_pairs.py
```diff
@@ -26,3 +28,29 @@ def test_cosine_hungarian_tolerance_01_shift_min5():
     matching_pairs = collect_peak_pairs(spec1, spec2, tolerance=0.2, shift=-5.0)
     assert len(matching_pairs) == 2, "Expected different number of matching peaks"
     assert matching_pairs == [(0, 0, 1.0), (1, 1, 1.0)], "Expected different matchin pairs."
+
+
+def test_cosine_hungarian_tolerance_01():
```
sverhoeven: Use https://docs.pytest.org/en/stable/parametrize.html to not duplicate tests

sverhoeven (changes requested):

tests/test_collect_peak_pairs.py
florian-huber: I added this, but still kept 2 (instead of 4 before) tests, one for a compiled version and one for  a pure python version.
I could probably also call ``.py_func`` somehow via ``@pytest.mark.parametrize`` but wasn't sure how to do this in a clean way (without spending too much time on it).

sverhoeven: ok

sverhoeven (approved):

florian-huber:
Thanks Stefan for the reviewing!

PR #93, opened 2020-07-02
Metabolomics usi

florian-huber:
Description from @sdrogers:

Added a new method to allow a `Spectrum` object to be created from the json returned in a post request from the  metabolomics USI (https://metabolomics-usi.ucsd.edu).
The USI provides a single access point for a large number of metabolomics MS2 data resources including both experimental and library spectra.


tests/test_usi.py
```diff
@@ -0,0 +1,34 @@
+from unittest.mock import Mock
+from unittest.mock import patch
+import numpy as np
+from matchms import Spectrum
+from matchms.importing.load_from_usi import load_from_usi
```
florian-huber:
```suggestion
from matchms.importing import load_from_usi
```

florian-huber: Added the function to ``__init__.py`` in importing, so the last past can go.
(by the way: if that's OK you can simply click on "commit suggestion" to change it)

sdrogers: This seems to have broken something in the testing pipeline?

sverhoeven: I corrected the tests, mock.patch should use requests directly. I also applied a lint fix and shortened the import as suggested by Florian.

tests/test_usi.py
```diff
@@ -0,0 +1,34 @@
+from unittest.mock import Mock
+from unittest.mock import patch
+import numpy as np
+from matchms import Spectrum
+from matchms.importing import load_from_usi
+
+
+@patch('requests.get')
+def normal_test(mock_get):
```
sverhoeven: Test is ignored as pytest needs methods to start with `test_`. This causes low coverage and SonarCloud failure.
```suggestion
def test_normal(mock_get):
```

matchms/importing/load_from_usi.py
```diff
@@ -0,0 +1,48 @@
+import json
+import numpy as np
+import requests
+from matchms import Spectrum
```
sverhoeven: This imports results in Spectrum being a module instead of class.
```suggestion
from ..Spectrum import Spectrum
```

sverhoeven (changes requested):
Once test_normal is enabled, it fails. Test should be fixed before merging.

tests/test_usi.py
florian-huber: Thanks!

florian-huber:
@sdrogers I changed the expected metadata in ``test_normal()`` to ``{"usi": "something", "server": "https://metabolomics-usi.ucsd.edu", "precursor_mz": None}``. Or did you want to add the actual url to the metadata in ``load_from_usi`` ?

sverhoeven (approved):
Looks good to me. to make even better

* remove duplicate `added` in changelog.
* add example to doc string

florian-huber:
Thanks a lot for the review Stefan!
I also added a short code example and changed the changelog as you suggested. I will merge it as soon as the tests pass.

PR #99, opened 2020-07-13
Cosine powers

sdrogers:
Hi @florian-huber -- see proposed modifications to the scoring functions to allow arbitrary powers. Need to write a few tests (will do so but no time just now).
Also fixed the expected output in one of the modified cosine tests that I think hadn't been updated since you changed the normalisation.


florian-huber:
Thanks a lot @sdrogers! I'll merge it with the branch and then I'll continue adding tests etc from there.

PR #98, opened 2020-07-08
Cosine extension

florian-huber:
- fix wrong denominator in cosine score calculation
- add options to raise peaks to given power

Issue #96, opened 2020-07-05
Cosine similarity misses weighting and removal of precursor


kaibioinfo:
* The requirement for normalizing intensities before calculating the cosine is not necessary. The cosine is, by definition, normalized.
* A classical paper about comparing spectra is Stein et al. 1994. There the suggest to
  * take the square root of the intensity, such that the cosine is not driven by the few peaks with highest intensity
  * multiply the intensity by the mass, such that high mass peaks are higher weighted than low mass peaks
* many implementations of cosine also remove peaks around the parent peak (often in a +-20 Da window), such that the parent and uninformative losses like H2O are not dominating the cosine. Instead of deleting them, one could also downweight them.
* peaks behind the precursor mass should not be considered for the cosine. This is possible by using the select_mass filter, but it is strange that nobody enforces this.

florian-huber:
Thanks for the hints and suggestions.
We extended the cosine score functions:
- removed the non-essential normalization requirement (PR #103).
- added option to Cosine score for raising peaks by power of m/z and/or intensity (PR #99 )

We are currently adding new filters to address the two last points raised above:
- Additional filters #107 to address the two last points raised above (PR #119)

florian-huber:
The remaining parts of this are now in a new issue #107 (and PR #119 ).

PR #103, opened 2020-07-27
remove non-essential normalization check

florian-huber:
With our recent change (#98) spectra do not need to be normalized to derive the cosine score. So, as suggested in #96 I here removed the normalization ``assert``.


sverhoeven (approved):
nice, less is more. Seems we did not test it so coverage goes up.

florian-huber:
Thanks Stefan for the quick review!

PR #112, opened 2020-08-03
Add intensity and mz power options to Cosine Hungarian

florian-huber:
Update ``CosineHungarian`` to be analogue to ``CosineGreedy`` after PR #98 

--> After merging this it seems time for a __new release__!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Read msp files to create Spectrum objects from it [#102](https://github.com/matchms/matchms/pull/102)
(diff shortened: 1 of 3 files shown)


sverhoeven (changes requested):
* Can you add this PR to https://github.com/matchms/matchms/blob/master/CHANGELOG.md#fixed
* Can you add tests to see that passing in non-defaults actually gives different results, similar to https://github.com/matchms/matchms/blob/master/tests/test_cosine_greedy.py#L91-L118

sverhoeven (approved):
Nice, to see more consistency between the Cosine methods

florian-huber:
I added a test using the option to raise powers and added this PR to the changelog. Thanks Stefan.

PR #115, opened 2020-08-04
Additional filtering methods

kiaashour:
Additional filtering methods:

remove_peaks_within_window: Keep peaks only if they are in the top k intense peaks within a desired mz (Da) window.
remove_spectra_above_intensity: Returns none if the number of peaks above a certain threshold relative intensity is less than desired.
remove_spectra_below: Returns none if the precursor mz of a spectrum is above a maximum mz value.
remove_spectra_within_tolerance: Remove peaks that are within an mz tolerance (in Da) of the precursor mz, excluding the precursor peak.


florian-huber:
Hi Kia, 
thanks a lot for your pull request!
Looks like good filter additions, so next steps should be:
- Solve the (mostly very minor) linting issues that the tests now complain about + add tests functions
- Then we will merge it into an own branch on matchms to do the final polishing
- Merge into matchms:master

I guess most of the linting things should be straightforward to change. Let me know if you need help with some remaining ones, or the tests!

kiaashour:
Hi Florian,

Great, thanks. I'll get onto it and let you know if I run into any issues.

matchms/filtering/remove_spectra_below.py
```diff
@@ -0,0 +1,24 @@
+from ..typing import SpectrumType
+
+
+def remove_spectra_below(spectrum: SpectrumType, max_mz: float = 1000) -> SpectrumType:
```
florian-huber:
```suggestion
def remove_spectra_above_mz(spectrum: SpectrumType, max_mz: float = 1000) -> SpectrumType:
```

florian-huber: ``remove_spectra_below`` sounded misleading to me. Maybe we could also use ``remove_spectra_above_precursor_mz`` but that is getting a bit long...

matchms/filtering/remove_spectra_within_tolerance.py
```diff
@@ -0,0 +1,35 @@
+import numpy
+from ..Spikes import Spikes
+from ..typing import SpectrumType
+
+
+def remove_spectra_within_tolerance(spectrum: SpectrumType, mz_tolerance: float = 17) -> SpectrumType:
```
florian-huber:
```suggestion
def remove_peaks_around_precursor_mz(spectrum: SpectrumType, mz_tolerance: float = 17) -> SpectrumType:
```

florian-huber: It's removing peaks not spectra and "within tolerance" also made me expect something else (could be that it's some standard in the field?).

matchms/filtering/remove_peaks_within_window.py
```diff
@@ -0,0 +1,40 @@
+import numpy
+from ..Spikes import Spikes
+from ..typing import SpectrumType
+
+
+def remove_peaks_within_window(spectrum: SpectrumType, k: int = 6,
```
florian-huber:
```suggestion
def remove_peaks_outside_top_k(spectrum: SpectrumType, k: int = 6,
```

florian-huber: Function name seemed to not match the Explanation in the docstring to me. Feel free to propose other names.

matchms/filtering/remove_spectra_above_intensity.py
```diff
@@ -0,0 +1,30 @@
+from ..typing import SpectrumType
+from . import select_by_relative_intensity
+
+
+def remove_spectra_above_intensity(spectrum: SpectrumType, no_peaks: int = 5,
```
florian-huber:
```suggestion
def remove_spectra_below_intensity(spectrum: SpectrumType, no_peaks: int = 5,
```

florian-huber: Sorry, I realize I am interpreting names probably exactly the opposite way. And in particular here I find it hard to find a good name. ``remove_spectra_few_high_peaks``? ``remove_spectra_few_high_intensity_peaks`` ... ?

florian-huber:
Thanks @kiaashour for the linting!
The error in the current tests does not seem to be related to your functions (maybe more something like this: https://github.com/numpy/numpydoc/issues/268). Let's simply see if it still occurs once it is merged into a matchms branch.

I also suggested name changes for your functions. Let me know if those work for you (or if you have other options).

florian-huber (changes requested):
Thanks for the linting.
I suggested some name changes, since the current name often made me expect other functionalities.
Once we solve this we can merge it into a matchms branch and add some unit tests etc.

matchms/filtering/remove_spectra_below.py
kiaashour: Yes I agree, the previous name was quite misleading. `remove_spectra_above_mz ` sounds fine to me.

matchms/filtering/remove_peaks_within_window.py
kiaashour: The name was wrongly worded. I think `keep_peaks_within_window` is what I meant to say, but `remove_peaks_outside_top_k` sounds better.

matchms/filtering/remove_spectra_above_intensity.py
kiaashour: No worries. The idea here was just to remove spectra that have less than some pre-determined number of peaks above a certain threshold intensity. So I think `remove_spectra_few_high_peaks` is probably the best name here.

matchms/filtering/remove_spectra_within_tolerance.py
kiaashour: Yeah should've been peaks and not spectra. "around" is more clear than just tolerance, so I think that's fine.

kiaashour:
Great, I've resolved the name changes. Would we able it to merge it at this stage? I thought there were some more linting checks to be resolved.

florian-huber:
Thanks for the name changes!
And yes, we would need to fix the failing checks. 
I would also like to add some unit tests to the ``tests\`` folder. But some of the checks from the continuous integration will only work when this PR is within matchms, so I would soon merge it into a separate branch within matchms and then we can continue to work on it for a 2nd PR from there (sorry, bit complicated, but currently that seems to work best).

kiaashour:
Ok great. I'll begin have a looking at some unit tests in the meantime.

florian-huber (approved):

florian-huber:
I will merge to own branch on matchms (mostly also to get access to SonarCloud checks).

Issue #107, opened 2020-07-30
Additional filtering methods


kiaashour:
Hey! I've been working on estimating FDRs in metabolomics and I've been using some new filtering methods that currently aren't in the matchms package. Here are the filtering methods I propose:

remove_peaks_within_window: Keep peaks only if they are in the top k intense peaks within a desired mz (Da) window.
remove_spectra_above_intensity: Returns none if the number of peaks above a certain threshold relative intensity is less than desired.
remove_spectra_below: Returns none if the precursor mz of a spectrum is above a maximum mz value.
remove_spectra_within_tolerance: Remove peaks that are within an mz tolerance (in Da) of the precursor mz, excluding the precursor peak.

florian-huber:
Hi Kia, 
sounds good!
Do you have some code on this already that you could make into a pull request?
(best to fork matchms first and then send a PR our way)

kiaashour:
Hi Florian,

Yes, I do indeed have have some code. I will send a PR soon. Let me know if any tests need to be written.

florian-huber:
This was now added with PR #119 .

PR #119, opened 2020-08-12
Additional filter functions

florian-huber:
@kiaashour added 4 new filters (see #107 ):

- ``remove_peaks_around_precursor_mz``
- ``remove_peaks_outside_top_k``
- ``require_minimum_of_high_peaks``
- ``require_precursor_below_mz``


florian-huber:
Hi @kiaashour , I again renamed 2 of the functions to make them consistent with ``require_minimum_number_of_peaks`` filter.
For the tests you could have a look at ``filter\test_require_minimum_number_of_peaks.py`` which should be easy to adapt for some of the new functions.

kiaashour:
Great. I've began writing the tests and should be able to upload them soon.

kiaashour:
Hi Florian. I've written some tests (one still remaining) and updated the filtering functions a bit, and was just wondering how I should push them to the PR. Right now I've sent another PR but I believe that's not needed?

florian-huber:
Thanks Kia for adding tests!
I gave you writing rights, so we should now be able to both work on this PR (and I will merge your changes into this one).

matchms/filtering/require_precursor_below_mz.py
```diff
@@ -0,0 +1,26 @@
...
+        All precursor mz values greater or equal to this
+        will return none. Default is 1000.
+    """
+
+    assert max_mz >= 0, "max_mz must be a positive floating point."
+    precursor_mz = spectrum.get("precursor_mz")
+    if precursor_mz:
+
+        if precursor_mz >= max_mz:
+            return None
```
florian-huber:
```suggestion
    if precursor_mz and precursor_mz >= max_mz:
        return None
```

florian-huber: SonarCloud complains about a "code smell" which is the tools' way of suggesting code improvements. I think the suggested change would solve that.

florian-huber:
Looking good! Just push the missing test and then SonarCloud should be happy (wants code coverage >= 80%).

By the way, SonarCloud is worth checking out! If you click on ``Coverage`` or ``Code smell`` it usually does a good job at pointing to thinks that can be improved.

We should also update the Changelog.

kiaashour:
Great, I'll add the remaining test now and have a look at the Changelog. Will keep an eye on SonarCloud too.

matchms/filtering/require_precursor_below_mz.py
kiaashour: My reason for the nested if statements was in case the precursor_mz was missing (so would be None, rather than 0).  But I suppose we can ignore that?

florian-huber: That will be dealt with by Python. 
If precursor_mz is None than it will not go further to check the ``and ...`` part.
(so if you change the order: ``if precursor_mz >= max_mz and precursor_mz`` that would indeed fail for ``None``)

kiaashour: Ah I see, so it wouldn't give a TypeError.

matchms/filtering/remove_peaks_around_precursor_mz.py
```diff
@@ -0,0 +1,34 @@
...
+    mzs, intensities = spectrum.peaks
+    new_mzs, new_intensities = mzs, intensities
+    if precursor_mz:
+        for i, mz in enumerate(mzs):
+
+            if abs(precursor_mz-mz) <= mz_tolerance and mz != precursor_mz:
+                new_mzs[i], new_intensities[i] = numpy.nan, numpy.nan
+
+        nans = numpy.isnan(new_mzs)
+        new_mzs, new_intensities = new_mzs[~nans], new_intensities[~nans]
```
florian-huber:
```suggestion
    if precursor_mz:
        peaks_to_remove = ((numpy.abs(precursor_mz-mzs) <= mz_tolerance) & (mzs != precursor_mz))
        new_mzs, new_intensities = mzs[~peaks_to_remove], intensities[~peaks_to_remove]
```

florian-huber: I guess that a numpy way of implementing this might be faster because it won't need the loop. And makes it a bit simpler code-wise. Should be doing exactly the same thing (I hope).

matchms/filtering/remove_peaks_outside_top_k.py
```diff
@@ -0,0 +1,40 @@
...
+    indices = [i for i in range(len(mzs)) if i not in top_k]
+    new_mzs, new_intensities = mzs, intensities
+    for i in indices:
+
+        compare = abs(mzs[i]-k_ordered_mzs) <= mz_window
+        if not numpy.any(compare):
+            new_mzs[i], new_intensities[i] = numpy.nan, numpy.nan
+
+    nans = numpy.isnan(new_mzs)
+    new_mzs, new_intensities = new_mzs[~nans], new_intensities[~nans]
```
florian-huber:
```suggestion
    keep_idx = [i for i in top_k]
    for i in indices:

        compare = abs(mzs[i]-k_ordered_mzs) <= mz_window
        if numpy.any(compare):
            keep_idx.append(i)

    keep_idx.sort()
    new_mzs, new_intensities = mzs[keep_idx], intensities[keep_idx]
```

florian-huber: I would rather avoid using the NaN's here as markers.

florian-huber:
Hi @kiaashour . Thanks for adding the missing test!
I went over the functions again and made some suggestions to avoid using NaN's as markers.

matchms/filtering/remove_peaks_around_precursor_mz.py
kiaashour: Yeah works just the same. Thanks.

florian-huber:
Thanks for the changes Kia. I updated the changelog and merged master into this branch.

kiaashour:
Great, thank you. Shall we wait for @sverhoeven to review before merging?

florian-huber:
Yes, we should wait for the review. That's usually a very good last check before merging.

matchms/filtering/remove_peaks_around_precursor_mz.py
```diff
@@ -0,0 +1,28 @@
...
+        Input spectrum.
+    mz_tolerance
+        Tolerance of mz values that are not allowed to lie
+        within the precursor mz. Default is 17 Da.
+    """
+
+    assert mz_tolerance >= 0, "mz_tolerance must be a positive floating point."
+    precursor_mz = spectrum.get("precursor_mz")
+    mzs, intensities = spectrum.peaks
+    if precursor_mz:
```
sverhoeven: If there is no precursor then spectrum is returned. I expected an exception to be thrown because the filter requires a precursor. See for example [add_losses()](https://github.com/matchms/matchms/blob/master/matchms/filtering/add_losses.py#L25)

tests/test_remove_peaks_around_precursor_mz.py
```diff
@@ -0,0 +1,39 @@
+import numpy
+from matchms import Spectrum
+from matchms.filtering import remove_peaks_around_precursor_mz
+
+
+def test_remove_peaks_around_precursor_mz_no_params_or_precursor():
+    """No parameters provided and no precursor mz present."""
```
sverhoeven: Use defaults instead of no parameters.
```suggestion
def test_remove_peaks_around_precursor_mz_defaults_no_precursor():
     """Using defaults and no precursor mz present."""
```

matchms/filtering/remove_peaks_outside_top_k.py
```diff
@@ -0,0 +1,40 @@
+import numpy
+from ..Spikes import Spikes
+from ..typing import SpectrumType
+
+
+def remove_peaks_outside_top_k(spectrum: SpectrumType, k: int = 6,
+                               mz_window: float = 50) -> SpectrumType:
+
+    """Keep peaks only if they are in the top k intense peaks
+       within a desired mz (Da) window.
```
sverhoeven: Dont understand what the mz window does. 
1. If one top peak has a peak (outside top k) near it within the mz window will it be kept or removed?
2. If 2 top k peaks are within the window will one be discarded?

sverhoeven (changes requested):
Please clone Spectrum instead of overwriting in-place in all the functions. In all other filtering function we use a cloned spectrum a we want to treat the Spectrum objects as immutable.
```python
```

florian-huber:
> 
> 
> Please clone Spectrum instead of overwriting in-place in all the functions. In all other filtering function we use a cloned spectrum a we want to treat the Spectrum objects as immutable.
> 
> ```python
> spectrum = spectrum_in.clone()
> ```

Good catch!
I added a test to all four test files to check if the input spectrum is properly cloned.

matchms/filtering/remove_peaks_outside_top_k.py
florian-huber: @kiaashour can you edit the description in the docstring?

To Stefans questions: 1. that will be kept. 2. no, all top k peaks are kept.

My docstring suggestion: ``"""Remove all peaks which are not within *mz_window* of at least one of the *k* highest intensity peaks of the spectrum.``

matchms/filtering/remove_peaks_around_precursor_mz.py
florian-huber: Just saw that also in ``add_losses()`` we have the assert after ``if precursor_mz``. 
We can 1) add the same assert function here. In that case we accept missing precursor_mz entries, but if there is an entry it must be a number.
What you suggest here is 2) add the assert before the if and make the function fail when there is no numerical ``precursor_mz`` (right?). To me that's fine as well and indeed seems cleaner. We just would have to check if that breaks something. And if we do, we should do it for both this function and ``add_losses``.

kiaashour: 2) would be clearer to me also. Can't really see it breaking anything in the filter function or add_losses.

kiaashour:
I think I will also add a check for when the input spectra are None as well as that's been done on all the other filter functions.

matchms/filtering/remove_peaks_around_precursor_mz.py
florian-huber: Let's do 2 then.

kiaashour: I think we should also add the same thing for `require_precursor_below_mz` for consistency.

florian-huber: Definitely! Good point.

kiaashour: Some of the tests fail when we add an assertion for the missing precursor mz. For `require_precursor_below_mz` the cloning test fails (raises an assertion error for missing precursor mz) and the test for when the precursor mz is absent fails as well. I suppose we can just remove the test for missing precursor mz and add a dummy precursor mz to the cloning test?

florian-huber: Yes, we can simply add a precursor_mz to the ``test_require_precursor_below_mz``.

kiaashour: Shall I also make the same changes for `add_losses`?

florian-huber: I'll change add_losses in a separate, tiny PR, just to not make this PR even bigger. (I also added an issue for that).

florian-huber:
Thanks for the changes @kiaashour !
I just added a few tests to check if the assert messages are raised correctly and asked Stefan to check again if his comments were all addressed.

matchms/filtering/remove_peaks_around_precursor_mz.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def remove_peaks_around_precursor_mz(spectrum_in: SpectrumType, mz_tolerance: float = 17) -> SpectrumType:
+
+    """Remove peaks that are within mz_tolerance (in Da) of
+       the precursor mz, exlcuding the precursor peak.
+
+    Parameters
+    ----------
+    spectrum
```
sverhoeven: Match docstring with function spec
```suggestion
    spectrum_in
```

matchms/filtering/remove_peaks_outside_top_k.py
```diff
@@ -0,0 +1,44 @@
...
+
+def remove_peaks_outside_top_k(spectrum_in: SpectrumType, k: int = 6,
+                               mz_window: float = 50) -> SpectrumType:
+
+    """Remove all peaks which are not within *mz_window* of at least one
+       of the *k* highest intensity peaks of the spectrum.
+
+    Parameters
+    ----------
+    spectrum
```
sverhoeven: Match docstring with function spec
```suggestion
    spectrum_in
```

matchms/filtering/require_minimum_of_high_peaks.py
```diff
@@ -0,0 +1,34 @@
...
+
+def require_minimum_of_high_peaks(spectrum_in: SpectrumType, no_peaks: int = 5,
+                                  intensity_percent: float = 2.0) -> SpectrumType:
+
+    """Returns None if the number of peaks with relative intensity
+       above or equal to intensity_percent is less than no_peaks.
+
+    Parameters
+    ----------
+    spectrum
```
sverhoeven: Match docstring with function spec
```suggestion
    spectrum_in
```

matchms/filtering/require_precursor_below_mz.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def require_precursor_below_mz(spectrum_in: SpectrumType, max_mz: float = 1000) -> SpectrumType:
+
+    """Returns None if the precursor_mz of a spectrum is above
+       max_mz.
+
+    Parameters
+    ----------
+    spectrum
```
sverhoeven: Match docstring with function spec
```suggestion
    spectrum_in
```

sverhoeven (approved):
Nice work. 100% coverage and useful additions to library.

Just some small docstring fixes.

matchms/filtering/remove_peaks_around_precursor_mz.py
kiaashour: Ok great.

kiaashour:
Shall we just keep both changes from master and this branch for the conflict, seeing that we'll be merging this into master?

florian-huber:
@kiaashour I indeed kept both changes in the changelog. 

I also added you to the list of contributors (``.zenodo.json`` and ``CITATION.cff``). Please check if your name and affiliation is correct, and add an orcid if you have one. Once that is done I'll merge it into master.

kiaashour:
Great, thank you. Looks all fine to me, added my orcid ID as well.

florian-huber:
Thanks a lot @kiaashour and @sverhoeven , great work!

PR #124, opened 2020-08-13
perfect tense of build is built

jspaaks:


florian-huber:
Sorry, that one got lost during merging I guess.

PR #142, opened 2020-09-01
Add scores_by_reference and scores_by_query to Scores

### LINKED ISSUE #78: Add methods to "Scores" object to better query and sort matches
Scores currently allows to either access the underlying similarity score array by ``scores.scores``, or to loop through the scores, for instance:
```python
for (ref, query, score, nmatch) in scores:
    print(ref.get("spectrumid"), query.get("spectrumid"), score, nmatch)
```
That is fine for the test examples so far. But when handling a lot of queries and references I realized that it is rather unintuitive. 

For example: 
If I have 10 query spectrums that I want to compare with a large database of -say- 10,000 spectrums, I typically would like to get the best matching reference spectrums per query spectrum.

Maybe we could add a class method ``sort_per_query``?
E.g. using ``numpy.argpartition(scores.scores, -top_n, axis=0)[-top_n:, :]``, with ``top_n`` the number of best matches to display per query...

florian-huber:
implemented: 
- ``Scores.scores_by_reference()``
- ``Scores.scores_by_query()``

### DIFF
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -121,31 +121,26 @@ class Scores:
         self._scores = self.similarity_function(self.references[:, 0], self.queries[0, :])
         return self
 
-    def scores_by_reference(self, references: ReferencesType) -> numpy.ndarray:
-        """Return all scores (not sorted) for the given reference spectrums.
+    def scores_by_reference(self, reference: ReferencesType) -> numpy.ndarray:
+        """Return all scores (not sorted) for the given reference spectrum.
 
         Parameters
         ----------
-        references
-            Single Spectrum or list or array of reference objects.
+        reference
+            Single reference Spectrum.
         """
-        if not isinstance(references, (list, numpy.ndarray)):
-            references = [references]
+        assert reference in self.references, "Given input not found in references."
+        selected_idx = int(numpy.where(self.references[:, 0] == reference)[0])
 
-        selected_idx = []
-        for ref in references:
-            assert ref in self.references, "Given input not found in references."
-            selected_idx.append(int(numpy.where(self.references[:, 0] == ref)[0]))
+        return list(zip(self.queries[0], self._scores[selected_idx, :].copy()))
 
-        return self._scores[selected_idx, :].copy()
-
-    def scores_by_query(self, queries: QueriesType) -> numpy.ndarray:
+    def scores_by_query(self, query: QueriesType) -> numpy.ndarray:
         """Return all scores (not sorted) for the given query spectrums.
 
         Parameters
         ----------
-        queries
-            Single Spectrum or list or array of query objects.
+        query
+            Single query Spectrum.
 
         For example
 
@@ -171,18 +166,13 @@ class Scores:
             queries = [spectrum_2, spectrum_3, spectrum_4]
 
             scores = Scores(references, queries, CosineGreedy()).calculate()
-            selected_scores = scores.scores_by_query([spectrum_3, spectrum_4])
+            selected_scores = scores.scores_by_query(spectrum_3)
 
         """
-        if not isinstance(queries, (list, numpy.ndarray)):
-            queries = [queries]
+        assert query in self.queries, "Given input not found in queries."
+        selected_idx = int(numpy.where(self.queries[0, :] == query)[0])
 
-        selected_idx = []
-        for query in queries:
-            assert query in self.queries, "Given input not found in queries."
-            selected_idx.append(int(numpy.where(self.queries[0, :] == query)[0]))
-
-        return self._scores[:, selected_idx].copy()
+        return list(zip(self.references[:, 0], self._scores[:, selected_idx].copy()))
 
     @property
     def scores(self) -> numpy.ndarray:
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 import numpy
-from matchms import Spectrum
 from matchms.typing import QueriesType
 from matchms.typing import ReferencesType
 from matchms.typing import SimilarityFunction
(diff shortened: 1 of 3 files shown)


matchms/Scores.py
```diff
@@ -121,6 +121,69 @@ def calculate_parallel(self) -> Scores:
         self._scores = self.similarity_function(self.references[:, 0], self.queries[0, :])
         return self
 
+    def scores_by_reference(self, references: ReferencesType) -> numpy.ndarray:
```
sverhoeven: As function name uses `reference` I expected it would work with a single reference as argument. Why is there support for multiple references? How do you know which score belongs to which reference?

If you want to get the scores for multiple references is not easier to fetch scores for each reference in a loop calling the function. 

I would like to see the function accept only a single reference Spectrum.

matchms/Scores.py
```diff
@@ -121,6 +121,69 @@ def calculate_parallel(self) -> Scores:
...
+        """
+        if not isinstance(references, (list, numpy.ndarray)):
+            references = [references]
+
+        selected_idx = []
+        for ref in references:
+            assert ref in self.references, "Given input not found in references."
+            selected_idx.append(int(numpy.where(self.references[:, 0] == ref)[0]))
+
+        return self._scores[selected_idx, :].copy()
```
sverhoeven: The function does not return to which query Spectrum each score belongs. Is this useful?

sverhoeven (changes requested):
Added some questions to scores_by_reference(), same questions hold for scores_by_query().

matchms/Scores.py
florian-huber: I wasn't sure either.
Alternatives would be to return 
1. A list of tuples as you suggested in #78 : ``[(Spectrum, score)]`` where score can also be tuple of course.
2. A sliced ``Scores`` instance. 

In each of those cases we shouldn't make actual copies of the spectrum objects. That's fine, but we should keep in mind that all those bindings we create could lead to unexpected results later?

florian-huber: (1) is of course much easier to implement, e.g. using 
```python
return list(zip(self.queries[0], self._scores[selected_idx, :].copy()))
```
, for (2) we would have to create a new Scores instance and probably also add a slicing method or similar...

sverhoeven: List of tuples would be my preference. Getting another Scores object, would require another call to get for example nr matches.

florian-huber: Agreed. I changed it.

matchms/Scores.py
florian-huber: I changed that.

sverhoeven: Great

tests/test_scores.py
```diff
@@ -130,3 +132,51 @@ def test_scores_next():
...
+    spectrum_4 = Spectrum(mz=numpy.array([100, 150, 200.]),
+                          intensities=numpy.array([0.6, 0.1, 0.6]),
+                          metadata={'id': 'spectrum4'})
+    references = [spectrum_1, spectrum_2, spectrum_3]
+    queries = [spectrum_3, spectrum_4]
+
+    scores = Scores(references, queries, CosineGreedy()).calculate()
+    selected_scores = scores.scores_by_reference(spectrum_2)
+
+    expected_result = [(scores.queries[:, i][0], scores.scores[1, i]) for i in range(2)]
```
sverhoeven, line 182:
```suggestion
    expected_result = [(scores.queries[:, i][0], scores.scores[1, i][0], scores.scores[1, i][1]) for i in range(2)]
```

tests/test_scores.py
```diff
@@ -130,3 +132,51 @@ def test_scores_next():
...
+    spectrum_4 = Spectrum(mz=numpy.array([100, 150, 200.]),
+                          intensities=numpy.array([0.6, 0.1, 0.6]),
+                          metadata={'id': 'spectrum4'})
+    references = [spectrum_1, spectrum_2, spectrum_3]
+    queries = [spectrum_2, spectrum_3, spectrum_4]
+
+    scores = Scores(references, queries, CosineGreedy()).calculate()
+    selected_scores = scores.scores_by_query(spectrum_4)
+
+    expected_result = [(scores.references[i][0], scores.scores[i, 2]) for i in range(3)]
```
sverhoeven, line 230:
```suggestion
    expected_result = [(scores.references[i][0], scores.scores[i, 2][0], scores.scores[i, 2][1]) for i in range(3)]
```

sverhoeven (changes requested):
Can the nested tuple be flattened, similar to `__iter__()`, so scores_by_reference returns [(query,ref, score, nmatches)] for CosineGreedy?

Now you get 
```
[(<matchms.Spectrum.Spectrum at 0x7f3f97a71650>, (0.0, 0)),
 (<matchms.Spectrum.Spectrum at 0x7f3f97a71690>, (0.1363196353181994, 1))]
```
I would like
```
[(<matchms.Spectrum.Spectrum at 0x7f3f97a71650>, 0.0, 0),
 (<matchms.Spectrum.Spectrum at 0x7f3f97a71690>, 0.1363196353181994, 1)]
```

tests/test_scores.py
florian-huber: I did try to unpack the tuples. For instance like this:
```python
    def scores_by_reference(self, reference: ReferencesType) -> numpy.ndarray:
        """Return all scores (not sorted) for the given reference spectrum.

        Parameters
        ----------
        reference
            Single reference Spectrum.
        """
        assert reference in self.references, "Given input not found in references."
        selected_idx = int(numpy.where(self.references[:, 0] == reference)[0])
        selected_scores = []
        for i, score in enumerate(self._scores[selected_idx, :].copy()):
            if isinstance(score, tuple):
                selected_scores.append((self.queries[0, i], *score))
            else:
                selected_scores.append((self.queries[0, i], score))
        return selected_scores
```
However that makes it slower (about 2-3 x). Maybe not a big issue

florian-huber: But we can also reconsider it later when we work on #143 I guess.

sverhoeven: Could switch to yielding instead of appending, the slowness will be spread around instead of having to wait a longer time once. 
Agree, lets address that in #143

sverhoeven (approved):
Nice addition to the library, works for me.

Did a small Sphinx fix myself.

florian-huber:
Thanks a lot Stefan for help and the review!

PR #151, opened 2020-10-14
Make functions fit for use of GCMS data

### LINKED ISSUE #83: Preparations for GC MS
We will soon try using matchms on GC MS data.
From what we can see already it might be good to define 2 different ``default_filters``.
E.g.:
- ``default_filters_lcms``
- ``default_filters_gcms`` (not using any filters that rely on precursor_mz or parent_mass)

florian-huber:
- raise exception when ``reduce_to_number_of_peaks`` is used with ``desired_ratio`` but without parent mass (e.g. for GC MS data). Before it would then just ignore the parameter, but that can be very unexpected behavior.
- fix small issue with msp importer function

### DIFF
--- a/matchms/filtering/reduce_to_number_of_peaks.py
+++ b/matchms/filtering/reduce_to_number_of_peaks.py
@@ -28,7 +28,7 @@ def reduce_to_number_of_peaks(spectrum_in: SpectrumType, n_required: int = 1, n_
         if parent_mass and ratio_desired:
             n_desired_by_mass = int(ceil(ratio_desired * parent_mass))
             return max(n_required, n_desired_by_mass)
-        elif not ratio_desired:
+        if not ratio_desired:
             return n_max
         raise ValueError("Cannot use ratio_desired for spectrum without parent_mass.")
 
--- a/matchms/filtering/reduce_to_number_of_peaks.py
+++ b/matchms/filtering/reduce_to_number_of_peaks.py
@@ -19,6 +19,8 @@ def reduce_to_number_of_peaks(spectrum_in: SpectrumType, n_required: int = 1, n_
         Maximum number of peaks. Remove peaks if more peaks are found.
     ratio_desired:
         Set desired ratio between maximum number of peaks and parent mass.
+        For spectra without parent mass (e.g. GCMS spectra) this will raise an
+        error when ratio_desired is used.
         Default is None.
     """
     def _set_maximum_number_of_peaks_to_keep():
@@ -26,7 +28,9 @@ def reduce_to_number_of_peaks(spectrum_in: SpectrumType, n_required: int = 1, n_
         if parent_mass and ratio_desired:
             n_desired_by_mass = int(ceil(ratio_desired * parent_mass))
             return max(n_required, n_desired_by_mass)
-        return n_max
+        elif not ratio_desired:
+            return n_max
+        raise ValueError("Cannot use ratio_desired for spectrum without parent_mass.")
 
     def _remove_lowest_intensity_peaks():
         mz, intensities = spectrum.peaks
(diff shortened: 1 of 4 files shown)


matchms/filtering/reduce_to_number_of_peaks.py
```diff
@@ -19,14 +19,18 @@ def reduce_to_number_of_peaks(spectrum_in: SpectrumType, n_required: int = 1, n_
...
     """
     def _set_maximum_number_of_peaks_to_keep():
         parent_mass = spectrum.get("parent_mass", None)
         if parent_mass and ratio_desired:
             n_desired_by_mass = int(ceil(ratio_desired * parent_mass))
             return max(n_required, n_desired_by_mass)
-        return n_max
+        if not ratio_desired:
+            return n_max
+        raise ValueError("Cannot use ratio_desired for spectrum without parent_mass.")
```
sverhoeven, line 33: Can you add a test for when ratio_desired=None and spectrum has no parent.

matchms/importing/load_from_msp.py
```diff
@@ -14,7 +14,7 @@ def parse_msp_file(filename: str) -> Generator[dict, None, None]:
     # Peaks counter. Used to track and count the number of peaks
     peakscount = 0
 
-    with open(filename, 'r') as f:
+    with open(filename, 'r', errors='replace') as f:
```
sverhoeven: To me it is dangerous to replace things while reading the users input file without notifying the user. 
Do you have an example where this was needed?

sverhoeven (changes requested):

matchms/importing/load_from_msp.py
florian-huber: Yes. I got an unicode error when I tried to re-run a notebook from Efrain (student of Simon).
Here is the notebook: https://github.com/florian-huber/spec2vec_gcms_data_analysis/blob/master/notebooks/FH_re_run_pipeline_01.ipynb
The file (also present in the repo) is "MoNA-export-GC-MS.msp".

sverhoeven: Hmm, on a cloned repo I was able to read the file with `f = open('MoNA-export-GC-MS.msp').read()` with ipython 7.18.1 Python 3.8.5 n Ubuntu 20.04 using

```python

In [1]: 
   ...: import os
   ...: import sys
   ...: import numpy as np

In [2]: msp_file ="MoNA-export-GC-MS.msp"

In [3]: from matchms.filtering import normalize_intensities
   ...: from matchms.filtering import reduce_to_number_of_peaks
   ...: from matchms.filtering import require_minimum_number_of_peaks
   ...: from matchms.filtering import select_by_mz
   ...: from matchms.importing import load_from_msp

In [4]: z = list(load_from_msp(msp_file))

In [5]: len(z)
Out[5]: 14847

```
I did not get an unicode error, so I am unable to reproduce.

florian-huber: I had it run on a Windows machine and get
``UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 135: character maps to <undefined>``

florian-huber: I changed it to ``open(filename, 'r', encoding='utf-8')`` (thought I'd tested that before, but now seems to run fine).

matchms/filtering/reduce_to_number_of_peaks.py
florian-huber: Added a test case for ratio_desired=None but with parent mass (the one without both is already present since default is None).

sverhoeven (approved):
OK, encoding should work for ascii and utf

florian-huber:
Thanks a lot Stefan for reviewing!

Issue #152, opened 2020-10-14
Currently it is not clear how users should query the scores


florian-huber:
In our simple code examples we use something like ``for (reference, query, score, n_matching) in scores:``

For most real-life applications this is not very feasible because scores will be a *huge* matrix!
We should either provide a better way to query the scores (which I would prefer), or change or comment the code examples.

sverhoeven:
There are the `scores_by_query(query)` and `scores_by_reference(reference)` methods on Scores class. 

What other way do you propose? A method to filter scores that returns a sparse matrix?

florian-huber:
I am not yet entirely sure.
What I certainly would like to add is:
- provide additional code examples (and/or update the current ones) that use `scores_by_query(query)` and `scores_by_reference(reference)`

In addition, it seems that a very common function/method that users may want is to return the top-*n* scores(+candidates) for every query. In principle that is not overly complicated and can be done using numpy functions (e.g. on top of the `scores_by...` method). But it still might not be straightforward to all users, which would make it a nice additional functionality.
I guess we should not include something like this as a method of `Scores`, but rather provide a separate function for it, say `return_top_scores` or `find_best_matches`.
Or what do you think?

sverhoeven:
Before due to unknown return of the similarity functions we had to pass in a lambda to use the Scores class to perform sorting.
Now that we are using BaseSimilarity class, we could add a order/sort method to it so that we can do the top-n . So only similarity functions which return more than a single number will need to override it.

florian-huber:
That sounds good to me!
Because as you said, in the only current code example we provide for `scores_by_query` we used a lambda function. That does the job, but usually numpy `argsort` or similar functions are much faster. Making those choices for the user by implementing a sort method would be really nice I think.

My first impulse would be to add a method on top of the current methods, e.g. `scores_by_query(query, sorted="False"/"ascending"/"descending")`?
But we could also implement an additional, separate method to `Scores`.

sverhoeven:
Which direction (ascending/descending) of the sort depends on the similarity function, so I would not add it as an argument. Just use the direction that the `BaseSimilarity.order(spectrums, similarities)` method implements. Or are you sometimes interested in the worst scoring pairs?

[...]

Issue #143, opened 2020-09-01
Consider making Scores.scores a structured array


florian-huber:
Since we are moving towards inheriting from a base class for the similarity scores, I think it could be nice to switch to structured arrays for ``Scores.scores``.

Based on #135 (or a test implementation in PR #139 ) we could change the base class to: 
```python
class BaseSimilarityFunction:
    """Similarity function base class.
    When building a custom similarity measure, inherit from this class and implement
    the desired methods.
    """
    # Set key characteristics as class attributes
    is_commutative = True
    # Set output data type, e.g. ("score", "float") or [("score", "float"), ("matches", "int")]
    score_datatype = ("score", "float")

    def pair(self, reference: SpectrumType, query: SpectrumType) -> float:
        """...
        """
        raise NotImplementedError

    def matrix(self, references: List[SpectrumType], queries: List[SpectrumType],
               is_symmetric: bool = False) -> numpy.ndarray:
        """...
        """
        n_rows = len(references)
        n_cols = len(queries)
        scores = numpy.empty([n_rows, n_cols], dtype=self.score_datatype)
        for i_ref, reference in enumerate(references[:n_rows]):
        ...
```

For ``CosineGreedy`` we would then have to add:
```python
class CosineGreedy(BaseSimilarityFunction):
    """... """
    score_datatype = [("score", "float"), ("matches", "int")]
```
For ``ParentmassMatch`` it would become:
```python
class ParentmassMatch(BaseSimilarityFunction):
    """... """
    score_datatype = ("score", "bool")
```

That has 2 advantages in my eyes:
- It makes clearer what ``Scores.scores`` actually contains (name and data type).
- It can even be used to do splits in cases where we have tuples, e.g. ``scores["score"]`` vs ``scores["matches"]``

florian-huber:
We could also enforce it a bit more by
```python
class BaseSimilarityFunction:
    """Similarity function base class.
    When building a custom similarity measure, inherit from this class and implement
    the desired methods.
    """
    # Set key characteristics as class attributes
    is_commutative = True
    # Set output data type, e.g. ("score", "float") or [("score", "float"), ("matches", "int")]
    score_datatype = ("score", "float")

    def compute_similarity_pair(self, reference: SpectrumType, query: SpectrumType) -> float:
        raise NotImplementedError

    def compute_similarity_matrix(self, references: List[SpectrumType], queries: List[SpectrumType],
                                  is_symmetric: bool = False) -> numpy.ndarray:
[...]

PR #153, opened 2020-10-19
Refactors Scores and add querying/sorting options

florian-huber:
Here I changed quite some things to make the API nicer to work with:
- address #143 --> make `Scores.scores` a structured array (only when score contains more than one component, e.g. score + matches)
- add sorting option to ``scores_by_reference`` and ``scores_by_query`` method (implemented via ``BaseSimilarity``)
- fix bug in ``Spectrum()``: getter for metadata entries did not use ``copy()``
- fix bug/issue in ``Spectrum()``: ``__eq__`` method compared metadata dictionaries simply with ``==``. However, that fails when numpy array are added to the metadata (which we do in the ``add_fingerprints`` filter).
- added tests for those cases

In #152 we had also discussed adding ``Scores.top_scores_by_query(query, n=np.Inf)`` and ``Scores.top_scores_by_reference(reference, n=np.Inf)``, but this is not done here yet. For time reasons I would postpone it for another round of additions.


florian-huber:
@sverhoeven Sorry for the back and forth (draft-not-draft...). Took me some time to find the bugs/issues that were causing weird behavior (they were in Spectrum), but that should be fixed now. Just added a missing unit test and now it should be ready for review.

tests/test_spectrum.py
```diff
@@ -26,6 +26,69 @@ def _create_test_spectrum_with_intensities(intensities):
...
+    peaks_mz += 100.0
+    assert numpy.all(spectrum.peaks.mz == numpy.array([100.0, 101.0])), "Expected different peaks.mz"
+    metadata = spectrum.metadata
+    metadata["added_info"] = "this"
+    assert spectrum.metadata == {'testdata': 1}, "Expected metadata to remain unchanged"
+
+
+def test_comparing_spectra_with_metadata():
+    """Test if spectra can be compared that contain numpy.arrays in the metadata.
+    (Failed in an earlier version)"""
```
florian-huber:
```suggestion
    """Test if spectra with (slightly) different metadata are correctly compared."""
```

integration-tests/test_user_workflow.py
```diff
@@ -44,23 +45,31 @@ def apply_my_filters(s):
...
+        (references[48], queries[46], numpy.array([(0.9990141860269471, 27)], dtype=score_datatype)[0]),
+        (references[46], queries[50], numpy.array([(0.9988793406908721, 22)], dtype=score_datatype)[0]),
+        (references[50], queries[46], numpy.array([(0.9988793406908721, 22)], dtype=score_datatype)[0]),
+        (references[57], queries[59], numpy.array([(0.9982171275552503, 46)], dtype=score_datatype)[0]),
+        (references[59], queries[57], numpy.array([(0.9982171275552503, 46)], dtype=score_datatype)[0]),
+        (references[73], queries[74], numpy.array([(0.9973823244169199, 23)], dtype=score_datatype)[0]),
+        (references[74], queries[73], numpy.array([(0.9973823244169199, 23)], dtype=score_datatype)[0]),
     ]
-    assert actual_top10 == expected_top10
+    assert [x[2][0] for x in actual_top10] == pytest.approx([x[2][0] for x in expected_top10], 1e-8), \
```
sverhoeven, line 68: Sad, that the asserts now must be done per column, making it harder to read. 
Looked at [numpy.testing.assert_allclose](https://numpy.org/devdocs/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose), but couldn't get it to work.

For posterity:
```python
    assert [x[0:2] for x in actual_top10] == [x[0:2] for x in expected_top10], \
        "Expected different references and queries."
    assert numpy.testing.assert_allclose(
        numpy.array([x[2] for x in actual_top10]),
        numpy.array([x[2] for x in expected_top10])
    ), "Expected different scores."
```

matchms/Scores.py
```diff
@@ -39,9 +39,9 @@ class Scores:
         similarity_measure = CosineGreedy()
         scores = calculate_scores(references, queries, similarity_measure)
 
-        for (reference, query, score, n_matching) in scores:
+        for (reference, query, score) in scores:
             print(f"Cosine score between {reference.get('id')} and {query.get('id')}" +
-                  f" is {score:.2f} with {n_matching} matched peaks")
+                  f" is {score['score']:.2f} with {score['matches']} matched peaks")
 
     Should output
```
sverhoeven, line 47: Line 95-97 should be updated, as there is no longer a similarity function which returns a tuple.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- scores_by_query and scores_by reference now accept sort=True to return sorted scores [#153](https://github.com/matchms/matchms/pull/153)
+
+### Changed
+
+- Scores.scores is now returning a structured array [#153](https://github.com/matchms/matchms/pull/153)
```
sverhoeven, line 16: Please also mention the `BaseSimilarity.sort` and `BaseSimilarity.score_datatype`, someone making a Spec2vec like library should be aware of the changes to the BaseSimilarity class.

matchms/Scores.py
```diff
@@ -128,39 +128,46 @@ def calculate(self) -> Scores:
...
+                            sort: bool = False) -> numpy.ndarray:
         """Return all scores (not sorted) for the given reference spectrum.
 
         Parameters
         ----------
         reference
             Single reference Spectrum.
+        sort
+            Set to True to obtain the scores in a sorted way (relying on the
+            sorting function from the given similarity_function).
```
sverhoeven: Use correct name of function and link it
```suggestion
            :meth:`~.BaseSimilarity.sort` function from the given similarity_function).
```

matchms/Scores.py
```diff
@@ -128,39 +128,46 @@ def calculate(self) -> Scores:
...
         """Return all scores (not sorted) for the given query spectrum.
 
         Parameters
         ----------
         query
             Single query Spectrum.
-
+        sort
+            Set to True to obtain the scores in a sorted way (relying on the
+            sorting function from the given similarity_function).
```
sverhoeven: Use correct name of function and link it
```suggestion
            :meth:`~.BaseSimilarity.sort` function from the given similarity_function).
```

matchms/Scores.py
```diff
@@ -128,39 +128,46 @@ def calculate(self) -> Scores:
...
-            else:
-                selected_scores.append((self.queries[i], score))
+        if sort:
+            query_idx_sorted = self.similarity_function.sort(self._scores[selected_idx, :])
+            selected_scores = list(zip(self.queries[query_idx_sorted],
+                                       self._scores[selected_idx, query_idx_sorted].copy()))
+        else:
+            selected_scores = list(zip(self.queries, self._scores[selected_idx, :].copy()))
+
         return selected_scores
```
sverhoeven: Nice to see functional rewrite. Could be simplified to
```suggestion
            query_idx_sorted = self.similarity_function.sort(self._scores[selected_idx, :])
            return list(zip(self.queries[query_idx_sorted],
                                       self._scores[selected_idx, query_idx_sorted].copy()))
        return list(zip(self.queries, self._scores[selected_idx, :].copy()))
```

matchms/Scores.py
```diff
@@ -192,12 +198,12 @@ def scores_by_query(self, query: QueriesType) -> numpy.ndarray:
...
-                selected_scores.append((self.references[i], *score))
-            else:
-                selected_scores.append((self.references[i], score))
+        if sort:
+            references_idx_sorted = self.similarity_function.sort(self._scores[:, selected_idx])
+            selected_scores = list(zip(self.references[references_idx_sorted],
+                                       self._scores[references_idx_sorted, selected_idx].copy()))
+        else:
+            selected_scores = list(zip(self.references, self._scores[:, selected_idx].copy()))
         return selected_scores
```
sverhoeven: Could be simplified to
```suggestion
            return list(zip(self.references[references_idx_sorted],
                                       self._scores[references_idx_sorted, selected_idx].copy()))
        return list(zip(self.references, self._scores[:, selected_idx].copy()))
```

matchms/similarity/BaseSimilarity.py
```diff
@@ -59,3 +62,21 @@ def matrix(self, references: List[SpectrumType], queries: List[SpectrumType],
...
+            1D Array of scores.
+
+        Returns
+        -------
+        idx_sorted
+            Indexes of sorted scores.
+        """
+        if scores.dtype.names is None:
+            return scores.argsort()[::-1]
+        return scores["score"].argsort()[::-1]
```
sverhoeven: Don't think the base class should look at the incoming dtype. It should sort according to the score_datatype. So it should only consist out of line 81. Line 82 is specific to the Cosine classes it should not be part of the base.

Classes which override score_datatype should have their own `sort()` if needed. The `sort() could be implemented in an intermediate abstract class or in the class itself.

readthedocs/index.rst
```diff
@@ -74,14 +74,14 @@ Below is a small example of using matchms to calculate the Cosine score between
...
+        (reference, query, score) = score
         # Ignore scores between same spectrum and
         # pairs which have less than 20 peaks in common
-        if reference is not query and n_matching >= 20:
+        if reference is not query and score['matches'] >= 20:
             print(f"Reference scan id: {reference.metadata['scans']}")
             print(f"Query scan id: {query.metadata['scans']}")
-            print(f"Score: {score:.4f}")
-            print(f"Number of matching peaks: {n_matching}")
+            print(f"Score: {score['score']:.4f}")
```
sverhoeven, line 83: https://github.com/matchms/matchms/blob/add_querying_options/README.rst#example should also use `score['score']` instead tuple splat.

sverhoeven (changes requested):
Overall looks good, docstrings render OK, additional tests result in nice coverage.
I like the approach of having a `BaseSimilarity.sort()` method. (I might have suggested it before, but good to see that it was also implementable in a unobtrusive way)

Calculation scores between 2 spectra is failing to return a structured array

I tried
```ipython
In [1]:         import numpy as np
   ...:         from matchms import calculate_scores
   ...:         from matchms import Spectrum
   ...:         from matchms.similarity import CosineGreedy
   ...: 
   ...:         spectrum_1 = Spectrum(mz=np.array([100, 150, 200.]),
   ...:                               intensities=np.array([0.7, 0.2, 0.1]),
   ...:                               metadata={'id': 'spectrum1'})
   ...:         spectrum_2 = Spectrum(mz=np.array([100, 140, 190.]),
   ...:                               intensities=np.array([0.4, 0.2, 0.1]),
   ...:                               metadata={'id': 'spectrum2'})

In [2]: similarity_measure = CosineGreedy()

In [3]: scores = calculate_scores([spectrum_1],[spectrum_2], similarity_measure)

In [6]: scores._scores
Out[6]: array([[(0.831479419283098, 1)]], dtype=object)

In [7]: list(scores)
Out[7]: 
[(<matchms.Spectrum.Spectrum at 0x7fb29ac050d0>,
  <matchms.Spectrum.Spectrum at 0x7fb29ac05070>,
  0.831479419283098,
  1)]

```

Instead of `0.831479419283098,  1` I expected to get back a dictionary or structured array.
Can you add a test for this use case?

integration-tests/test_user_workflow.py
florian-huber: Yes, I also wasted some time on this :D ...

matchms/Scores.py
florian-huber: Nice, thanks, looks much better this way (somehow I keep overlooking those things a bit...).

tests/test_scores.py
```diff
@@ -2,36 +2,41 @@
...
+    score_datatype = [("score", numpy.unicode_, 16), ("len", numpy.int32)]
+
     def __init__(self):
         """constructor"""
 
     def pair(self, reference, query):
         """necessary pair computation method"""
         s = reference + query
-        return s, len(s)
+        return numpy.array([(s, len(s))], dtype=self.score_datatype)
```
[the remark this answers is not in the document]
florian-huber: That is indeed more consistent. I added this to our scoring functions.

matchms/Scores.py
```diff
@@ -178,10 +185,9 @@ def scores_by_query(self, query: QueriesType) -> numpy.ndarray:
             references = [spectrum_1, spectrum_2, spectrum_3]
             queries = [spectrum_2, spectrum_3, spectrum_4]
 
-            scores = Scores(references, queries, CosineGreedy()).calculate()
-            selected_scores = scores.scores_by_query(spectrum_4)
-            selected_scores.sort(key=lambda s: s[1], reverse=True)
-            print([np.round(x[1], 3) for x in selected_scores])
+            scores = calculate_scores(references, queries, CosineGreedy())
+            selected_scores = scores.scores_by_query(spectrum_4, sort=True)
+            print([x[1]["score"].round(3) for x in selected_scores])
```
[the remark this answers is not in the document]
florian-huber: Great, thanks!

readthedocs/index.rst
florian-huber: Good catch. I changed the example.

tests/test_scores.py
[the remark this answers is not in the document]
florian-huber: Thanks Stefan. I now implemented the structured arrays for the pair methods as well.

matchms/similarity/BaseSimilarity.py
florian-huber: I reduced it to line 81 as suggested. In fact, that even works for all scores we implemented so far (without need to define own sort() method).

florian-huber:
Thanks a lot for the reviewing Stefan, that was very helpful! I believe I could address your comments and will merge this PR now.

PR #159, opened 2020-11-30
Improve cosine score performance

florian-huber:
- considerable performance gain by making better use of Numba and Numpy.
Main gain comes from a better search algorithm for matching peaks (``find_matches()`` using Numba and the fact that peaks are sorted).
For spectra with few peaks (<100) there is no huge difference, but the scaling changed quite drastically. Now it scales linear with the number of peaks, before it was polynomial. For instance for spectra with around 1000 peaks the new implementation is about 10x faster, for spectra with 2000 peaks about 20x faster.
- larger performance gain of ``CosineHungarian`` implementation due to faster implementation in scipy (>= 1.4.0) and the now better matching peak detection.

### DIFF
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -1,4 +1,5 @@
 from typing import Tuple
+import numpy
 from matchms.typing import SpectrumType
 from .BaseSimilarity import BaseSimilarity
 from .spectrum_similarity_functions import collect_peak_pairs
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -87,11 +87,11 @@ class CosineGreedy(BaseSimilarity):
             matching_pairs = collect_peak_pairs(spec1, spec2, self.tolerance,
                                                 shift=0.0, mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
-            if matching_pairs is not None:
-                matching_pairs = numpy.array(matching_pairs)
-                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
-                return matching_pairs
-            return None
+            if matching_pairs is None:
+                return None
+            matching_pairs = numpy.array(matching_pairs)
+            matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
+            return matching_pairs
 
         spec1 = get_peaks_array(reference)
         spec2 = get_peaks_array(query)
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -86,7 +86,8 @@ class CosineGreedy(BaseSimilarity):
             matching_pairs = collect_peak_pairs(spec1, spec2, self.tolerance, shift=0.0,
                                                 mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
-            matching_pairs = sorted(matching_pairs, key=lambda x: x[2], reverse=True)
+            if matching_pairs.shape[0] > 0:
+                matching_pairs = matching_pairs[np.argsort(matching_pairs[:,2])[::-1], :]
             return matching_pairs
 
         spec1 = get_peaks_array(reference)
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -94,5 +94,7 @@ class CosineGreedy(BaseSimilarity):
         spec1 = get_peaks_array(reference)
         spec2 = get_peaks_array(query)
         matching_pairs = get_matching_pairs()
-        return score_best_matches(matching_pairs, spec1, spec2,
-                                  self.mz_power, self.intensity_power)
+        if matching_pairs.shape[0] > 0:
+            return score_best_matches(matching_pairs, spec1, spec2,
+                                      self.mz_power, self.intensity_power)
+        return float(0), 0
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -87,7 +87,7 @@ class CosineGreedy(BaseSimilarity):
                                                 mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
             if matching_pairs.shape[0] > 0:
-                matching_pairs = matching_pairs[np.argsort(matching_pairs[:,2])[::-1], :]
+                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:,2])[::-1], :]
             return matching_pairs
 
         spec1 = get_peaks_array(reference)
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -84,17 +84,19 @@ class CosineGreedy(BaseSimilarity):
         """
         def get_matching_pairs():
             """Get pairs of peaks that match within the given tolerance."""
-            matching_pairs = collect_peak_pairs(spec1, spec2, self.tolerance, shift=0.0,
-                                                mz_power=self.mz_power,
+            matching_pairs = collect_peak_pairs(spec1, spec2, self.tolerance,
+                                                shift=0.0, mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
-            if matching_pairs.shape[0] > 0:
-                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
-            return matching_pairs
+            if matching_pairs is not None:
+                matching_pairs = numpy.array(matching_pairs)
+                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:,2])[::-1], :]
+                return matching_pairs
+            return None
 
         spec1 = get_peaks_array(reference)
         spec2 = get_peaks_array(query)
         matching_pairs = get_matching_pairs()
-        if matching_pairs.shape[0] > 0:
+        if matching_pairs is not None:
             return score_best_matches(matching_pairs, spec1, spec2,
                                       self.mz_power, self.intensity_power)
         return float(0), 0
--- a/matchms/similarity/CosineGreedy.py
+++ b/matchms/similarity/CosineGreedy.py
@@ -87,7 +87,7 @@ class CosineGreedy(BaseSimilarity):
                                                 mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
             if matching_pairs.shape[0] > 0:
-                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:,2])[::-1], :]
+                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
             return matching_pairs
 
         spec1 = get_peaks_array(reference)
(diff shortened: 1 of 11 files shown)


matchms/similarity/CosineHungarian.py
```diff
@@ -56,7 +56,11 @@ def get_matching_pairs():
             matching_pairs = collect_peak_pairs(spec1, spec2, self.tolerance, shift=0.0,
                                                 mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
-            return sorted(matching_pairs, key=lambda x: x[2], reverse=True)
+            if matching_pairs is not None:
+                matching_pairs = numpy.array(matching_pairs)
+                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
+                return matching_pairs
+            return None
```
sverhoeven: Same as in matchms/similarity/CosineGreedy.py can you do `if..return None`.

matchms/similarity/CosineGreedy.py
```diff
@@ -83,14 +84,19 @@ def pair(self, reference: SpectrumType, query: SpectrumType) -> Tuple[float, int
...
+            matching_pairs = collect_peak_pairs(spec1, spec2, self.tolerance,
+                                                shift=0.0, mz_power=self.mz_power,
                                                 intensity_power=self.intensity_power)
-            matching_pairs = sorted(matching_pairs, key=lambda x: x[2], reverse=True)
-            return matching_pairs
+            if matching_pairs is not None:
+                matching_pairs = numpy.array(matching_pairs)
+                matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
+                return matching_pairs
+            return None
```
sverhoeven: Can you bailout early, thats easier to read for me
```suggestion
            if matching_pairs is None:
                return None
            matching_pairs = numpy.array(matching_pairs)
            matching_pairs = matching_pairs[numpy.argsort(matching_pairs[:, 2])[::-1], :]
            return matching_pairs
```

matchms/similarity/CosineGreedy.py
```diff
@@ -83,14 +84,19 @@ def pair(self, reference: SpectrumType, query: SpectrumType) -> Tuple[float, int
...
 
         spec1 = get_peaks_array(reference)
         spec2 = get_peaks_array(query)
         matching_pairs = get_matching_pairs()
-        return score_best_matches(matching_pairs, spec1, spec2,
-                                  self.mz_power, self.intensity_power)
+        if matching_pairs is not None:
```
sverhoeven, line 98: Same as in get_matching_pairs() can you do `if..return float(0),0`.

matchms/similarity/CosineHungarian.py
```diff
@@ -69,15 +73,15 @@ def get_matching_pairs_matrix():
...
             matching_pairs_matrix = numpy.ones(matrix_size)
-            for match in matching_pairs:
-                matching_pairs_matrix[paired_peaks1.index(match[0]),
-                                      paired_peaks2.index(match[1])] = 1 - match[2]
+            for i in range(matching_pairs.shape[0]):
+                matching_pairs_matrix[paired_peaks1.index(matching_pairs[i, 0]),
+                                      paired_peaks2.index(matching_pairs[i, 1])] = 1 - matching_pairs[i, 2]
             return paired_peaks1, paired_peaks2, matching_pairs_matrix
 
         def solve_hungarian():
```
sverhoeven, line 86: Line below has typo, `agorithm` should be `algorithm`

matchms/similarity/ModifiedCosine.py
```diff
@@ -93,11 +94,20 @@ def get_matching_pairs():
...
 
         spec1 = get_peaks_array(reference)
         spec2 = get_peaks_array(query)
         matching_pairs = get_matching_pairs()
-        return score_best_matches(matching_pairs, spec1, spec2,
-                                  self.mz_power, self.intensity_power)
+        if matching_pairs.shape[0] > 0:
+            return score_best_matches(matching_pairs, spec1, spec2,
+                                      self.mz_power, self.intensity_power)
+        return float(0), 0
```
sverhoeven: Same as in matchms/similarity/CosineGreedy.py can you do `if..return float(0),0` followed with `return score_best_matches`.

matchms/similarity/spectrum_similarity_functions.py
```diff
@@ -5,69 +5,111 @@
...
-        mz = spec1[idx, 0]
-        matches = numpy.where((numpy.abs(spec2[:, 0] - spec1[idx, 0] + shift) <= tolerance))[0]
-        for match in matches:
-            power_prod_spec1 = ((mz ** mz_power) * (intensity ** intensity_power))
-            power_prod_spec2 = ((spec2[match][0] ** mz_power) * (spec2[match][1] ** intensity_power))
-            matching_pairs.append((idx, match, power_prod_spec1 * power_prod_spec2))
 
-    return matching_pairs
+@numba.njit
+def find_matches(spec1, spec2, tolerance, shift):
```
sverhoeven: Add typings
```suggestion
def find_matches(spec1: numpy.ndarray, spec2: numpy.ndarray, tolerance: float, shift: float = 0) -> List[Tuple[int, int]]:
```

matchms/similarity/spectrum_similarity_functions.py
```diff
@@ -5,69 +5,111 @@
...
+    Parameters
+    ----------
+    spec1:
+        Spectrum peaks and intensities as numpy array. Peak mz values must be ordered.
+    spec2:
+        Spectrum peaks and intensities as numpy array. Peak mz values must be ordered.
+    tolerance
+        Peaks will be considered a match when <= tolerance appart.
+    shift
+        Shift spectra peaks by shift. The default is 0.
```
sverhoeven: On line 78 you are shifting only the peaks in the second spectra. Can the docs reflect this?
```suggestion
        Shift peaks of second spectra by shift. The default is 0.
```

matchms/similarity/ModifiedCosine.py
```diff
@@ -93,11 +94,20 @@ def get_matching_pairs():
             nonzero_pairs = collect_peak_pairs(spec1, spec2, self.tolerance, shift=mass_shift,
                                                mz_power=self.mz_power,
                                                intensity_power=self.intensity_power)
-            unsorted_matching_pairs = zero_pairs + nonzero_pairs
-            return sorted(unsorted_matching_pairs, key=lambda x: x[2], reverse=True)
+
+            if zero_pairs is None:
```
sverhoeven, line 98: There is no test which resulted in zero_pairs to be filled.

matchms/similarity/spectrum_similarity_functions.py
```diff
@@ -5,69 +5,111 @@
...
-    score = 0.0
-    used_matches = []
-    for match in matching_pairs:
-        if not match[0] in used1 and not match[1] in used2:
-            score += match[2]
-            used1.add(match[0])  # Every peak can only be paired once
-            used2.add(match[1])  # Every peak can only be paired once
-            used_matches.append(match)
+    for i in range(matching_pairs.shape[0]):
+        if not matching_pairs[i, 0] in used1 and not matching_pairs[i, 1] in used2:
```
sverhoeven, line 106: In tests `if` was never false. Can a test be written for that case?

sverhoeven (approved):
Existing tests kept working so I expect that there are no changes in the scores.
Nice to see even more tests and to see that newer dependency makes things faster.
The readthedocs still renders OK.

Some inline suggestions for typings, branch coverage and `if ... return` style.

matchms/similarity/spectrum_similarity_functions.py
```diff
@@ -5,69 +5,111 @@
...
         case the peak intensity products will not depend on the m/z ratios.
-    intensity_power: float, optional
+    intensity_power:
         The power to raise intensity to in the cosine function. The default is 1.
 
     Returns
     -------
-    matching_pairs : list
-        List of found matching peaks.
+    matching_pairs : numpy array
```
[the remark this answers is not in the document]
florian-huber: Unfortunately Numba complained when I did ``matching_pairs = numpy.array(matching_pairs)`` but I will check again.

matchms/similarity/CosineGreedy.py
florian-huber: Good point. Looks better that way.

matchms/similarity/spectrum_similarity_functions.py
florian-huber: Thanks!

matchms/similarity/spectrum_similarity_functions.py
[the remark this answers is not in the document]
sverhoeven: Ah, numba probably does not like a var changing types, using different var name might help

matchms/similarity/CosineHungarian.py
florian-huber: done.

matchms/similarity/CosineGreedy.py
florian-huber: Thanks. Changed this.

matchms/similarity/CosineHungarian.py
florian-huber: done.

matchms/similarity/spectrum_similarity_functions.py
florian-huber: Good point, was added.

matchms/similarity/ModifiedCosine.py
florian-huber: Done.

matchms/similarity/ModifiedCosine.py
florian-huber: Weird.  that case will occur in the tests, not sure why sonarcloud does not recognize it though.

matchms/similarity/spectrum_similarity_functions.py
florian-huber: Same here, that case will occur in the tests.

matchms/similarity/ModifiedCosine.py
sverhoeven: In noticed them when I run `pytest --cov --cov-report html` then there are yellow lines in htmlcov/*html. Maybe you can only run the test you expect to cover that line to see what the coverage looks like.

On the other hand having 99% coverage is very good already, we can also let it go.

matchms/similarity/spectrum_similarity_functions.py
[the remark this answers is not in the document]
florian-huber: I think it is rather something with the different types what are within the list... but not sure.
The error is ``TypingError: Invalid use of Function(<intrinsic range_iter_len>) with argument(s) of type(s): (enumerate(iter(list(int64))))``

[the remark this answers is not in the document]
florian-huber: OK, seems to work if I do ``return numpy.array(matching_pairs.copy())``.

matchms/similarity/ModifiedCosine.py
florian-huber: I would indeed leave it at that for now.

matchms/similarity/spectrum_similarity_functions.py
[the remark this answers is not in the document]
florian-huber: I changed this and removed the now redundant ``matching_pairs = numpy.array(matching_pairs)`` lines.

florian-huber:
Thanks a lot for the great review Stefan, very much appreciated!

PR #167, opened 2021-01-04
refactor CI workflow

### LINKED ISSUE #150: restructure CI workflows
With the new pypi workflows (#148) we could maybe restructure the workflows a bit.
The pip workflows are much faster to run. So I would suggest we move the ``pylint``, ``isort``, and ``prospector`` things to that workflow and only run the slower conda ones if those pass?

florian-huber:
Addresses #150 
(I mostly adapted the refactored CI workflow from Spec2Vec)


.github/workflows/CI_build.yml
```diff
@@ -7,35 +7,113 @@ on:
...
+        run: isort --check-only --diff --conda-env matchms-dev --recursive --lines-after-imports 2 --force-single-line --no-lines-before FUTURE --no-lines-before STDLIB --no-lines-before THIRDPARTY --no-lines-before FIRSTPARTY --no-lines-before LOCALFOLDER matchms/ tests/ integration-tests/
+      - name: SonarCloud Scan
+        if: github.repository == 'matchms/matchms'
+        uses: sonarsource/sonarcloud-github-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
+  build_pypi:
+    name: Pypi and documentation build / python-${{ matrix.python-version }} / ${{ matrix.os }}
```
sverhoeven: Documentation is not tested here
```suggestion
    name: Pypi / python-${{ matrix.python-version }} / ${{ matrix.os }}
```

.github/workflows/CI_build.yml
```diff
@@ -45,71 +123,39 @@ jobs:
...
         shell: bash -l {0}
         run: |
           make coverage doctest html
         working-directory: readthedocs/
         env:
           SPHINXOPTS: "-n"  # enable nit-picky mode
       - name: Check documentation coverage threshold
-        if: matrix.python-version == '3.7'
-        shell: bash -l {0}
+        if: matrix.os == 'ubuntu-latest'
```
sverhoeven: No matrix in job so this line can be removed.
```suggestion
```

sverhoeven (changes requested):
Job graph looks good. Will definitely speed up, by catching most common mistakes in first job.

The tests requiring rdkit (`tests/test_utils.py`) are not tested. Could you add a pytest to `anaconda_dev` job. 

Also the SonarCloud analysis will miss the coverage of those tests. To get a bit higher coverage you could run the analysis in the `anaconda_dev` job. As coverage is already high I don't think it is absolutely needed.

florian-huber:
Thanks Stefan, good point. I to hastily adapted the `Spec2Vec` refactored workflow for this here and then missed the `rdkit` issue.
I now added another test and sonarcloud run to the `anaconda_dev` run.

sverhoeven (approved):

florian-huber:
Thanks a lot for reviewing Stefan!

Issue #71, opened 2020-06-08
Add consistent logging


florian-huber:
Currently several filters use print statements when they change metadata.
As Stefan mentioned (in PR #34, https://github.com/matchms/matchms/pull/34#issuecomment-640434174), it would probably be cleaner to do this via logging.

- replace print statements by logging
- allow logging to file (if you run many filters over a lot of data it would be nice to get a report of the done changes)

florian-huber:
Another solution could also be to use a `verbose=0,1,2` setting, similar to gensim, which would define the level of print statements that will be called. (logging seems cleaner though)

florian-huber:
Started playing with the logging (PR #175 ).
Some TODO's from that start:
- add own matchms logger
- add example (or wrapper) to write matchms logging to file
- switch off rdkit logging which for our use case is too low level (could also be made optional I guess). --> https://github.com/rdkit/rdkit/issues/2683

florian-huber:
Hi @hechth, let's continue the logging discussion here.

> @florian-huber Then now might be a good time! Do you already have some preferred python logger in mind?

So far, I was simply thinking of the basic `logging`, something like:
```
import logging
log = logging.getLogger(__name__)
```

PR #175, opened 2021-01-29
Implement better logging

florian-huber:
Start addressing #71


florian-huber:
outdated, replaced by #271

Issue #178, opened 2021-02-19
Fix bioconda recipe


florian-huber:
Nice to see that `matchms` has been added to Bioconda (https://github.com/bioconda/bioconda-recipes/pull/25425)!
Thanks @martenson.

The recipe takes the package from pypi's `.tar.gz` which we just realized does not contain the package data.
--> @sverhoeven suggested removing the `include_package_data=True` from `setup.py`

In addition I started a PR to update the dependencies: https://github.com/bioconda/bioconda-recipes/pull/26909

florian-huber:
Once done, update badge (see https://anaconda.org/bioconda/matchms/badges).

florian-huber:
With #179 the package_data is now in the bioconda package and it no longer fails.
I started a PR with bioconda to update the dependencies, but tests there are currently still failing...
See: https://github.com/bioconda/bioconda-recipes/pull/26909

florian-huber:
Recipe changes for bioconda have been merged (https://github.com/bioconda/bioconda-recipes/pull/26909). 
That should correctly add rdkit.
However, for `matchms 0.8.1` there is now two builds (https://anaconda.org/bioconda/matchms/files), so running
```
conda create --name matchms python=3.8
conda activate matchms
conda install --channel bioconda matchms
```
...in my case used the build number 0 (without rdkit). This should be fixed with the next release I guess.

sverhoeven:
I am trying to install the `matchms-0.8.1-pyh3252c3a_1.tar.bz2` from bioconda, but it is failing with

```shell
conda install -c bioconda matchms=0.8.1=pyh3252c3a_1
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: - 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed

ResolvePackageNotFound: 
  - rdkit[version='>=2020.03.1']

```

I guess that is why it is installing build number=0.

Looking at https://anaconda.org/bioconda/rdkit/files there has not been an update since 4 years and is missing version 2020.03.1, while on conda-forge rdkit is much more recent.
Using `conda install -c conda-forge  -c bioconda matchms` made it install build=1.

[...]

PR #179, opened 2021-02-19
Update setup.py

florian-huber:
Addressing part of #178


florian-huber (commented):
I checked with `python setup.py sdist` and the package data was included in the `.tar.gz` file.

PR #181, opened 2021-02-24
convert string charges to int during filtering

davidwhealey:
Some json spectra contain string-formatted charges.  This pull request adds a filter for string-formatted charges to filtering.make_charge_scalar


florian-huber:
Thanks for the pull request @davidwhealey !
Currently the tests fail because some tests do not run properly from a fork (see #182). So, for now I would move your pull request to a branch within matchms and then we can work on it from there.

PR #191, opened 2021-03-03
Added require_precursor_mz_above_0 and tests

niekdejonge:


matchms/filtering/require_precursor_mz_above_0.py
```diff
@@ -0,0 +1,29 @@
...
+        Input spectrum.
+    """
+    if spectrum_in is None:
+        return None
+
+    spectrum = spectrum_in.clone()
+
+    precursor_mz = spectrum.get("precursor_mz", None)
+    if precursor_mz is None:
+        return None
```
florian-huber:
```suggestion
    if precursor_mz is None:
        pepmass = spectrum.get("pepmass", None)
        assert pepmass is None or not isinstance(pepmass[0], (float, int)), \
            "Found 'pepmass' but no 'precursor_mz'. Consider applying 'add_precursor_mz' filter first."
        return None
```

florian-huber: This is only a safety check in case people are note aware of the `add_precursor_mz` step which should come before running your new filter function.

matchms/filtering/require_precursor_mz_above_0.py
```diff
@@ -0,0 +1,29 @@
+from ..typing import SpectrumType
+from ..typing import Union
+
+
+def require_precursor_mz_above_0(spectrum_in: SpectrumType
+                                 ) -> Union[SpectrumType, None]:
+
+    """Returns None if there is no precursor_mz or if <=0
```
florian-huber:
```suggestion
def require_precursor_mz(spectrum_in: SpectrumType
                         ) -> Union[SpectrumType, None]:

    """Returns None if spectrum has no valid precursor_mz.
```

florian-huber: I think we can leave out the `_above_0` part, but that would also have to be added to the rest of your code (tests...).

matchms/similarity/ModifiedCosine.py
```diff
@@ -90,8 +90,15 @@ def get_matching_pairs():
...
                                             mz_power=self.mz_power,
                                             intensity_power=self.intensity_power)
-            message = "Precursor_mz missing. Apply 'add_precursor_mz' filter first."
-            assert reference.get("precursor_mz") and query.get("precursor_mz"), message
+            message_precursor_missing = \
+                "Precursor_mz missing. Apply 'add_precursor_mz' filter first."
+            assert reference.get("precursor_mz") \
+                   and query.get("precursor_mz"), message_precursor_missing
+            message_precursor_below_0 = "Precursor_mz has to be higher than " \
+                                        "0, apply require_precursor_mz first"
```
florian-huber:
```suggestion
            message_precursor_below_0 = "Expect precursor_mz to be positive number." \
                                        "Apply 'require_precursor_mz' first."
```

florian-huber (changes requested):
Nice work Niek!
Also very good extensive testing!

I added some small comments.
In addition, can you add the changes to `CHANGELOG.md` (let me know if that is unclear)? 
Finally, also don't forget to add yourself to `CITATION.cff` and `.zenodo.json` ;)

matchms/filtering/require_precursor_mz.py
```diff
@@ -0,0 +1,34 @@
+from ..typing import SpectrumType
+from ..typing import Union
```
florian-huber:
```suggestion
from typing import Union
from ..typing import SpectrumType
```

florian-huber: bit confusing, but one is the matchms.typing and one is the general typing module

niekdejonge: Yes of course, well spotted :)

florian-huber (approved):

Issue #66, opened 2020-06-03
Install instructions on anaconda don't work


sverhoeven:
The install instructions on https://anaconda.org/nlesc/matchms don't work.

Trying

```shell
conda install -c nlesc matchms
```

It is unable to find the dependencies.

The install command `conda install --channel nlesc --channel bioconda --channel conda-forge matchms` in the README.md does work.

We should somehow make sure the install instructions on https://anaconda.org/nlesc/matchms work or are corrected in the description.

florian-huber:
With the addition of matchms to Bioconda (#178 ), the following will work:
```
conda install -c bioconda matchms
```

florian-huber:
We will then have matchms in `bioconda` and in the `nlesc` channel... should we then remove the 2nd?

sverhoeven:
Doh, Instructions on https://anaconda.org/bioconda/matchms also don't because rdkit must come from conda-forge channel.

Yes, lets drop nlesc channel
1. Remove all the anaconda build stuff from CI workflows and docs as building anaconda package is now handled by external bioconda recipe.
2. Add CI workflow where just rdkit is installed with conda and runs inchi/smiles tests
2. Only mention `pip` or `conda install -c conda-forge -c bioconda matchms` as installation methods.
3. Keep old versions of matchms in Anaconda NLeSC channel so spec2vec can gradually switch to bioconda.

florian-huber:
Sounds like a plan!
(and I guess you meant `conda install -c conda-forge -c bioconda matchms`)

PR #193, opened 2021-03-05
Update conda_publish.yml

florian-huber:
See also #66 

@sverhoeven I don't know how to trigger this. I now added a cron job (running 2 times per week). Having it triggered by a new Bioconda matchms release would of course be better, but I have no clue how that would be done. Let me know if you have an idea on this!

### DIFF
--- a/.github/workflows/conda_publish.yml
+++ b/.github/workflows/conda_publish.yml
@@ -1,49 +1,10 @@
-name: Anaconda Publish
+name: Verify Bioconda Publish
 
 on:
-  release:
-    types: [published]
+  schedule:
+    -cron: '23 23 * * 0,4'
 
 jobs:
-
-  publish:
-    name: publish / python-3.7 / ubuntu-latest
-    runs-on: 'ubuntu-latest'
-    steps:
-      - uses: actions/checkout@v2
-      - uses: conda-incubator/setup-miniconda@v2
-        with:
-          activate-environment: matchms-build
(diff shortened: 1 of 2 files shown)


.github/workflows/conda_publish.yml
```diff
@@ -1,50 +1,11 @@
-name: Anaconda Publish
+name: Verify Bioconda Publish
 
 on:
-  release:
-    types: [published]
+  schedule:
+    -cron: '23 23 * * 0,4'
```
sverhoeven: It should be run when after matchms is updated on bioconda. So least worse is to use [manual trigger](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow).
```suggestion
  workflow_dispatch:
```

Doing it `At 23:23 on Sunday and Thursday.` seems a bit much.

sverhoeven (approved):
Whats remains of the workflow should work.
I would suggest to trigger the workflow manually.

We should add the instructions for doing bioconda bumpversion and running this workflow to the [make new release docs](https://github.com/matchms/matchms/blob/master/CONTRIBUTING.md#you-want-to-make-a-new-release-of-the-code-base).

PS. Looking at bioconda docs I also saw matchms has also been Dockerized at https://biocontainers.pro/tools/matchms

florian-huber:
Thanks a lot Stefan for the review. 
I also added the manual workflow mention to `CONTRIBUTING.md`.

PR #211, opened 2021-03-18
Changed find_matches function to only take m/z values

### LINKED ISSUE #208: find_matches function expects 2d array with m/z and intensity, but uses only m/z
**Describe the bug**
The function is being given/passed information which is not necessary and used.
See https://github.com/matchms/matchms/blob/5cf15bcc89128f1bffacd62a8688062edfd8d06e/matchms/similarity/spectrum_similarity_functions.py#L50-87

**Expected behavior**
The function should take only the m/z values of the peaks, or the spectrum objects.
It could also become a class function of the spectrum, which would then ensure the assumption that the arrays are sorted.

**Additional context**
This would make the code more easy to read and concise, given that the Spectrum class already has a sorted list of only m/z values.

hechth:
Renamed parameters to be only m/z values from spectra. Changed handling to expect 1d arrays.

Adapted test cases to work with new format and renamed corresponding variables.

Fixes #208

### DIFF
--- a/matchms/similarity/spectrum_similarity_functions.py
+++ b/matchms/similarity/spectrum_similarity_functions.py
@@ -33,7 +33,7 @@ def collect_peak_pairs(spec1: numpy.ndarray, spec2: numpy.ndarray,
     matching_pairs : numpy array
         Array of found matching peaks.
     """
-    matches = find_matches(spec1, spec2, tolerance, shift)
+    matches = find_matches(spec1[:, 0], spec2[:, 0], tolerance, shift)
     idx1 = [x[0] for x in matches]
     idx2 = [x[1] for x in matches]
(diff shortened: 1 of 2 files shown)


florian-huber (approved):
Thanks for the pull request. Looks very good so can be merged!

One sidenote: the tests failed because it was running from your fork which gives `Sonarcloud` no access to the required token (in secrets of the repo). For larger pull requests you can maybe better push to a new branch within matchms so that all tests can run.

hechth:
@florian-huber Thank you very much - I can post the PR again from a different branch inside matchms in order to run the tests. If they all succeed I would merge it.

florian-huber:
Sounds perfect @hechth! If tests pass, you can simply go ahead and merge it.

PR #213, opened 2021-03-25
add negative ionmode adducts

florian-huber:
I added 3 negative ionmode adducts which I encountered several times when working with GNPS data to the table of known adducts.


florian-huber:
@justinjjvanderhooft, I added 3 negative ionmode adducts that I frequently encountered in the dataset. To keep good practices (4 eyes before we merge it 😄 ) --> can you briefly check they are OK?

Obviously the list is still not exhaustive, but we can always add more later. Or people can suggest additional ones if they encounter them.

justinjjvanderhooft:
@florian-huber I can see the ones you added, but could you indicate how you annotated them? That is not completely clear to me from the code lines. I.e., what is "Ser"? And what is the source of the annotations?

florian-huber:
Good question Justin!
The `[M-H]-/[M-Ser]-` was just added because we had `[M-H-/M-Ser]-` on the list. That, however, doesn't answer the question of what `Ser` is... but there were quiet some of those in the all-GNPS dataset.
We could also remove these two "known adducts" and add a conversion rule that will simply take the first `[M-H]-` part into account?

justinjjvanderhooft:
So am I right that these added annotations came from the list of GNPS with higher-frequency adduct annotations? In that case, it makes sense to add them but only if we understand their origin. I mean, Ser could refer to Serine, but I am not quite sure that this is a thing in mass spec (I haven't seen it yet....)

florian-huber:
Just send out a tweet to see if some mass-spec people respond and point us to something for `[M-Ser]-`...

florian-huber:
OK. I now removed the `Ser` adducts from the known adduct table (seems they are not really adducts).
I added them to the list of our conversion rules so that `[M-H]-/[M-Ser]- --> [M-H]-` and `[M-H-/M-Ser]- --> [M-H]-`

florian-huber:
@justinjjvanderhooft I will merge the changes for now (see above comment). But we can always return to this issue and update our lists with some community help!

PR #218, opened 2021-04-07
add `.to_numpy` method

### LINKED ISSUE #214: Move get_peaks_array function to utils.py or to Spectrum class
The function should be moved to general utility since it also makes sense in other places, for example, msp export.

https://github.com/matchms/matchms/blob/5497b7363af1d555b8d1ab0162e01458ac7fbf53/matchms/similarity/spectrum_similarity_functions.py#L90-L92

florian-huber:
Addressing #214


hechth (approved):
Maybe the name 'as_numpy' would be a tiny bit more fitting as it is a property and not a method - but this is very much nitpicking :D Thank you very much!

PR #223, opened 2021-04-13
Implemented reading additional metadata keys into 'precursor_mz' field using filter.

### LINKED ISSUE #222: Add more possible field names to add precursor mz values from metadata
**Problem**
The code below currently only allows precursor m/z related data in `pepmass` or `precursor_mz` while fields such as `percursormz` etc. are ignored.

**Solution**
Expand the dictionary of allowed keys that can be used to store the precursor m/z values in the file's metadata.

https://github.com/matchms/matchms/blob/04af51cec2eedf6ca2c398b44b614e387e309f16/matchms/filtering/add_precursor_mz.py#L4-L24

hechth:
Added `precursormz` and `precursor_mass` to the list of fields which are parsed into the `precursor_mz` field using the `add_precursor_mz` filter and did some minor code refactoring.

Introduced small utility function to get the first common element in two iterables to find the first key contained in the list of possible keys and metadata keys.

Fixes #222


tests/test_add_precursor_mz.py
```diff
@@ -45,18 +46,26 @@ def test_add_precursor_mz_only_pepmass_present():
...
 
-def test_add_precursor_mz_no_precursor_mz():
+@pytest.mark.parametrize("key, value, expected", [
+    ["precursor_mz", "444.0", 444.0],
+    ["precursormz", "15.6", 15.6],
+    ["precursormz", 15.0, 15.0],
+    ["precursor_mass", "17.887654", 17.887654],
+    ["pepmass", (33.89, 50), 33.89],
+    ["pepmass", "None", None],
+    ["pepmass", None, None]])
```
florian-huber: Nice expansion of the tests!

matchms/filtering/add_precursor_mz.py
```diff
@@ -12,10 +25,16 @@ def add_precursor_mz(spectrum_in: SpectrumType) -> SpectrumType:
...
 
-    if isinstance(spectrum.get("precursor_mz", None), str):
-        spectrum.set("precursor_mz", float(spectrum.get("precursor_mz").strip()))
-    elif spectrum.get("precursor_mz", None) is None:
-        pepmass = spectrum.get("pepmass", None)
+    precursor_mz_key = get_first_common_element(spectrum.metadata.keys(), _accepted_keys)
+    precursor_mz = spectrum.get(precursor_mz_key)
+
+    if isinstance(precursor_mz, str):
+        precursor_mz = float(precursor_mz.strip())
```
florian-huber: We will always keep some non-covered corner cases I guess. Currently it will fail for string entries such as `"n/a"`. Not sure we have to cover that.
(feel free to skip this one until someone really has this issue)

florian-huber (approved):
Really nice refactoring and expansion to other keys. Really good work @hechth !
(I left one minor comment but I am not even sure this is a real issue, so feel free to skip)

matchms/filtering/add_precursor_mz.py
hechth: I implemented some basic checking if the string can be converted, this should cover most non-sensical values and it also prints an error message.

PR #226, opened 2021-04-14
remove slow loop

florian-huber:
- remove slow for loop


florian-huber (commented):
I consider this a minor (and straighforward) change, so I simply did self-review here...

PR #228, opened 2021-05-23
Fix quote issue from code example in readme.

tshauck:
Trying do use double quotes in an f-string using double quotes causes an error.


florian-huber (approved):
Thanks a lot @tshauck for spotting & fixing this code example!

PR #230, opened 2021-05-24
Made MSP export format compatible with MS-FINDER & MS-DIAL

hechth:
Changed the 'NUM PEAKS' metadata key as being the last one exported before the actual peak data.
This ensures that the peak data is loaded correctly in tools like MS-FINDER and MS-DIAL.

Fixes #229

PR #231, opened 2021-05-26
Update CI_build.yml

florian-huber:


florian-huber:
Hi @sverhoeven !
I here tried a few things to fix the failing conda verify tests in #230, but with no success.
Do you have any further ideas how to solve this --> #232 ?

sverhoeven:
I will have a look at it on 10 Jun 2021

sverhoeven:
I would suggest we to remove the conda building workflows from this repo as use https://anaconda.org/bioconda/matchms and their [build system](https://bioconda.github.io/contributor/build-system.html) for building and verifying.

There are some things we can do to make the bioconda package more integrated
1.  ~When doing a GH release/PyPI release then trigger a version update in bioconda using https://bioconda.github.io/contributor/updating.html#updating-recipes~ seems to already be working, see https://github.com/bioconda/bioconda-recipes/pull/28382
2. ~Adding more tests to https://github.com/bioconda/bioconda-recipes/blob/master/recipes/matchms/meta.yaml~  [bioconda](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/matchms/meta.yaml) is recommending not to run unit tests on their system

From `.github/workflows/CI_build.yml` I think we should keep the `anaconda_dev` job, but remove `anaconda_build` and `anaconda_verify` jobs.
We keep `.github/workflows/conda_publish.yml` so we can run the matchms test suite using the bioconda package.

florian-huber:
Outdated --> will be closed

PR #236, opened 2021-06-16
Fix derive parent_mass, when charge = 0

niekdejonge:
When the charge is set at None in the metadata and no ioncharge is given. The charge will be set to 0.
In add parent_mass the charge is used to derive the parent mass from the precursor_mz and the charge. If the charge is set to 0, this results in a parent mass of 0.0, which is unexpected behavior. 
This pull request adds a quick fix, since it will now print the warning "Not sufficient spectrum metadata to derive parent mass." if charge = 0. 

Alternative might be automatically using a charge of 1, when the charge is set to 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


florian-huber (approved):
I added a test for charge=0. If the test pass it should be fine to merge...

Issue #237, opened 2021-06-28
Developer setup


hechth:
**Describe the bug**
Setting up the conda environment as described [here](https://github.com/matchms/matchms#documentation-for-developers) causes an error.

**To Reproduce**
Follow the developer documentation.

**Expected behavior**
All dependencies for development should be installed.

**Screenshots**

**Desktop (please complete the following information):**
 - Windows

**Additional context**
Error message:
```
(matchms-dev) C:\Users\473355\git\hechth\matchms>pip3 install --editable .[dev]  
Traceback (most recent call last):
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\Scripts\pip3-script.py", line 5, in <module>
    from pip._internal.cli.main import main
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\cli\main.py", line 9, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\cli\autocompletion.py", line 10, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\cli\main_parser.py", line 8, in <module>
    from pip._internal.cli import cmdoptions
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\cli\cmdoptions.py", line 23, in <module>
    from pip._internal.cli.parser import ConfigOptionParser
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\cli\parser.py", line 12, in <module>
    from pip._internal.configuration import Configuration, ConfigurationError
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\configuration.py", line 21, in <module>
    from pip._internal.exceptions import (
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_internal\exceptions.py", line 8, in <module>
    from pip._vendor.requests.models import Request, Response
  File "C:\Users\473355\Miniconda3\envs\matchms-dev\lib\site-packages\pip\_vendor\requests\__init__.py", line 43, in <module>
    from pip._vendor import urllib3
[...]

PR #238, opened 2021-06-29
Create CI for [dev] environment

florian-huber:
Test if developer instructions in readme work on different platforms.
(See #237)

Depending on merge order with #240, `Python 3.9` must also be added to CI matrix for [dev] test.


sverhoeven (approved):
Good work, this PR works, but can be improved.

The README also includes instructions for non-conda installation. Dont think OS Python is that different from conda Python. So dont think non-conda Python checks are needed.

The workflow uses bash for each os as its needed for the setup-miniconda action. Bash might hide problems that a Windows users could have. For example slashes in paths or env vars. So having a dedicated Windows job which uses cmd or powershell as shell might help catch those problems.

My suggestions to get the quickest and minimal amount of jobs would be:
1. Remove the matrix from .github/workflows/CI_build.yml:build_pypi job, as only ubuntu-latest and py37 is used in .github/workflows/CI_publish_pypi.yml workflow
2. Move `.github/workflows/CI_dev_environment.yml:test_dev` job to .github/workflows/CI_build.yml and make it depend on `first_check` job with `needs: first_check`. So latest ubuntu/latest python is always tested first.

florian-huber:
Thanks a lot for reviewing Stefan!
I added the suggested changes to the CI.

PR #239, opened 2021-06-29
add Python 3.9 support

florian-huber:
Address (part of) #240.

### DIFF
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -50,7 +50,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.7', '3.8']
+        python-version: ['3.7', '3.8', '3.9']
         exclude:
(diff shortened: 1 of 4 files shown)


sverhoeven (changes requested):
Looks good. 

However I would also use python 3.9 in the .github/workflows/CI_build.yml:first_check job and 3.9 in test_with_conda job.

And I would also add 3.9 to classifiers in setup.py, see https://github.com/matchms/matchms/blob/4c7de6bf9586b8092a62e952731b0a009a6e67f7/setup.py#L44

florian-huber:
Good points Stefan.
I added it to `setup.py`. But I ran into issues with `typing` hints between 3.8 and 3.9, so I left the pylint testing with 3.8.

sverhoeven (approved):
Hmm, typing errors seems to be false positive of old pinned pylint version in prospector, see https://github.com/PyCQA/prospector/issues/418

florian-huber:
Thanks for reviewing Stefan!

PR #242, opened 2021-07-04
proposed filtering modifications

adelabriere:
Hi,

I was trying to train a spec2vec model using some preprocessing steps.

My mgf does not include an adduct charge or parent_mass field, but it includes an a PEPMASS field and therefore I can use add_mz_precursor to fill the mz_precusor field. The reduce_to_number_of_peaks function is not working however as the resulting spectrum does not have a parent_mass field.

Looking at the implementation I don t see a reason why this filter can't work with a mz_precusor only, I therefore propose to include it.

I hope it makes sense,

Alexis

### DIFF
--- a/matchms/filtering/reduce_to_number_of_peaks.py
+++ b/matchms/filtering/reduce_to_number_of_peaks.py
@@ -26,6 +26,8 @@ def reduce_to_number_of_peaks(spectrum_in: SpectrumType, n_required: int = 1, n_
     """
     def _set_maximum_number_of_peaks_to_keep():
         parent_mass = spectrum.get("parent_mass", None)
+        if parent_mass is None:
+            parent_mass = spectrum.get("mz_precursor", None)
         if parent_mass and ratio_desired:
             n_desired_by_mass = int(ceil(ratio_desired * parent_mass))
             return min(max(n_required, n_desired_by_mass), n_max)


florian-huber:
Hi @adelabriere,
thanks for making this issue.

First I would need to understand a bit better what part of  `reduce_to_number_of_peaks` you want to use, I think.
Currently it can be use two ways to reduce the number of peaks:
1. if `ratio_desired` is None --> keep no more than  `n_max` highest intensity peaks.
2. if `ratio_desired` is set to a float value --> scale the maximum number of peaks to keep based on `parent_mass`.

For (1) it should already work with parent mass, so I guess you mean to do (2) but then based on mz_precusor instead of parent_mass, right?

I principle that could be added. My only worry here is that the original idea was that larger compounds (= larger parent mass) would on average produce a higher number of "meaningful" peaks. The precursor m/z in many cases would reflect the same here, but for charges different than +/-1 that will not be correct.

Let me know if you think this is still usefull to have!
If so I would still add an additional flag/option to the function (e.g. `accept_precursor_mz=False/True`).

---
Otherwise as a hack, for now, you could run:
```python
spectrums = [s.set("parent_mass", s.get("precursor_mz")) for s in spectrums]
```
which will simply add the precursor_mz as parent_mass. Not a particularly nice/clean thing to do though.

adelabriere:
Hi, thanks for your time, both filtering approaches make sense, I planned to test both actually as I never used the second version and I don t know if it is useful on our data.

> I principle that could be added. My only worry here is that the original idea was that larger compounds (= larger parent mass) would on average produce a higher number of "meaningful" peaks. The precursor m/z in many cases would reflect the same here, but for charges different than +/-1 that will not be correct.

I agree, however I think multiply charged MS-MS are actually quite rare in metabolomics, and are anyway never matched by any similarity measure (maybe your package, sorry I did not test it on that :S). Therefore I would argue that having this filter working "in all cases" would be nice, given than in the vast majority of the case, the parent mass and the precursor mass are close.

Or I would change the spec2vec example as the vast majority of MS-MS spectra won t have the required fields to compute parent_mass but all of them will have the PEPMASS field to compute the mz_precursor.

niekdejonge:
Hi, I did run into a similar problem and would like to propose a slightly different solution. 

Matchms already has the function add_parent_mass, however this will not add a parent mass if the charge is not specified. 
Since in most cases people do not work with multiply charged MS-MS spectra, like you mentioned. 
I think it would be a good suggestion to still derive the parent mass from the mz_precursor, if the charge is not specified. 

Instead of adding this in set_maximum_number_of_peaks_to_keep, I would like to suggest adding this option to add_parent_mass. So apply the following workflow: Use charge if available, else assume charge == 1. 

@florian-huber, do you agree? 

I created a seperate pull request, with this proposed change #252

florian-huber:
This should be mostly fixed with #252 .
If you now run the `default_filter()` (or `derive_ionmode()`  and `correct_charge()`) followed by `add_parent_masss()`, than this problem should be gone. 
Please let me know (or re-open this) if that's not the case. Thanks!

PR #243, opened 2021-07-05
add networking to __init__

florian-huber:


florian-huber:
Seems a minor, uncritical change so I will simply merge it.

PR #252, opened 2021-10-13
Derive parent mass from mz precursor if charge is unknown

niekdejonge:
See #242, for additional discussion on this topic. 

Matchms already has the function add_parent_mass, however this will not add a parent mass if the charge is not specified.
Since in most cases people do not work with multiply charged MS-MS spectra.
I think it would be a good suggestion to still derive the parent mass from the mz_precursor, if the charge is not specified.
So apply the following workflow: Use charge if available, else assume charge == 1.

When creating this pull request, I noticed that the charge also changes if the PROTON_MASS is added or removed. Since, using negative mode is actually getting more common, this would give issues in this case. Do you maybe know of some other info field of information that can be used to differ between positive and negative mode?

I think it will be important to change this, since in all 3 case study datasets I used, the charge (and parent mass) was not given. So I assume that this will be a very common problem for people that want to use matchms. 

If there is no good solution maybe add a comment in the printed message when parent mass cannot be derived to run 
spectrums = [s.set("parent_mass", s.get("precursor_mz")) for s in spectrums] to set the parent mass (or alternatively the charge).

### DIFF
--- a/matchms/filtering/add_parent_mass.py
+++ b/matchms/filtering/add_parent_mass.py
@@ -32,36 +32,38 @@ def add_parent_mass(spectrum_in: SpectrumType, estimate_from_adduct: bool = True
     spectrum = spectrum_in.clone()
     adducts_dict = load_adducts_dict()
 
-    if spectrum.get("parent_mass", None) is None or overwrite_existing_entry:
-        parent_mass = None
-        charge = spectrum.get("charge")
-        adduct = clean_adduct(spectrum.get("adduct"))
-        precursor_mz = spectrum.get("precursor_mz", None)
-        if precursor_mz is None:
-            print("Missing precursor m/z to derive parent mass.")
-            return spectrum
+    if spectrum.get("parent_mass", None) and not overwrite_existing_entry:
+        return spectrum
 
-        if estimate_from_adduct and adduct in adducts_dict:
-            multiplier = adducts_dict[adduct]["mass_multiplier"]
-            correction_mass = adducts_dict[adduct]["correction_mass"]
-            parent_mass = precursor_mz * multiplier - correction_mass
+    parent_mass = None
+    charge = spectrum.get("charge")
+    adduct = clean_adduct(spectrum.get("adduct"))
+    precursor_mz = spectrum.get("precursor_mz", None)
+    if precursor_mz is None:
+        print("Missing precursor m/z to derive parent mass.")
+        return spectrum
 
-        if parent_mass is None and charge is not None and charge != 0:
-            # Otherwise assume adduct of shape [M+xH] or [M-xH]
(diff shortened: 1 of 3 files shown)


florian-huber:
Hi @niekdejonge ... I missed to follow up on this. What is the status? Is this still needed/ongoing?

niekdejonge:
Hi @florian-huber. I currently solved this in the MS2Query software, by automatically adding a charge, when None is given. But it might be better to solve this in the Matchms software, since in my (limited) experience, it is very common for spectra to have part of their charges annotated as None. I am not sure what the best solution is, but in the previous comment I give a few possible solutions. Maybe we can discuss this on Tuesday.

niekdejonge:
Decided to go for a solution were the ionmode is used, to determine the parent mass if the charge is not given. The charge is than assumed to be -1 or +1. Also a note is added that the filter functions correct_charge and derive_ionmode, might be usefull to run beforehand. 

@florian-huber Some errors are returned, that seem to be unrelated to this pull request, since I also saw them for other pull request, are you already looking at these, or shall I make an issue about this? (the errors are related to test_spectrum and creating plots)

florian-huber:
@niekdejonge I saw that test were failing because dependencies (matplotlib) had changed in the meantime, and the master branch underwent a few updates. So I merged it into this branch.

Then I tried to stick to your idea and refactored the filter function. Let me know if that looks OK to you.
(I realized that the original version had a few too many if's)

matchms/filtering/add_parent_mass.py
```diff
@@ -32,28 +32,53 @@ def add_parent_mass(spectrum_in: SpectrumType, estimate_from_adduct: bool = True
...
+        return charge
+    if spectrum.get('ionmode') == "positive":
+        print("Missing charge entry, but positive ionmode detected. "
+              "Consider prior run of `correct_charge()` filter.")
+        return 1
+    if spectrum.get('ionmode') == "negative":
+        print("Missing charge entry, but negative ionmode detected. "
+              "Consider prior run of `correct_charge()` filter.")
+        return -1
+    return 0
```
niekdejonge, line 87: If no ionmode and no charge are given, maybe also print: "No charge or Ionmode found consider prior run of "derive_ionmode()" and "correct_charge()""

niekdejonge:
@florian-huber Looks good and is cleaner now. 
Even though the use of get_charge() has similarities to correct_charge, I think this functionality is a nice addition, since this function does not alter the saved charge in the metadata, which might make this solution a more suitable option than using correct_charge(). 
Only one small comment above, for the rest I think is ready to be merged.

matchms/filtering/add_parent_mass.py
florian-huber: Good point. I will add that before merging. Thanks!

PR #255, opened 2021-11-01
Create metadata_export.py

### LINKED ISSUE #372: [feature request] add functionality to export metadata in tabular format
To my knowledge, there is no functionality to export metadata of one or multiple spectra to `csv`, `tsv`, or other tabular formats. Such a feature could be useful in molecular networking workflows. For example, [MetGem](https://metgem.github.io/) does not read metadata from `.msp` files but expects users to supply a [metadata table](https://metgem.readthedocs.io/en/latest/user_manual/import.html#metadata).

Expand `exporting` module with a function to export metadata. The function would take a list of spectra, a path, and an output format as inputs.
Optionally, the user could pass the keys to be exported. Not sure if that's a good idea or not.

**Any good starting points?**
An example of writing metadata to csv:
```python
import csv

spectra: List[matchms.Spectrum]
keys: Set[str]  # union of all metadata keys to be exported

with open("metadata.csv", "w", newline="") as f:
    writer = csv.DictWriter(f, keys)
    writer.writeheader()
    writer.writerows([spectrum.metadata for spectrum in spectra])
```

Alternatively, this can be done by creating a `pandas.DataFrame` and exporting with `pandas.DataFrame.to_csv` but this will introduce a new dependency to `matchms`.

**Scientific reference**
Possible use case: molecular networking with MetGem https://doi.org/10.1021/acs.analchem.8b03099

florian-huber:
Closes #372


hechth:
@florian-huber do you mind if I'd finalize the implementation and update it for the reworked metadata format? Or what is the current status of this - seems quite complete to me, only needs some tests?

PR #257, opened 2021-11-02
Linting updates

### LINKED ISSUE #256: complaints from pylint updates
pylint updates have become a bit stricter it seems...
That leads to failing tests in #253 and #255 and should be handled in a separate PR for linting (either linting or make pylint less strict...)

florian-huber:
Code linting triggered by #256.


florian-huber (commented):
Most things are entirely un-critical linting changes.
Only thing worth mentioning is the addition of `encoding="utf-8"` as default for a number of `open()` occurences.
Still, I will merge this after self-review.

PR #258, opened 2021-11-03
Support importing MGF from file-like object

bernardolk:
* Added tests for both import by file path and file-like object
* Changed type annotations and docstring to indicate support

Turns out pyteomics already supported receiving a file-like (TextIO for instance) object as input for creating the MGF object. So I just changed the docs and added tests to make it clear that we support it.


florian-huber:
Nice, thanks @bernardolk. I will have a look at it today!

florian-huber:
GitHub workflows were not called (we need to change the settings I guess). So I will first merge it into a local branch to run the tests.

Issue #106, opened 2020-07-29
Spectrum hashing


kiahmcintosh:
Currently `hash(spectrum)` for a Spectrum object gives an unhashable type error. Could a hash function be added?

sverhoeven:
Just out of curiosity, for what are you using the hash function?

kiahmcintosh:
I am doing molecular networking using the NetworkX package. NetworkX uses dictionaries to store the data so nodes in the network are required to be hashable.

florian-huber:
@sverhoeven what would you use to hash?
Only issue I see is that we have no default ID for the Spectrum object and it will rely on what metadata is parsed.
Could it be something like:
```python
class Spectrum: 

...

    def __hash__(self):
        return hash(self._metadata.get("spectrumid", "no spectrumid"))
```

sverhoeven:
By only using id, a filtered Spectrum would have the same hash as unfiltered, while their content is different. Using a hash based on id might be OK if you don't have same identifier in a set or as dict keys.

When you don't have an id fallback to use all data of Spectrum. This could be done by giving JSON string to hash(). It will be very slow to generate so cache the hash so it does not have to be recalculated each time and unset it during clone().

florian-huber:
For the GNPS data we worked with, the ``spectrumid`` field was unique. But it is not standard for all the other import formats.
And all other field can also undergo changes during filtering/processing (e.g. title/name changes etc.).

We hence might have to add a better, more standardized ID field to the spectrum objects. 
Maybe this might even need a method that sees all spectrums?
For instance along those lines:
```python
add_id_to_spectrums(spectrums, default_id="spectrumid"):
    for i, spectrum in enumerate(spectrums):
        if spectrum.get(default_id):
            assert spectrum.get(default_id) not in ids_dict.keys(), "Non-unique identifier found."
            ids_dict[i] = spectrum.get(default_id)
        else:
            spectrum.set(default_id, str(i))
```
Just using the counter as id seems a bit poor, but I am no so sure we can find one standard field that does it well enough...

florian-huber:
In #10 @sdrogers pointed to SPLASH which sounds like a good solution for the spectrum hashing!
--> https://github.com/berlinguyinca/spectra-hash/blob/master/python/splash/splash.py

Not a pypi or conda package, but the actual needed code is just a few lines which could be added to matchms.

florian-huber:
@sdrogers how commonly used is SLASH as a unique spectrum identifier?
[...]

PR #259, opened 2021-11-03
add __hash__ method to Spectrum

florian-huber:
Adressing #106 

Added here:
`Spectrum()` class now contains 3 new methods:
- `.metadata_hash()`- computes sha256-based hash from metadata (json)
- `.spectrum_hash()`- computes sha256-based hash from peak mz & intensities (insprired by SPLASH)
- `.__hash__`- combines both other hashes and converts it to integer so that `hash(spectrum)` works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `Spectrum()` objects now also allows generating hashes, e.g. `hash(spectrum)` [#259](https://github.com/matchms/matchms/pull/259)
+- `Spectrum()` objects can generate `.spectrum_hash()` and `.metadata_hash()` to track changes to peaks or metadata [#259](https://github.com/matchms/matchms/pull/259)
+
 ### Changed
 
 - Code linting triggered by pylint update [#257](https://github.com/matchms/matchms/pull/257)
--- a/matchms/Spectrum.py
+++ b/matchms/Spectrum.py
@@ -1,4 +1,5 @@
 from typing import Optional
+import hashlib
 import numpy
 from matplotlib import pyplot
 from .Spikes import Spikes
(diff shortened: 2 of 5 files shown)


sdrogers:
would it be better to just import that git repo as a dependency? E.g. in the Metabolomics USI req file (last line):
https://github.com/mwang87/MetabolomicsSpectrumResolver/blob/master/requirements.txt
??

florian-huber:
We could add the github repo as dependency although I am not a huge fan of this. Here I was only trying to see how much work it would be to add the necessary code to matchms itself.
Doing that I also found some other issues:
The main thing is that the SPLASH method has many (hidden) parameters, and some of them are chosen in a way that (to me) doesn't make sense for matchms. They use a parameter `INTENSITY_PRECISION = 0` which means that spectrum intensities are binned to integers. In our case we often operate with intensities scaled to [0, 1] so that intensity difference would not make it into the generated hash. We can of course simply change the parameter, but that would already mean that we will get different hashes anyway.
I would thus argue that we can also switch to a slightly different approach --> I'll suggest some things in #106

florian-huber:
@hechth, @sdrogers, @sverhoeven has any of you time (and interest) to review this PR?

hechth:
I can have a look in the afternoon eventually 😊

hechth:
I'm currently having a look at the tests and started some refactoring to reduce the usage of constructors and so on - I think this PR is fine, I'd just put the refactorings from my side up in a separate PR since they will also affect different functionality. What do you think?

The main reason is that if there will ever be the decision to change the Spectrum constructor, we will have to change all the tests etc., which creates a big barrier for contributions.

@florian-huber I could also already put up some of the planned changes for the files in this PR so that you get an idea.

florian-huber:
Thanks @hechth ! That sounds like good refactoring steps.
If I understand you well, those refactorings would not change the API for the user, right? 
As long as the changes of this PR won't be affected (from the user perspective) I would say we first try merging this PR and then start a new one for the refactoring. What do you think?

(mostly because what you describe sounds substantial, so nothing we could add with a few tiny changes)

hechth:
@florian-huber No, no API changes, just internal (not even the internals of the package, just the internals of how the tests work - making them parameterized, using builders etc.). So I think merging this PR first would be good, then I can put up a draft with my suggestions :).

PR #261, opened 2021-11-11
Refactoring of tests with builder & parameters

hechth:
The tests are all relying on the constructor of the `Spectrum` class - in case this changes, all tests have to be rewritten/adapted.

The [_builder pattern_](https://refactoring.guru/design-patterns/builder) can be used to avoid this, by introducing an abstraction in form of a builder, which is the only place that calls the constructor of a class. The class can be modified with the builder. This allows adding parameters to the constructor without actually breaking the tests, as only the builder will have to be changed and can be used with a default value for the new parameter, which doesn't have to be optional itself in turn.


florian-huber:
Just because I saw that you also got into some isort strictness issues... I would be open to make the import linting more flexible. Just let me know of make an isssue (for options which are currently set in `setup.cfg` check https://github.com/PyCQA/isort)

hechth:
> Just because I saw that you also got into some isort strictness issues... I would be open to make the import linting more flexible. Just let me know of make an isssue (for options which are currently set in `setup.cfg` check https://github.com/PyCQA/isort)

I don't mind that ^^ The GitHub action tells you where the problems are so I'm okay with it.

hechth:
The harmonization tests could overall use some improvements as they all test for somewhat similar conditions

florian-huber:
Only had a brief look so far, but that looks very nice!
Just let me know when you are done and I will do a more thorough code review.

florian-huber:
Just a note: I got a failing test in the main branch, apparently due to recent changes in matplotlib. I fixed that in #264 (just in case you run into the same thing here).

hechth:
I don't experience that locally since I still use my old conda environment ^^ Thanks for the heads up, I can maybe update it.

tests/builder_Spectrum.py
```diff
@@ -0,0 +1,36 @@
+import numpy
+from matchms import Spectrum
+
+
+class SpectrumBuilder:
```
florian-huber, line 5: Things from /tests/ won't end up in the actual documentation. But it would still be nice to have a short docstring explaining what this class is meant for.

tests/test_harmonize_undefined_inchi.py
```diff
@@ -1,66 +1,21 @@
...
+
+
+@pytest.mark.parametrize("metadata, aliases, undefined, expected", [
+    [{"inchi": ""}, ["", "N/A", "NA", "n/a"], "", ""],
+    [{"inchi": "n/a"}, ["", "N/A", "NA", "n/a"], "", ""],
+    [{"inchi": "N/A"}, ["", "N/A", "NA", "n/a"], "", ""],
+    [{"inchi": "NA"}, ["", "N/A", "NA", "n/a"], "", ""],
+    [{"inchi": "nan"}, ["nodata", "NaN", "Nan", "nan"], "", ""],
+    [{"inchi": "nan"}, ["nodata", "NaN", "Nan", "nan"], "n/a", "n/a"]
+])
```
florian-huber, line 13: Nice test refactoring. Thanks!

florian-huber:
Super nice test refactoring so far!
I also like the builder concept since it makes tests easier to read (and in the future: easier to write).

One thought on it though: It think we should also add a test for `SpectrumBuilder` itself, where we compare it with spectra generated the "old way".

hechth:
> Super nice test refactoring so far! I also like the builder concept since it makes tests easier to read (and in the future: easier to write).
> 
> One thought on it though: It think we should also add a test for `SpectrumBuilder` itself, where we compare it with spectra generated the "old way".

Yeah, agree with that!

It primarily reduces the number of calls to the constructor which makes adding additional parameters possible.

hechth:
@florian-huber Is there an option to check the log entries out of order or so, since checking them exactly can be quite complex and makes the tests quite intricate. I agree that the logging has to be tested with the actual functions to see if they actually behave properly.

florian-huber:
@hechth I would say we can merge this (once tests pass), what do you think?

I would prefer having this in the master branch soonish so that I can also include it in `development`, which would save double-work. And, it would also be a great feature to have to better test things with the additional Metadata functionalities I am working on.

florian-huber:
I will merge this into master. There might be a few tests that we further want to refactor according to the new pattern, but that can also be done in future PRs. 
Thanks so much for this huge effort @hechth !

hechth:
Hi, thank you very much for taking care of it and merging the PR! I have been on vacation this week so sorry for being absent.

PR #262, opened 2021-11-11
Import file like mgf

### LINKED ISSUE #254: Load spectra from already loaded MGF bytes or file-like object
I would like to use a file selector widget to select and upload an MGF file from a user's machine into the application's memory (which is not currently running at the user's machine) and use it with matchms. Having only path to file option for importing the MGF data makes this use-case impossible (without extra steps).

I would like to use a file selector widget to select and upload a MGF file from a user's machine into memory of the application and then pass that file-like or bytes object to the MGF loader _instead of passing a local path of the MGF file to it_

**Any good starting points?**
I think we would need to modify the MGF object constructor (or parents) to support this

Thanks

florian-huber:
Original PR from @bernadolk (#258):

> * Added tests for both import by file path and file-like object
> * Changed type annotations and docstring to indicate support
>
> Turns out pyteomics already supported receiving a file-like (TextIO for instance) object as input for creating the MGF object. So I just changed the docs and added tests to make it clear that we support it.


florian-huber:
Thanks for the PR @bernardolk !
Feel free to add your name etc. to the contributers list in https://github.com/matchms/matchms/blob/master/.zenodo.json
(or let me know what to add there)

Issue #263, opened 2021-11-15
Make filter "strictness" more consistent (and revisit design)


florian-huber:
Over the time the number of filters in `matchms.filtering` has grown (and continues to grow).
It seems about time to revisit them and check consistency regarding missing or false entries: 

+ How do filters handle 1) missing and 2) incorrect (= not usable for the intendet purpose) entries?
Some will only do stuff when possible and return unchanged spectra otherwise (e.g. `add_fingerprint()`), others will print a warning (e.g. `add_parent_mass()`). 
--> Should be consistens (see also missing logging #71 )

+ Currently we have `add_xyz()` type of filters, or -very similer- `derive_xyz_from_abx()` type of filters to add field entries. Those won't fail if an entry is not found. Stricter processing can be achieved by combining with a second filter of type `require_xyz()`.
I wonder if adding a `strictness` parameter to the filters would make that workflow easier?

hechth:
I agree with the requirement/strictness thing, since we eventually want to ensure everything has something, otherwise, fail, but sometimes we might also just reduce the input data and _filter_ the ones for which deriving/adding metadata can be done successfully.

florian-huber:
While implemting the logging (#271) I will also try to make the filter outputs a bit more consistent.

florian-huber:
Outdated.

PR #265, opened 2021-11-16
Implemented filters to add retentiontion time & index

### LINKED ISSUE #260: Add retention data filters
As discussed in #251 , adding RI and RT filters to clean the metadata in the spectrum object would be good.

One additional idea would be for the retention time filter to allow passing a conversion factor which converts the RT to seconds and use this as a canonical unit for RT?

hechth:
Implemented filters for retention time and retention index to add them to the metadata.

The current versions don't perform any dedicated error handling or strictness as this should be implemented in a harmonized way across all filters, so it can be added on later, see #263 .

Closes #260


matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
+from typing import Any
+from typing import List
+from matchms.utils import get_first_common_element
+from ..typing import SpectrumType
+
+
+_retention_time_keys = ["retention_time", "retentiontime", "rt", "scan_start_time"]
```
niekdejonge: One additional variation I came across was "RT_Query" (output of GNPS in MGF files)

niekdejonge: And "rtinseconds" in a mgf file

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+        spectrum (SpectrumType): Spectrum from which to read the values.
+        target_key (str): Key under which to store the value.
+        accepted_keys (List[str]): List of accepted keys from which a value will be read (in order).
+
+    Returns:
+        SpectrumType: Spectrum with value from first accepted key stored under target_key.
+    """
+    rt_key = get_first_common_element(spectrum.metadata.keys(), accepted_keys)
+    value = spectrum.get(rt_key)
+    spectrum = safe_store_value(spectrum, value, target_key)
```
niekdejonge, line 66: Here the first common element is selected, for which you check if it is a valid entry. 
If the input of metadata would be:
{"rt": "None", "retentiontime": 12.17} 
I believe the "retention_time" will currently be set to None (since "rt" is the first common element)
It would be better if it keeps checking all common elements until one returns something else than None. 

I would also not have expected the example above, but it is actually quite common, the first example I could find had this as input:
{'retention_time': [], 'scan_start_time': [0.627583333333]}

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+        spectrum (SpectrumType): Spectrum to which to add 'value' in 'target_key'.
+        value (Any): Value to parse into 'target_key'.
+        target_key (str): Name of the key in which to store the value.
+
+    Returns:
+        SpectrumType: Spectrum with added key.
+    """
+    if value is not None:   # one of accepted keys is present
+        try:
+            value = float(value)
```
niekdejonge: For some weird reason multiple file types I have seen so far had the retention time stored as list of floats like this:
'scan_start_time': [0.627583333333]
Which would not be loaded in the current setup. 
So as suggested change, something like:
if isinstance(value, list):
    if len(value) == 1:
        value = value[0]

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
+from typing import Any
+from typing import List
+from matchms.utils import get_first_common_element
+from ..typing import SpectrumType
+
+
+_retention_time_keys = ["retention_time", "retentiontime", "rt", "scan_start_time"]
+_retention_index_keys = ["retention_index", "retentionindex", "ri"]
```
niekdejonge, line 10: The retention time is sometimes given in minutes and sometimes in seconds (really annoying again...). It would be best if this would be handled by this function, but I don't have any good suggestions myself. 
So mostly a note that you are aware of this and maybe someone has a good suggestion how to solve this.

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
+from typing import Any
```
niekdejonge, line 1: @hechth Looks good! Clear functions that were easy to follow for me, below some suggestions for solutions for weird exceptions that you can come across in common data formats.

matchms/filtering/add_retention.py
hechth: Thank you very much for those valuable inputs!

matchms/filtering/add_retention.py
hechth: I didn't expect this but it can be changed to repeat the process until it finds something useable.

matchms/filtering/add_retention.py
hechth: I'm actually aware of this and dealt with that some times. I thought about making this eventually available as an optional parameter for the function.

matchms/filtering/add_retention.py
florian-huber: That's indeed good to know, thanks!

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Returns:
+        SpectrumType: Spectrum with added key.
+    """
+    if value is not None:   # one of accepted keys is present
+        try:
+            value = float(value)
+            rt = value if value >= 0 else None  # discard negative RT values
+        except ValueError:
+            print("%s can't be converted to float.", value)
```
florian-huber:
```suggestion
            print(f"{value} can't be converted to float.")

```

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+    spectrum = spectrum_in.clone()
+
+    target_key = "retention_time"
+    spectrum = add_retention(spectrum, target_key, _retention_time_keys)
+
+    return spectrum
+
+
+def add_retention_index(spectrum_in: SpectrumType) -> SpectrumType:
+    """Add retention idnex into 'retention_index' key if present.
```
florian-huber:
```suggestion
    """Add retention index into 'retention_index' key if present.

```

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+            value = float(value)
+            rt = value if value >= 0 else None  # discard negative RT values
+        except ValueError:
+            print("%s can't be converted to float.", value)
+            rt = None
+        spectrum.set(target_key, rt)
+    return spectrum
+
+
+def add_retention(spectrum: SpectrumType, target_key: str, accepted_keys: List[str]) -> SpectrumType:
```
florian-huber: This is a helper function for `add_retention_time()`, right? We could add an underscore to indicate that: `_add_retention()`.

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+    Negative values and those not convertible to a float result in 'retention_time'
+    being 'None'.
+
+    Args:
+        spectrum_in (SpectrumType): Spectrum with retention time information.
+
+    Returns:
+        SpectrumType: Spectrum with harmonized retention time information.
+    """
+    spectrum = spectrum_in.clone()
```
florian-huber, line 84:
```suggestion
    if spectrum_in is None:
        return None

    spectrum = spectrum_in.clone()

```

florian-huber: Not sure this is necessary here. But that's how it is currently done in many filter functions.

matchms/filtering/add_retention.py
```diff
@@ -0,0 +1,83 @@
...
+def add_retention_index(spectrum_in: SpectrumType) -> SpectrumType:
+    """Add retention idnex into 'retention_index' key if present.
+
+    Args:
+        spectrum_in (SpectrumType): Spectrum with RI information.
+
+    Returns:
+        SpectrumType: Spectrum with RI info stored under 'retention_index'.
+    """
+    spectrum = spectrum_in.clone()
```
florian-huber, line 104:
```suggestion
    if spectrum_in is None:
        return None

    spectrum = spectrum_in.clone()

```

tests/test_add_retention.py
```diff
@@ -0,0 +1,54 @@
...
+    spectrum_in = Spectrum(mz=numpy.array(
+        [], "float"), intensities=numpy.array([], "float"), metadata=metadata)
+
+    spectrum = add_retention_index(spectrum_in)
+    actual = spectrum.get("retention_index")
+
+    if expected is None:
+        assert actual is None
+    else:
+        assert actual == expected and isinstance(actual, float)
```
florian-huber, line 57:
```suggestion
        assert actual == expected and isinstance(actual, float)

def test_empty_spectrum():
    spectrum_in = None
    spectrum = add_retention_time(spectrum_in)
    assert spectrum is None, "Expected different handling of None spectrum."

    spectrum = add_retention_index(spectrum_in)
    assert spectrum is None, "Expected different handling of None spectrum."
```

florian-huber:
Great additions @hechth!
I left a few comments (rather minor). Guess the actual cases mentioned by Niek will be more critical (looks like a lot of inconsistencies in the way people store the required metadata).

matchms/filtering/add_retention.py
hechth: Resolved in https://github.com/matchms/matchms/pull/265/commits/42f0a75c71faef8479dbd6fc2e37f2c560f25bef

matchms/filtering/add_retention.py
hechth: Thank you for the thorough review!

matchms/filtering/add_retention.py
hechth: Please see the changes in https://github.com/matchms/matchms/pull/265/commits/e22ed8f5278a6efe95e1146a3d334d1f7d3dc840

matchms/filtering/add_retention.py
hechth: Please see the changes in https://github.com/matchms/matchms/pull/265/commits/e22ed8f5278a6efe95e1146a3d334d1f7d3dc840

matchms/filtering/add_retention.py
hechth: Please see the additions in https://github.com/matchms/matchms/pull/265/commits/e22ed8f5278a6efe95e1146a3d334d1f7d3dc840

florian-huber (approved):
Really nice, thanks @hechth !

florian-huber:
Looks nice and complete to me, so I will merge it.
@niekdejonge @hechth if anything else needs to be added simply re-open the issue and/or start a new PR.

hechth:
I hope/think that this should be it, so no further additions.

Thank you for the great review @niekdejonge and @florian-huber !

PR #271, opened 2021-11-26
Add logging

### LINKED ISSUE #71: Add consistent logging

### LINKED ISSUE #270: Make print statements optional
Request:
Make the print statements in derive_adduct_from_name() optional. 

Current behaviour:
The function derive_adduct_from_name() prints print(f"Removed adduct {adduct_from_name} from compound name.") and         print(f"Added adduct {adduct_cleaned} to metadata.") for each spectrum. When running on a large dataset (in my case all GNPS spectra > 400.000 spectra), this results in a massive number of print statements. When run in jupyter notebook, this makes it very slow to later load in the jupyter notebook again. 

Improved behaviour:
Add a variable to the function to make printing statements optional.

florian-huber:
Adds logging to matchms with the option to set the level of logging, e.g. to "INFO" (the default is "WARNING") or even highter to "ERROR" to silence all warnings. This can be done using:
```python
from matchms.logging_functions import set_matchms_logger_level

set_matchms_logger_level("INFO")
```

An alternative is to silence the ouput stream and only write it to a logging file:
```python
from matchms.logging_functions import add_logging_to_file

add_logging_to_file("sample.log", loglevel="INFO", remove_stream_handlers=True)
```
If people add custom things and later want to remove stuff (e.g. handlers) from the matchmslogger, the matchms logger can be reset to its default by:
```python
from matchms.logging_functions import reset_matchms_logger

reset_matchms_logger()
```
The same logger (under different name) can also be re-used in other related libraries such as Spec2Vec:
```python
from matchms.logging_functions import _init_logger

_init_logger("spec2vec")
```

Things done:
- define matchms logger in `logging_functions.py` with several functions for users to adapt the logger.
- replaced all print statements with logger.info or logger.warning
- added a few additional logger.info or logger.warning statements where useful
- update test and code examples
- a few minor linting things here and there


florian-huber:
This is not fully done yet. In particular for the parameter choices of the `logger` itself I am not sure.
(I never implemented such logging for a full Python library before)
So, please feel free to throw ideas/suggestions/edits at me @hechth .

matchms/filtering/add_losses.py
```diff
@@ -20,7 +24,7 @@ def add_losses(spectrum_in: SpectrumType, loss_mz_from=0.0, loss_mz_to=1000.0) -
 
     spectrum = spectrum_in.clone()
 
-    precursor_mz = spectrum.get("precursor_mz")
+    precursor_mz = spectrum.get("precursor_mz", None)
     if precursor_mz:
         assert isinstance(precursor_mz, (float, int)), ("Expected 'precursor_mz' to be a scalar number.",
                                                         "Consider applying 'add_precursor_mz' filter first.")
```
hechth, line 30: Is it possible to write the messages of asserts also to the logger with a different level?

matchms/filtering/derive_adduct_from_name.py
```diff
@@ -39,12 +43,12 @@ def derive_adduct_from_name(spectrum_in: SpectrumType,
     if adduct_from_name and remove_adduct_from_name:
         name_adduct_removed = " ".join([x for x in name_split if x != adduct_from_name])
         spectrum.set("compound_name", name_adduct_removed)
-        print(f"Removed adduct {adduct_from_name} from compound name.")
+        logger.info("Removed adduct %s from compound name.", adduct_from_name)
```
hechth, line 46: It seems like some `derive_xyz` filters have logging output on success and some on failure - do we want to make this more homogeneous?

matchms/filtering/derive_inchi_from_smiles.py
```diff
@@ -19,6 +23,6 @@ def derive_inchi_from_smiles(spectrum_in: SpectrumType) -> SpectrumType:
             inchi = inchi.rstrip()
             spectrum.set("inchi", inchi)
         else:
-            print("Could not convert smiles", smiles, "to InChI.")
+            logger.warning("Could not convert smiles %s to InChI.", smiles)
```
hechth: Same here - no info on success, but on failure.

hechth (commented):
Great work! I'll check it out in action today. Regarding the tests, I'd suggest waiting with changing all tests to check this as there is also the test refactoring PR and I assume we might get some merge conflicts in there.

matchms/filtering/add_losses.py
florian-huber: Yes, good point.
I can add those as `logger.error()` messages.

matchms/filtering/derive_inchi_from_smiles.py
florian-huber: True, this also connects to #263.
For now, I can do a first scan and add those missing info-statements for successes at least for the code touched in this PR. We probably might do another round in the future to see where else logging statements would be helpful.

matchms/filtering/add_losses.py
florian-huber: Sorry, looked again and am less sure.
The asserts will stop the code execution anyway so usually there will at max be 1 logged message. It can be done, e.g. via 
```python
try:
    assert isinstance(precursor_mz, (float, int)), my_error_message
except AssertionError as err:
    logging.exception(err)
    raise
```
But that means a lot of additional lines of code...

matchms/filtering/derive_inchi_from_smiles.py
florian-huber: Should be mostly done by https://github.com/matchms/matchms/pull/271/commits/cb362654e762ac9c10ff3850057689f74a1d3866 
(time-wise I would suggest to further add logging in another PR)

matchms/filtering/add_losses.py
florian-huber: Maybe we can keep that as a new issue for future PRs.

hechth (approved):

florian-huber:
Thanks for reviewing @hechth . I will merge it now and hope it will not too much complicate #261 (but I guess we will see some merge conflicts for sure...)

hechth:
@florian-huber no worries I'll take care of that.

PR #272, opened 2021-12-07
Error handling for string precursor m/z values.

janusson:
When a string instead of a float or integer is passed, the value is converted to a float object.


florian-huber:
Was solved with #275

Issue #249, opened 2021-09-17
Add handling MSP files with peak comments


hechth:
**Is your feature request related to a problem? Please describe.**
Currently, matchms doesn't handle comments in MSP files ([example file](https://github.com/matchms/matchms/blob/master/tests/rcx_gc-ei_ms_20201028_perylene.msp). This functionality could be added.

**Describe the solution you'd like**
Adding a `comments` attribute to the `Spikes` class which holds an array of all comments. Non-existing comments can be represented as `None`.

This option (unfortunately) would require quite substantial changes. One more option would be to add the optional comments as a `dict` or indexed list (index being the peak number to which the comment belongs) to the `Spectrum` class.

Another option would be to replace the complete `Spikes` classes internal data handling with a powerful framework like `pandas`, but I'm not sure whether this is intended (I myself am not sure if this is really necessary or the best solution, so I'm open to suggestions!).

florian-huber:
Hi @hechth !
So far, we kept the `Spikes` class fairly minimal and only designed it for the peaks itself (m/z and intensities). So, we didn't add things like precursor m/z etc. explicitly either. I wonder why we would need to do this for `comments` (which seems less essential). 

In the current design, comments would be added to the `metadata` of `Spectrum` and we as far as I see we would only have to change the MSP exporting function. The importing function should already add the comments field to the metadata. Wouldn't that do? (maybe I'm overlooking things...)

A side not: `Spectrum` objects now store the metadata as a normal Python dictionary. If there's enough added bonus it might indeed be worth considering using DataFrames instead in the future.

hechth:
@florian-huber thanks for the explanations and the quick response! Yeah, I've seen that the `Spikes` class is very minimal and only for the peaks. The problem I'm talking about are the comments which can be added to a peak as can be seen in the example below:

```
NAME: Perylene
SCANNUMBER: -1
RETENTIONTIME: -1
RETENTIONINDEX: 2886.9
PRECURSORMZ: 252.09323
PRECURSORTYPE: [M]+
IONMODE: Positive
SPECTRUMTYPE: Centroid
FORMULA: C20H12
INCHIKEY: CSHWQDPOILHKBI-UHFFFAOYSA-N
INCHI: 
SMILES: C1=CC2=C3C(=C1)C1=CC=CC4=C1C(=CC=C4)C3=CC=C2
AUTHORS: Price et al., RECETOX, Masaryk University (CZ)
COLLISIONENERGY: 70eV
[...]

PR #277, opened 2021-12-09
Add handling MSP files with peak comments (issue #249)

maximskorik:
* modified `load_from_msp` to read peak comments
* modified `save_as_msp` to write peak comments to a file and increased the tab size to avoid misaligned comments in the output file
* added peak-comments handling to `Spectrum` class: the comments can now be stored as an `mz: comment` dictionary in `peak_comments` property of `Spectrum()` objects

**Notes:**
0d7d513 selects **mz** & **intensity** rather than **rt** as the commit message suggests


matchms/exporting/save_as_msp.py
```diff
@@ -1,84 +1,95 @@
...
+        if not _is_num_peaks(key):
+            outfile.write(f"{key.upper()}: {value}\n")
+
+
+def _format_peak_comment(mz: Union[int, float], peak_comments: Dict):
+    """Format peak comment for given mz to return the quoted comment or empty string if no peak comment is present."""
+    peak_comment = peak_comments.get(mz, None)
+    if peak_comment is None:
+        return ""
+    return f"\"{peak_comment}\""
```
hechth: Maybe check if it can be converted to a string without errors?

matchms/importing/load_from_msp.py
```diff
@@ -31,6 +33,9 @@ def parse_msp_file(filename: str) -> Generator[dict, None, None]:
 
                 for peak in peak_pairs:
                     mz, intensity = get_peak_values(peak)
+                    comment = get_peak_comment(peak)
+                    if comment is not None:
+                        peak_comments.update({mz: comment})
```
hechth, line 38: I think this could be integrated into getting peak values which could then return a triple

hechth (commented):
@florian-huber If you could also have a look at this would be great, also regarding the overall design and strategy.

florian-huber:
> If you could also have a look at this would be great, also regarding the overall design and strategy.

Yes, I can definitely have a look at it, too. It might take a few days though (please also feel free to ping me if I don't get back by mid-next week).

matchms/exporting/save_as_msp.py
```diff
@@ -1,84 +1,95 @@
...
+    _write_metadata(spectrum.metadata, outfile)
+    _write_peaks(spectrum.peaks, spectrum.peak_comments, outfile)
+    outfile.write(os.linesep)
+
+
+def _write_peaks(peaks: Spikes, peak_comments: Spectrum.peak_comments, outfile: IO):
+    outfile.write(f"NUM PEAKS: {len(peaks)}\n")
+    for mz, intensity in zip(peaks.mz, peaks.intensities):
+        peak_comment = _format_peak_comment(mz, peak_comments)
+        outfile.write(f"{mz}\t{intensity}\t{peak_comment}\n".expandtabs(12))
```
florian-huber:
```suggestion
        outfile.write(f"{mz}\t{intensity}{peak_comment}\n".expandtabs(12))
```

florian-huber: I would not add the tab here, because that will be added even if no peak comment is found.
(better add it in `_format_peak_comment()`)

matchms/exporting/save_as_msp.py
florian-huber:
```suggestion
    return f"\"\t{peak_comment}\""
```

matchms/exporting/save_as_msp.py
```diff
@@ -1,84 +1,95 @@
...
+
+def _write_metadata(metadata: dict, outfile: IO):
+    for key, value in metadata.items():
+        if not _is_num_peaks(key):
+            outfile.write(f"{key.upper()}: {value}\n")
+
+
+def _format_peak_comment(mz: Union[int, float], peak_comments: Dict):
+    """Format peak comment for given mz to return the quoted comment or empty string if no peak comment is present."""
+    peak_comment = peak_comments.get(mz, None)
```
florian-huber, line 81: In most cases this will work.
But this will create issues once mz values are transformed, e.g. by some filtering function. Say a function applies binning to the mz values in spectrum.peaks.mz, then the keys won't match the peak_comments dictionary anymore.

Could be this is an exotic use-case. But we could add something to avoid this, say:
```python
# add to default parameters:
peak_tolerance=1e-05

# add to _format_peak_comment:
comment_keys = list(peak_comments.keys())
if mz in comment_keys:
    peak_comment = peak_comments.get(mz, None)
elif np.isclose(mz, comment_keys, rtol=peak_tolerance).any():
    new_key = comment_keys[np.isclose(mz, comment_keys, rtol=1e-05).argmax()]
    peak_comment = peak_comments.get(new_key, None)
else:
    peak_comment = None
```

florian-huber: If we add this, we should also add a test for it.

florian-huber:
I had a look at it and it generally looks good to me.
I think adding `peak_comments` to Spectrum and not to Spikes is a good idea, because it avoids including the comments in all other Spikes methods (sorting etc.), see also #249 .

One concern I have (see my comments) is that there is no functional connection between `spectrum.peak_comments` and `spectrum.peaks`. That means 
- if peaks are removed during filtering, the peak comments will remain - maybe not a big issue?
- if mz values are altered during filtering, say if some binning or other minor correction is done, then the mz values will not match a key of `peak_comments` anymore. I suggested a first workaround for this by addiding a tolerance around the mz<->key matching.

matchms/exporting/save_as_msp.py
maximskorik: It did not come to my mind. Thank you for a good point!

maximskorik:
Hello @florian-huber,
I very much agree with your concern on the lack of connection between `Spectrum.peak_comments` and `Spectrum.peaks`, so I went ahead and added the connection in 4368796. Now, after each reassignment of `peaks`, the comments will be reevaluated. 
I went with your suggestion to use `mz_tolerance` to find matching peaks as this seems to be the best solution since it requires the slightest changes to the existing functionality. I moved this workaround to the `Spectrum` class and added the mz tolerance as a class attribute. This parameter can be changed with `Spectrum.update_peak_comments_mz_tolerance(tolerance)` or by using this method on any instance of the `Spectrum`.
Also, now, if any commented peaks are binned, the comments won't be lost, but will be appended to the new peak (see the test in 4368796).
I understand that these changes further extend the size of `Spectrum`, so if you think that any of these is unnecessary or can be implemented better, I will be grateful for any critique and advice.

florian-huber:
Thanks @maximskorik . I will have a look at it!

florian-huber:
I will merge it to a local branch first to run tests (and continue working on it for a bit).
My general impression is that it looks good. I plan to change the implementation together with some bigger refactoring in the developments branch soon.

PR #284, opened 2021-12-23
Issue 249 comments

### LINKED ISSUE #249: Add handling MSP files with peak comments

florian-huber:
Local version of initial PR (#277) from @maximskorik

> * modified `load_from_msp` to read peak comments
> * modified `save_as_msp` to write peak comments to a file and increased the tab size to avoid misaligned comments in the output file
> * added peak-comments handling to `Spectrum` class: the comments can now be stored as an `mz: comment` dictionary in `peak_comments` property of `Spectrum()` objects

> **Notes:**
> 0d7d513 selects **mz** & **intensity** rather than **rt** as the commit message suggests


florian-huber:
Hi @hechth I just requested a review on this. But simply let me know if you have no time for this.

I think this is a decent solution to issue #249 . It still has the intrinsic problem that peak comments are both metadata and related to the actual peaks. Depending on people use those features we could re-evaluate things at a later point in time. So would be great if you could check if the general API for working with those peak comments looks OK to you.

Since the peak comments are now stored in `metadata` we can easily migrate them to the Metadata class that I am working on in the development branch. In principle, that will also give us more freedom to keep them separate from the "other" metadata if we want to.

hechth:
The PR looks good, only I'd move the tests related to the peak comments to a new file - I might also add some more tests if it is okay for you.

florian-huber:
Perfect, thanks!
Please feel free to add some more tests and move them to a new file.

hechth:
Looks good to me. I removed writing the peak comments as metadata to the MSP header, as they are written with the peaks.

One more thing - do we want to make saving the peak comments to the MSP optional, since some software can't deal with it? Like this, this would actually add the possibility to remove the peak comments if they are present? I mean currently this would have to be done inside matchms, which is also not a big deal.

hechth (approved):

florian-huber:
Thanks a lot for the review and additional changes @hechth !
I just made the writing of peak comments optional (default is set to writing them to file).
Once that passes I will merge this PR.

hechth:
Amazing, thank you very much!

PR #287, opened 2022-01-03
move peak_comments into metadata

florian-huber:


florian-huber:
Hi @maximskorik. I would prefer to move the `peak_comments` dictionary into `metadata` to avoid making it a too prominent part of the `Spectrum` class. Let me know if this would work for you as well.

PR #293, opened 2022-01-06
Development

### LINKED ISSUE #186: check inconsistent entries for spectrumid vs spectrum_id

### LINKED ISSUE #285: Small bug: metadata order matter for equal check
When comparing two spectra with same metadata (but filled in in a different order) those will not be found to be equal.
This is due to the current implementation of `__metadata_eq()` and will be fixed in `developments`.

### LINKED ISSUE #298: Better handling of "PEPMASS" field
When refactoring the code I realized that we do not fully make use of the `PEPMASS` field which sometimes is used to add a tuple of (precursor_mz, precursor_intensity, charge).

(see for instance https://proteomicsresource.washington.edu/mascot/help/data_file_help.html) 

- also read the intensity and charge from that field
- handle those odd charge values of `"1+"` or `"2-"`

### LINKED ISSUE #299: Incorrect Comment field handling for `load_from_msp()`
When importing `msp` files from MassBank, `load_from_msp` does not correctly handle the `Comments:` field.

Reason:
Heuristic  code `for s in splitted_line[1][2:-1].split('" "'):` does not generalize and should be replaced by something better.

### LINKED ISSUE #302: Allow handling charges of style "2+", "1-" ...
Some sources include charges like "1-" or "2+" in their data which cannot simply be turned into integers.

florian-huber:
This PR grew very large because several connected things got refactored.

## Metadata class
The metadata attribute in `Spectrum` is no longer a Python dictionary, but an own [`Metadata` class](https://github.com/matchms/matchms/blob/development/matchms/Metadata.py). This design change allows to add methods that belong to metadata handling to that class, such as an `__eq__` method to check equality (that method now also fixes #285). The main difference, for the user is that `Metadata` now processes/harmonizes the dictionary keys per default such that:
- all keys must be lower-case

If `matchms_key_style=True` (which is the default and not meant to be changed by users) then also:
- spaces become underscores
- keys are converted according to a newly added [matchms key conversion table](https://github.com/matchms/matchms/blob/development/matchms/data/known_key_conversions.csv) 

## Metadata harmonization
From the `Spectrum` class, metadata entries undergo basic harmonization unless specified otherwise.
This is done via `metadata_harmonization` which by default is set to True.
If True it will run the following:
- `pepmass` values are interpreted (because in mgf files they can contain up to three values: mz, intensity, charge)
- `ionmode` entries are harmonized
- `charge` entries are harmonized
- `precursor_mz` entries are harmonized (mostly: converted to float)
- TODO: harmonize `retention_time` and `retention_index` entries (mosty means converting to float I guess?)

## Other changes
- Added a `interpret_pepmass()` filter which is also now part of `default_filters()`
- `Spikes` class was remaned to `Fragments` (felt better suited name, but I don't have any strong opinions here)
- started to replace `numpy.something` by the standard `np.something` because I got annoyed to always change my style when working on matchms (again, no strong opinion). This is not yet done to not further add to the "Files changed", but could be fixed with a simple find/replace
- fixed handling charges like "2-", "1+" (issue #302)
- added test for saving + re-importing mgf files (so far only checking a few basic metadata entries, to be expanded later)


florian-huber:
Hi @hechth sorry for pinging you all the time about such (lengthy) PRs...
Obviously: feel free to ignore!

Also possible would be that we have a chat and I give you a quick tour to show you what changed so that we can discuss it more easily.

hechth:
@florian-huber I'll have a look on Monday - if we could arrange a meeting for the next week would be great!

matchms/Metadata.py
```diff
@@ -0,0 +1,109 @@
...
+        self._data.key_replacements = _key_replacements
+        self._data = _interpret_pepmass_metadata(self._data)
+        if self.get("ionmode") is not None:
+            self._data["ionmode"] = self.get("ionmode").lower()
+        if self.get("ionmode") is None:
+            self._data["ionmode"] = "n/a"
+        self._data = _add_precursor_mz_metadata(self._data)
+        charge = self.get("charge")
+        if not isinstance(charge, int) and not _convert_charge_to_int(charge) is None:
+            self._data["charge"] = _convert_charge_to_int(charge)
```
hechth: This method should already harmonize the metadata, so apply all filters which `harmonize` a field, so make sure that various representations are captured, or how does it behave? If this should be something somewhere in between, I think making this a specific filter, or incorporating it into the default filter would make sense?

matchms/Spectrum.py
```diff
@@ -137,15 +128,15 @@ def plot_histogram():
...
                 middles = (lefts + rights) / 2
                 widths = rights - lefts
                 return edges, middles, widths
 
             bin_edges, bin_middles, bin_widths = calc_bin_edges_intensity()
-            counts, _ = numpy.histogram(self.peaks.intensities, bins=bin_edges)
+            counts, _ = np.histogram(self.peaks.intensities, bins=bin_edges)
             histogram_ax.set_ylim(bottom=intensity_from, top=intensity_to)
             pyplot.barh(bin_middles, counts, height=bin_widths, color="#047495")
             pyplot.title(f"histogram (n_bins={n_bins})")
```
hechth: I think the whole plotting part should be moved to a separate class that takes a spectrum object and then plots it, so a spectrum plotter would actually be a good class to add, in order to decouple the visualization and actual spectrum object.

matchms/filtering/add_retention.py
```diff
@@ -10,7 +10,8 @@
 logger = logging.getLogger("matchms")
 
 
-_retention_time_keys = ["retention_time", "retentiontime", "rt", "scan_start_time", "RT_Query"]
+_retention_time_keys = ["retention_time", "retentiontime", "rt", "scan_start_time",
+                        "RT_Query", "rt_query"]
```
hechth: I think upper and lower case should be covered by default already anyway?

matchms/typing.py
```diff
@@ -1,16 +1,13 @@
...
-from typing import Optional
+from typing import NewType
 from typing import Tuple
 from typing import Union
 import numpy
-from .Spectrum import Spectrum
 
 
-SpectrumType = Optional[Spectrum]
+SpectrumType = NewType("Spectum", object)
```
hechth, line 8:
```suggestion
SpectrumType = NewType("Spectrum", object)
```

hechth: I have never seen this construct before - how is it different?

tests/builder_Spectrum.py
```diff
@@ -24,10 +24,11 @@ def with_metadata(self, metadata: dict):
         self._metadata = metadata.copy()
         return self
 
-    def build(self) -> Spectrum:
+    def build(self, harmonize_defaults: bool = False) -> Spectrum:
         spectrum = Spectrum(mz=self._mz,
                             intensities=self._intensities,
-                            metadata=self._metadata)
+                            metadata=self._metadata,
+                            harmonize_defaults=harmonize_defaults)
```
hechth: The builder's `.build()` method shouldn't have any parameters, as all constructor parameters should be set with individual methods.

tests/test_interpret_pepmass.py
```diff
@@ -0,0 +1,83 @@
...
+    assert "Overwriting existing precursor_mz 202 with new one: 203" in caplog.text, \
+        "Expected different log message"
+
+
+def test_interpret_pepmass_intensity_present(caplog):
+    """Test if example inputs are correctly converted when entries already exist"""
+    spectrum_in = Spectrum(mz=numpy.array([100, 200.]),
+                           intensities=numpy.array([0.7, 0.1]),
+                           metadata={'pepmass': (203, 44, "2-"),
+                                     'precursor_intensity': 100})
```
hechth:
```suggestion
    mz=numpy.array([100, 200.])
    intensities=numpy.array([0.7, 0.1])
    metadata={'pepmass': (203, 44, "2-"), 'precursor_intensity': 100}
    spectrum_in = SpectrumBuilder().with_mz(mz).with_intensities(intensities).with_metadata(metadata).build()
```

tests/test_interpret_pepmass.py
```diff
@@ -0,0 +1,83 @@
+import numpy
+import pytest
+from matchms import Spectrum
```
hechth:
```suggestion
from .builder_spectrum import SpectrumBuilder
```

matchms/importing/load_from_msp.py
```diff
@@ -84,62 +135,18 @@ def get_peak_comment(rline: str) -> str:
 def parse_metadata(rline: str, params: dict):
     """ Reads metadata contained in line into params dict. """
     splitted_line = rline.split(":", 1)
-    if splitted_line[0].lower() == 'comments':
+    if splitted_line[0].lower() == 'comments' and "=" in splitted_line[1]:
```
hechth, line 138: Why does the comment line have to contain an `=` sign?

matchms/importing/load_from_json.py
```diff
@@ -55,11 +59,12 @@ def as_spectrum(dct: dict) -> Union[dict, Spectrum, None]:
...
     # Recognize Spectrum by peaks_json key
     if 'peaks_json' in dct:
-        return dict2spectrum(dct)
+        return dict2spectrum(dct, harmonize_defaults=harmonize_defaults)
     return None
 
 
-def dict2spectrum(spectrum_dict: dict) -> Union[Spectrum, None]:
+def dict2spectrum(spectrum_dict: dict,
+                  harmonize_defaults: bool) -> Union[Spectrum, None]:
```
hechth: What is the difference between `Union[Spectrum, None]` and `Optional[Spectrum]`?

hechth (commented):
There are many changes - overall looks good, I just have a couple of bigger points to make.

As already mentioned, I think the plotting should go entirely into a different module to be separated from the actual spectra and fragments + metadata logic. I think this is very straightforward to implement and achieve, so I don't think this needs a lot of consultation.

Secondly, the builder should not be changed to mimic a constructor with the build function - this is something that I can clean up afterwards - it is also no issue, as it isn't related to actual functionality.

The last thing is the option to `harmonize defaults` which is a little bit unclear to me how this is using the filters or how it is different from the filters. Also how the list of supported conversions and the filter classes interact doesn't seem entirely clean.

tests/builder_Spectrum.py
florian-huber: True, should become part of the metadata related method.

matchms/typing.py
florian-huber: This was to avoid circular imports. I ran into issues because the Spectrum class was imported at places like this one before the actual Spectrum class definition got imported.

hechth: Do you remember in which actual places? Circular imports are something that usually indicate that the interaction entities are somehow tightly couple and it could be beneficial to create another class which actually sits in between and contains the closely coupled code.

matchms/filtering/add_retention.py
florian-huber: That's right. I wanted to first discuss the Metadata harmonization part, but I would indeed argue to keep all keys lower case within matchms so that the "RT_Query" could go.

matchms/typing.py
florian-huber: I guess one of those places is seen here: https://github.com/matchms/matchms/runs/4659699326?check_suite_focus=true

```
_______________ ERROR collecting tests/test_add_compound_name.py _______________
ImportError while importing test module '/home/runner/work/matchms/matchms/tests/test_add_compound_name.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_add_compound_name.py:3: in <module>
    from matchms import Spectrum
matchms/__init__.py:1: in <module>
    from . import exporting
matchms/exporting/__init__.py:8: in <module>
    from .save_as_json import save_as_json
matchms/exporting/save_as_json.py:4: in <module>
    from ..Spectrum import Spectrum
matchms/Spectrum.py:7: in <module>
    from .filtering.add_precursor_mz import _add_precursor_mz_metadata
matchms/filtering/__init__.py:39: in <module>
    from .add_compound_name import add_compound_name
matchms/filtering/add_compound_name.py:2: in <module>
    from ..typing import SpectrumType
matchms/typing.py:6: in <module>
    from .Spectrum import Spectrum
E   ImportError: cannot import name 'Spectrum' from partially initialized module 'matchms.Spectrum' (most likely due to a circular import) (/home/runner/work/matchms/matchms/matchms/Spectrum.py)
```

hechth:
@florian-huber since this has been in the draft stage for some time I could maybe help you with finalizing the PR? According to you, what would still need to be added for this be merged into the main branch? I'm asking since merging this would be great before continuing with the pipelines, masking and harmonization of filters.

florian-huber:
Thanks for asking! I am actually working on it right now. Currently I tried to move the metadata default filters from `Metadata` to the `Spectrum` class.

matchms/Metadata.py
florian-huber: I now changed it. Here in `Metadata` only a basic key style harmonization happens (lower case, underscores, known key word conversions).
The actual harmonization of the values (e.g. changing charge to int) is now called from `Spectrum` via `_apply_metadata_harmonization()`.

matchms/filtering/add_retention.py
florian-huber: I removed the upper case version.

tests/builder_Spectrum.py
florian-huber: I moved it out of the build() method. Please check if it looks better now.

florian-huber:
@hechth I think this is mostly done from my point of view. There are a few comments I still need to answer. Most importantly we should add the retention time/index entry harmonization to the new method in the `Spectrum` class. Let mw know if you want to do this or if I should include it.

matchms/Spectrum.py
```diff
@@ -72,30 +83,31 @@ def __init__(self,
...
         """
-        if metadata is None:
-            self._metadata = {}
-        else:
-            self._metadata = metadata
-        self.peaks = Spikes(mz=mz, intensities=intensities)
+        self._metadata = Metadata(metadata)
+        if metadata_harmonization is True:
+            self._apply_metadata_harmonization()
+        self.peaks = Fragments(mz=mz, intensities=intensities)
```
hechth, line 94: I assume the `retention_time` and `retention_index` harmonization should go in here? Or what exactly did you mean?

matchms/Spectrum.py
```diff
@@ -72,30 +83,31 @@ def __init__(self,
...
+
+    def _apply_metadata_harmonization(self):
+        metadata_filtered = _interpret_pepmass_metadata(self.metadata)
+        if metadata_filtered.get("ionmode") is not None:
+            metadata_filtered["ionmode"] = self.metadata.get("ionmode").lower()
+        metadata_filtered = _add_precursor_mz_metadata(metadata_filtered)
+        charge = metadata_filtered.get("charge")
+        if not isinstance(charge, int) and not _convert_charge_to_int(charge) is None:
+            metadata_filtered["charge"] = _convert_charge_to_int(charge)
+        self._metadata = Metadata(metadata_filtered)
```
hechth: Or should it be used via the `add_retention_time` and `add_retention_index` methods in this part?

matchms/filtering/add_precursor_mz.py
```diff
@@ -1,16 +1,32 @@
...
 from matchms.utils import get_first_common_element
-from ..typing import SpectrumType
 
 
 logger = logging.getLogger("matchms")
 
 
-_accepted_keys = ["precursor_mz", "precursormz", "precursor_mass"]
+_default_key = "precursor_mz"
+_accepted_keys = ["precursormz", "precursor_mass"]
```
hechth, line 9: So this exists in parallel with the metadata harmonization, right?

matchms/Spectrum.py
florian-huber: Yes, I would add it here. I guess we could use the helper functions from add_retention.py and apply it here (or change the structure of add_retention a bit such that there is a helper function which only changes the metadata, that's what I did for `add_precursor_mz`).

matchms/filtering/add_precursor_mz.py
```diff
@@ -1,16 +1,32 @@
 import logging
 from matchms.utils import get_first_common_element
-from ..typing import SpectrumType
```
hechth, line 3: What was wrong with importing the SpectrumType and having the type hints? Is it now due to the circular imports?

matchms/filtering/add_precursor_mz.py
florian-huber: Yes, this is in principle a duplicate now. I would say it could go now, or what do you think?

matchms/filtering/add_precursor_mz.py
florian-huber: Might work now, I'll have to check. Before I got an issue because of a loop: typing (imports Spectrum) --> add_precursor_mz (imports typing) --> Spectrum (imports add_precursor_mz)

matchms/Spectrum.py
hechth: So the harmonization basically only transfers the value if it is present in one of the keys, without changing the actually stored value, am I right?

florian-huber: We have a bit of a duplication here at the moment, because the PickyDict part will already do most of the key harmonization. So if we have data coming in with "retentiontime" or "PrecursorMass" then those keys will automaticall have been replaced by "retention_time" and "precursor_mz" even before this method is called.

What is happening here is the harmonization of the entries itself. So convert strings to numbers (e.g. charge: "2+" --> 2) or make ionmode entries lower-case.

matchms/filtering/add_precursor_mz.py
hechth: since `scan_start_time` is not among the harmonized keywords I'm leaving it here for now

matchms/Spectrum.py
hechth: So converting strings to numbers is also already part of PickyDict?

florian-huber: No, sorry maybe that wasn't very clear. PickDict only handles the dictionary keys. The string-to-number conversions are part of the functions in this part (e.g. `_convert_charge_to_int`).

hechth: ah, so these are the default filters - that is how I implemented it now. I think making this distinction also makes sense: harmonization -> store value in right place, default_filter: make sure value is valid if it is present

The nice part is then that the default filters only have to check the canonical keyword

matchms/Metadata.py
hechth: I've see it, looks good and makes sense :).

hechth:
> I think this is mostly done from my point of view. There are a few comments I still need to answer. Most importantly we should add the retention time/index entry harmonization to the new method in the `Spectrum` class. Let mw know if you want to do this or if I should include it.

I think this is resolved in https://github.com/matchms/matchms/pull/293/commits/7da981e4e9b85b1a643e57cfc69780357636144a

matchms/importing/load_from_msp.py
hechth: @florian-huber if we resolve the remaining points here I think we could merge the development branch.

florian-huber: The `=` was added because I saw two types of comment field uses. In one there is a lot of pairs like `"license=CC BY-NC-SA" "exact mass=247.06333"` which are then addded to the metadata under the respective names. But it is also possible to use the comment field for any other arbitrary message I guess. In that case the former code would simply discard the comment field altogether. Now it will remain in the metadata as `comments`.

hechth: Ah okay, I get it. Thanks for explaining!

matchms/Spectrum.py
florian-huber: Good ideas. But I would rather not tough this in this PR, maybe here --> #311 ?

hechth (approved):

matchms/Spectrum.py
hechth: Also we openend up the wider discussion over at OpenMS regarding visualizations, so we'll see.

florian-huber:
Huge thanks @hechth for helping to implement those changes!
Happy with the result and will merge it into the main branch now.
The `development` branch will stay for the next steps to come on the way towards 1.0.0 :rocket:

PR #303, opened 2022-01-14
update and move plotting functionality

### LINKED ISSUE #282: Revisit spectrum plotting
Currently, `Spectrum` objects have a simple plotting function. That could be improved in a number of ways:

- move the actual function out of the `Spectrum` class so that it can also be used elsewhere
- add option to visually compare spectra
- check for other libraries to visually plot/inspect spectra and see if we could use them
- revisit design/layout, for instance switch from pyplot to OO-style so that it returns an axes object which could be used for further plots elsewhere.

florian-huber:
- plotting functions are now moved to `matchms.plotting` submodule
- plot is now prettier (uses a lot of the code from `spectrum_utils`)
- more types of plots possibe:
`spectrum.plot()` - plot individual spectrum
`spectrum.plot_against(spectrum2)` - mirror plot
`plot_spectra_array(spectrums)` - plots a grid of all spectra in list
- possible to add spectrum.peak_comments to individual peaks by setting `annotate_ions=True`
- functions will return axes object so that it will be easy to also combine multiple plots elsewhere.
- `plot(with_histogram=True)` option was removed. Could be added again, but I haven't seen people using it.

TODO: add example plots to readthedocs


florian-huber:
@bittremieux I here used parts of your `spectrum_utils` code (we defined spectrum classes differently so I couldn't use it straightaway as a package).

Let me know if that is an issue and if you would like to be listed as contributor to matchms.

bittremieux:
Not a problem at all! I'm glad my code was useful. I'm currently also working on some new features/improvements in spectrum_utils, but it is relatively slow going in between some other projects. You know how it is. 😉

spectrum_utils plotting has recently been integrated directly in [Pyteomics](https://pyteomics.readthedocs.io/en/latest/examples/example_annotation.html) and partially duplicated for plotting in [pyOpenMS](https://github.com/OpenMS/OpenMS/pull/5615). I think it could be useful to discuss with these different developers whether there are synergies and how we can have the best interoperability and compatibility between our various packages. Would be cool to include matchms into that discussion as well.

florian-huber:
Thanks for the quick answer! 
Joining forces would hence realy makes sense. I guess the might be ways to make some plotting aspects generic enough so that they could work for several packages at the same time. Let me know if I can help with anything or check where and how to bring matchms into this dicussion.

bittremieux:
Sounds good. Yes, we should have a discussion at some point how we can generalize plotting across packages and which features we want to support. I'll follow up on that with you in the future to start this discussion with other developers as well.

Also let me know if you have found ways to improve the plotting, I'll port that to spectrum_utils as well then.

hechth:
It would be great to get an overview over the different plotting packages for mass spectra in python and to maybe unify/harmonize those a bit :).

florian-huber:
Here, only a self-review was done. I agree that it makes sense to come back to this in the wider context of having a more general python-based plotting library for mass spectra!

PR #306, opened 2022-01-19
changes peaks name for student class

florian-huber:


florian-huber:
This was just to demonstrate to students how continuous integration works...

Issue #313, opened 2022-02-07
safe_convert_to_float in add_retention raises error


niekdejonge:
When safe_convert_to_float is run on an empty list it results in the error: 

TypeError: float() argument must be a string or a number, not 'list'

Value is within a try block. However the exception only accepts a value error, while this is a TypeError. 

Solution:
Also add TypeError to exception block.

florian-huber:
Fixed with #314 . Will be part of matchms 0.13.0

PR #314, opened 2022-02-07
solve bug in safe_convert_to_float

niekdejonge:
Solves the issue in #313.

The error occurs when an empty list is passed to safe_convert_to_float


hechth (approved):
Hi, thanks for catching this one! I didn't encounter this bug so didn't check for it.

niekdejonge:
I can imagine. So many different exceptions in the formatting...

florian-huber (approved):
Looks good. Thanks for catching (and fixing) this one.

Issue #251, opened 2021-10-08
Add RI filtering to matching process


hechth:
**Is your feature request related to a problem? Please describe.**
The retention time/index of a mass spectrum is an additional source of information that can be used to find the best matches, as it is an orthogonal piece of information.

**Describe the solution you'd like**
To find the spectra which are to be matched to a compound, a simple RI filter based on a set threshold can be used to remove the compounds which don't fall into this range.

florian-huber:
I agree, including the retention time would be nice!
Were you thinking of an additional filter function (under `matchms.filtering`) together with the retention time being added metadata of the spectra?

hechth:
This is a tricky one: for the RI, we'd actually like to apply the filter for each spectrum while matching.

Example:
I want to match spectra A, B and C with RIs 1300, 1400 and 1450 to the reference library with spectra D, E and F with RI 1350, 1400 and 1500. I specify an RI tolerance of 60.

The filtering should therefore then only result in matches of A to D, B to D and E and C to E and F.
This means that the filter should act per spectrum, not per actual score matrix computation.

I imagine multiple options to implement this:
1. As a post-processing step -> but this takes away the performance benefit of having to compare less spectra
2. As a pre-processing step -> this removes the matrix-style operation, but might not be that bad performance-wise.

This actually means that it would be more a part of the matching than the filtering.

If it should only be inside the filtering, then one way would be to just match always a single spectrum to the reference library and iterate the filtering and matching step.

@florian-huber Any ideas or hints on that?

sdrogers:
Hola
I wonder if it could still be done within the matrix world - it would be easy enough to make a spec by spec masking matrix (with 1 where the difference is within a limit) and then skip these in the loop to compute scores (or just multiply the score matrix in the end element wise with the mask, if vectorising still proves quicker?)
Is the precursor mass match filter done as a post-process? Because it’s basically the same operation…

florian-huber:
OK, thanks @hechth! I think I now understand better what you meant.
And I basically would agree with @sdrogers . The RI matching is very similar to the `PrecursorMzMatch` which will simply output a matrix of 1 and 0 (True/False) for matches within tolerance or not. That's a very fast operation.

The way I see it we should indeed add some masking functionality to the score calculation. That would be nice for your RI matching, but it would also make sense for other piplines, say computing Cosine score, but only if the precursor-m/z are within a decent tolerance etc.

```python
from matchms import calculate_scores_pipeline, Spectrum
[...]

Issue #165, opened 2020-12-09
Parallelize score calculations


florian-huber:
While we optimized all key scores quiet a bit, many scores currently just do Python loops when comparing large arrays of spectra (most importantly: ``CosineGreedy`` and ``ModifiedCosine``).

Seems only natural to add a parallel implementation of the basic double-loop --> Check with Johan?

Adafede:
Hi there! Was about to post this...new year, new resolutions? Clearly upvoted! 👍🏼

florian-huber:
I spend a bit of time to check if I could implement this using `numba` but that didn't work out because of too many non-numba style functions being invovled. Could maybe be fixed, but I won't have the time for doing this.

Other options might be to use `joblib` or `asyncio`, but I probably won't be able to work on that anytime soon.

hechth:
This is something that might become of interest also for our work, so it might get addressed in the next couple of months.

florian-huber:
Sound great! Let me know if I can help with anything.

Just a minor comment regarding expectations: 
we optimized the indivual pair-wise score computation quite a bit. But the double-for-loop remains and can certainly be parallelized. However, unless run on a bigger infrastructure my expectations on speed gain are modest, maybe about 2-fold on a normal computer with 4-8 cores.
(due to the extra overhead you create by parallelizing)

In practice I expect a much bigger impact will come from smarter pipelines: that means only computing those scores that seem interesting enoug (e.g. within a certain precusor m/z). That would be #268

florian-huber:
More recent issued on that is #325, so this will be closed.

PR #315, opened 2022-02-09
Metadata match

florian-huber:
Starting to sketch a `MetadataMatch` similarity class.


matchms/similarity/MetadataMatch.py
```diff
@@ -0,0 +1,107 @@
...
+        Metadata match between 2 and 3 is 1.00
+        Metadata match between 2 and 4 is 0.00
+
+    """
+    # Set key characteristics as class attributes
+    is_commutative = True
+    score_datatype = bool
+
+    def __init__(self, field: str,
+                 matching_type: str = "equal_match",
```
hechth, line 68: Wouldn't the easiest thing here be to actually pass a comparison function which has to evaluate to true and takes two values which are compared?

florian-huber: That would indeed work as well.
Only issue I see is that this cannot generally be optimized and will only work for pair-wise comparisons. For the two method given here I could at least do some numby optimization which I just pushed.

We could also do a combination here. So, offer the two default types I sketched and also allow passing a custom function. Would that work?

hechth: Could we restrict the functions to things which can be parallelized or so? I thought about passing like `==` operator, a `<=` etc. ... but maybe having both options of taking a some keywords which have numba implementations and allowing to pass a lambda with simple parallel for would be good. Btw., has matchms any parallelism, like parallel for loops at the moment?

florian-huber: So far parallel computation is depending on the respective scores. By default, no parallel execution will happen. I tried several things, but so far I haven't found a good way to implement parallel execution in Python that will work for all scores and/or all systems. The options which tend to work on all scores/systems can create huge bottlenecks and sometimes make things much slower than without parallel execution (e.g. using Python multiprocessing, or even dask).

hechth: Okay - I have little experience with parallelism in python, so likely it is better to keep that discussion somewhere else.

florian-huber: Yes, we have still the issue #165, but could also update that at some point.

hechth:
@florian-huber This looks pretty good so far and can also be used somewhat independently from the other things. What do you think would still be required to make this a full-fledged PR? Because this would likely already implement #251 .

Let me know if I can contribute to this in any way 👍

florian-huber:
@hechth, good  you mentioned #251. I agree that this PR should fix this issues. You also mentione a more flexible comparison function that will take custom functions as input. I would leave this out of this PR and either add it later, or (preferentially) create an entirely new score for this `CustomSimilarity` or similar which will take any user function of the form `custom_function(spectrum1, spectrum2) --> score`.

Let me know if this PR can be merged or if you want any further changes for now.

hechth (approved):
I think this is good for now and can be merged - we can always build upon it later. I think it would be best to do things step by step 👍

florian-huber:
Thanks a lot for the review and suggestions Helge!

Issue #268, opened 2021-11-22
Design calculate_scores_pipeline


florian-huber:
In #251 we discussed how to add different similarity scores, or how to combine masking steps with similarity score computations.

One such option would be to introduce a function/class that allows users to easily stack multiple operations such that the computation time can be reduced significantly (when compared to a naive sequence of simliarity computations).
One example: Two large sets of spectra should be compared using a computationally expensive score (e.g. Cosine). But many pairs can easily be sorted our based on other criteria, e.g. their metadata. For instance: two spectra with very different precusors can be skipped etc.

For this we could implement something like this:
```python
from matchms import calculate_scores_pipeline, Spectrum
from matchms.similarity import CosineGreedy, MetadataMatch, PrecursorMzMatch

spectrums_1 = [...]
spectrums_2 = [...]

scores = calculate_scores_pipeline(spectrums_1, spectrums_2,
                                   masks=[MetadataMatch(field="retention_time", tolerance=60.0),
                                          PrecursorMzMatch(tolerance=50.0, tolerance_type="Dalton")],
                                   score=CosineGreedy())
```
This would first run `MetadataMatch()` on all-vs-all spectrums_1 vs spectrums_2 and fill a matrix with 1 (True) or 0 (False).  
For all True values of matrix: run `PrecursorMzMatch()` and update matrix.  
For all True values of matrix: run `CosineGreedy()`

hechth:
I think it might also make sense to come back to #210 and separate the pair finding function (if we already split it into a pipeline). This way we could also easily implement a forward and reverse score.

florian-huber:
Sorry, I am not sure I fully get the point.
The main issue I see for the pipeline described above is that we no longer need to compute the score for each possible pair. Only for those which pass the masking steps (say, all which have a retention time difference within a set tolerance).
For this I only need a function (or method) that can compute pair-wise scores. Those should already work for all `matchms.similariy` --> `similarity_score.pair(spectrum1, spectrum2)`.

Why would be need to go into the internals of the pair-wise computation?

florian-huber:
Here a few more thoughts on possible designs.

**Option 1** stacking method for `Scores`
This is different from what was sketched above, but could work out as well.
```python
scores = calculate_scores(references=spectrums_1,
                          queries=spectrums_2,
                          similarity_functions=PrecursorMzMatch(tolerance=50.0, tolerance_type="Dalton"),
                          sparse=True)

[...]

PR #316, opened 2022-02-10
Implement sparse scores matrix

florian-huber:
Moving to sparse arrays has some consequences for users. In particular the `Scores` objects will change quite a bit. From a user perspective most code should work as before, but some things will break with the current API
- scores == 0 will not be stored (we could add an option to store them, too, but that might not be necessary). This means that some of the tests had to be changed.
- **Slicing works differently now!** Addressing individual entries remains the same `scores.scores[999, 101]` will still return a particular entry. But only if no more than one score was computed. More generally, the user now has to run `scores.scores[999, 101, "my_score"]` or `scores.scores[999, 101, 2]` to refer to one particular score stored in `Scores`.
- Slicing rows/columns or entire arrays will not return a full 1D or 2D array, but (row_index, col_index, values) tuples only containing the actually stored entries (hence ignoring zeros/empty). This is faster and more efficient for large-scale arrays, but will require adjustment of code using this functionalities. 
- As an alternative users can now get an array of scores by running `scores.get_scores_array("my_score")` or `scores.get_scores_array("my_score", array_type="coo")` which will give an dense numpy array or a sparse scipy-COO array (we can of course also add csr, csc etc.).


hechth:
I think we have to distinguish between two cases which got somewhat mixed up now.

1. We compute all combinations but only store the non-zero ones in a sparse matrix, which is returned from the score computation.
2. We have an input telling us which combinations to compute and the results are returned in a sparse matrix.

florian-huber:
> I think we have to distinguish between two cases which got somewhat mixed up now.
> 
>     1. We compute all combinations but only store the non-zero ones in a sparse matrix, which is returned from the score computation.
> 
>     2. We have an input telling us which combinations to compute and the results are returned in a sparse matrix.

Yes, that's true. My idea would be that the sparse array class should be able to handle both, and the pipeline class will decide which way to go. But there is indeed some design and functionality decision we need to make (maybe better in #268 though).

florian-huber:
Outdated!
Was continued in #368 (and led to https://github.com/florian-huber/sparsestack)

PR #318, opened 2022-02-15
Switch to sparse stacked scores (using structured numpy array)

florian-huber:
- using a new `StackedSparseScores` class which is made in 2D COO-style but can stack multiple scores for every `row, col` pair. In addition, `StackedSparseScores` has more slicing capabilities when compared to usual COO implementations (e.g. scipy) and bring methods that `Scores` can use to add and filter scores. 
- Actual data in `StackedSparseScores` is stored in `_data` as a numpy structured array (my first tries used a dict of numpy arrays, see #316 ).

Example code:
```python
import numpy as np
from matchms import StackedSparseScores

scores1 = np.random.random((12, 10))
scores2 = np.random.random((12, 10))

scores_array = StackedSparseScores(12, 10)
scores_array.add_dense_matrix(scores1, "scores_1")
scores_array.filter_by_range("scores_1", low=0.5)

# Add second scores and filter
scores_array.add_dense_matrix(scores2, "scores_2")
scores_array.filter_by_range("scores_2", low=0.1, high=0.4)

# Scores can be accessed using (limited) slicing capabilities
scores_array[3, 4]  # => scores_1 and scores_2 at position row=3, col=4
scores_array[3, :]  # => tuple with row, col, scores for all entries in row=3
scores_array[:, 2]  # => tuple with row, col, scores for all entries in col=2
scores_array[3, :, 0]  # => tuple with row, col, scores_1 for all entries in row=3
scores_array[3, :, "scores_1"]  # => same as the one before

# Scores can also be converted to a dense numpy array:
scores2_after_filtering = scores_array.to_array("scores_2")
```

Todos:
[ ] add sparse option to `.matrix()` method to allow storing values during calculation in a sparse array. This will significantly reduce memory costs for scores such as "PrecursorMzMatch" or "MetadataMatch".


hechth:
@florian-huber I didn't manage to have a look at the `Scores` rework for some time but now we might pick up here again. I just thought whether it wouldn't make sense to actually have multiple `Scores` implementations (`dense` and `sparse`) with a common API in form of the current `scores` interface? The current `calculate_scores` method could then (depending on the metric or a user parameter) return the respective implementation which fits for this usecase.

What do you think about that? This way, we could keep both score implementations in place and use the one which is better for the usecase.

florian-huber:
In principle, having two Scores implementations is possible. But the most recent development version of `Scores` in [pipeline](https://github.com/matchms/matchms/tree/pipeline) has a different API compared to the current implementation. I fear that it would be a lot of work to develop and maintain two versions of `Scores`.

The assumption behind it is: Most of the time, sparse arrays will be equally good or better to use. If people really need dense arrays then they can either directly skip the scores object and use the `.matrix()` method of the similarity scores, or use the `Scores` `.to_array()` method.

florian-huber:
Outdated --> is now #370

PR #321, opened 2022-02-16
Remove losses

### LINKED ISSUE #319: No longer store losses but compute on the fly
Most parts of #283 were addressed in #293. 
What remains is:

- remove the add_losses() filter and make losses a method rather than an attribute in the Spectrum class (losses don't have to be stored!).
- maybe: add properties for most frequently used metadata entries (e.g. spectrum.precursor_mz)

florian-huber:
- Storing losses as part of spectrum seems a waste of resources (given how quick it is to compute).
- But: This will break pipelines that use `spectrum.losses`


matchms/Spectrum.py
```diff
@@ -203,13 +204,22 @@ def metadata(self):
...
+            losses_mz = (precursor_mz - peaks_mz)[::-1]
+            losses_intensities = peaks_intensities[::-1]
+            # Add losses which are within given boundaries
+            mask = np.where((losses_mz >= loss_mz_from)
+                            & (losses_mz <= loss_mz_to))
+            losses = Fragments(mz=losses_mz[mask],
+                               intensities=losses_intensities[mask])
+            return losses
+        logger.warning("No precursor_mz found. Consider applying 'add_precursor_mz' filter first.")
+        return None
```
hechth, line 222: Is it possible in python to have a property that is in fact a proper function? This way, the compatibility with the `spectrum.losses` pipelines would be retained.

I'm not sure on how to incorporate the filtering here though.

hechth:
@niekdejonge maybe you could give some input on this? This has been open for a while but I think it is still relevant.

matchms/Spectrum.py
niekdejonge: That seems like a nice solution

niekdejonge:
@hechth @florian-huber I agree with both points, it makes sense to remove this and make this a function. But it will indeed also break functionality of Spec2Vec and MS2Query. For MS2Query this is not directly a problem, since it is set to Matchms 0.13.0 anyway, so we can make this switch when upgrading to matchms >0.18.0.

florian-huber:
Hi @niekdejonge and @hechth , I picked this up again in #681 so I will close this.

Following your suggestion, Helge, I now added `losses` as a `@property` so that it can still simply be called by `my_spectrum.losses`. Only if we want to adapt the min_mz and/or max-mz values we need to call `my_spectrum.compute_losses(...)`.

PR #322, opened 2022-02-16
Refactoring Similarity metrics, peak matching & tests

hechth:
While sketching possible layouts and implementations for the pipeline class and general updates, I wen't back to the issue of peak matching (finding which peak pairs to match) and the implementation of the cosine score in matchms is still very confusing to me. So I went and refactored the tests in the `test_spectrum_similarity_functions.py` file and how the peak pairs are collected and the matching pairs computed seems very odd to me - maybe this could be a bit simplified or clarified.

To me it also seems that maybe the mz and intensity values have been interchanged by time and tests written in such a way that they pass - not by actually checking valid results. Also, the shape of the matching_pairs return value seems odd - how can this contain floating point numbers? I assume that these values should be indices.

```python
@pytest.mark.parametrize("numba_compiled", [True, False])
@pytest.mark.parametrize("shift, expected_pairs, expected_matches", [
    (0.0, [[2., 2., 1.], [3., 3., 1.]], (2, 3)),
    (-5.0, [[0., 0., 0.01], [1., 1., 0.01]], (2, 3)),
    (-20.0, None, None)
])
def test_collect_peak_pairs(numba_compiled, shift, expected_pairs, expected_matches, spectra):
    """Test finding expected peak matches for given tolerance."""
    spec1, spec2 = spectra

    func = get_function(numba_compiled, collect_peak_pairs)
    matching_pairs = func(spec1, spec2, tolerance=0.2, shift=shift)

    if expected_matches is not None:
        matching_pairs = numpy.array(matching_pairs)
        assert matching_pairs.shape == expected_matches, "Expected different number of matching peaks"
        assert numpy.allclose(matching_pairs, numpy.array(expected_pairs), atol=1e-8), "Expected different values."
    else:
        assert matching_pairs is None, "Expected pairs to be None."
```

The score computation overall seems to be valid - but I don't feel confident enough to be able to tell that from the test cases.


hechth:
@florian-huber nevermind, I figured how the layout works - it is just very confusing.

florian-huber:
@hechth you made it a draft again. Let me know when this is ready to be reviewed.
In general, I think it is indeed very nice to refactor those tests and I can double-check if they do test the proper things. Some of them might be a bit outdated or (as you mentioned) miss a clear route to the comparison values in the tests.

florian-huber:
Another note. The code implementation was in the past heavily optimized for speed. Could very well be that we should also not optimize it a bit more for readibility.

hechth:
@florian-huber I figured that, yeah, it seems like they are heavily optimized for performance. I might add that to the tests and make performance comparisons of refactored versions of the functionality and the original - maybe we can find a way how to keep the performance and actually have better code readability 👍 .

hechth:
@florian-huber what I had in mind as a layout is something like this:

matchms-Version 2 drawio

The current Greedy and Hungarian versions are actually heuristics on how to compute the actual score from the given matches with the scores from the metric - so basically we have 3 steps
* How to determine the matching peaks / what to include in the scoring (PeakPairer)
* How to compute the similarity of two peaks (SimilarityMetric)
* Which peak similarities to use to use each peak once and get a score. (Heuristic)

So the SpectralSimilarity uses these 3 independent classes to actually compute a score - this way we can combine all sorts of methods and implementing new ones becomes actually quite easy. But that would actually mean also taking out the SimilarityMetric into an own class.

The existing metrics would then simply be a child class of SpectralSimilarity where the arguments for the superclass constructor are already filled in, so to say as a shortcut to the generic method.


This would be a way of refactoring the Similarity metrics to be more flexible and easier to extend. I had the same thought already in the beginning but didn't really know how to do it or how things are connected. I can do a refactoring like this fairly quick, but we will lose the performance I assume. Keeping the performance might be a bit tricky but I think worth trying and maybe achievable!

florian-huber:
Thanks for sketching a refactoring of those scores @hechth!

If you have time to work on it please feel free to give it a try and we can see if we can achive the same performance with a cleaner code layout!

Few comments:
- I wouldn' t name it `SpectralSimilarity` class, because many other spectral similarities will not fit into this scheme (such as Spec2Vec and MS2DeepScore that we worked on). So, a more specific name for such types of scores would be good (e.g. something in the direction of Cosine or peak-matching?)
- the peak pairing is currently already a seperate function in CosineGreedy and Hungarian, so moving this to an own function (or class) should be OK. But it is very cheap to compute so it might not need an own class with API?
- The peak pairing currently only refers to detecting potentially matching peaks. It is the actual peak matching that is most computationally expensive (every peak can only be matched once --> "assignment problem") and this is the only actual difference between Hungarian and Greedy.
- The actual score computation (using intensities of the matched peaks) is currently happening within the same functions and would have to be separated.

hechth:
@florian-huber That sounds good! At the moment, I am not very familiar with measuring the performance of functions in python and I will need some time to get a handle on that. So for the moment I'd say we can merge this small refactoring PR here for the tests and I will get back to a proper refactor according to what we discussed here some time later. I think it would be good to get matchms into a clean shape for now and make a new release so that we can also properly test things in the actual applications.

florian-huber (approved):
Thanks for the test refactoring, looks good to me!

PR #331, opened 2022-03-07
Added rtinseconds as possible field

niekdejonge:
In MGF files created by GNPS the retention time is sometimes stored as "rtinseconds"


hechth (approved):

hechth:
@niekdejonge I think we should also add that to the global key conversion table for consistency if we want this to be generally recognized.

florian-huber:
Thanks Niek and Helge for this PR! I added rtinseconds to the table and will merge it.

PR #337, opened 2022-03-24
Updated test cases and reduced smiles & inchi regexp strictness

### LINKED ISSUE #334: Is valid smiles functions filters out valid SMILES
The code in the function below actually filters out valid smiles for small molecules - I suspect the regexp check which comes before the check based on RDKit to be behaving wrong. For example `C` or `CF` are valid SMILES but are filtered by the function.

https://github.com/matchms/matchms/blob/a7b687d08da50f4a08fa92483399d5ea96ba601c/matchms/metadata_utils.py#L109-L133

### LINKED ISSUE #335: Valid InChI filter fails for molecules with no Carbon or Hydrogen layer
The regexp fails for things like molecular neon or chlorine as they don't have a carbon or hydrogen layer

https://github.com/matchms/matchms/blob/a7b687d08da50f4a08fa92483399d5ea96ba601c/matchms/metadata_utils.py#L99

hechth:
Close #334 
Close #335 

I reduced the regexp filters to detect wrong InChI and Smiles to a more lenient behaviour to accept small smiles such as `C` or `CF` which are valid and exist in certain databases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- metadata filtering: made prefilter check for SMILES and InChI more lenient, eventually resulting in longer runtimes but more accurate checks [#337](https://github.com/matchms/matchms/pull/337)
+
 ## [0.15.0] - 2022-03-09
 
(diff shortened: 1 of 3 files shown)


florian-huber (approved):
Good fix and nice cleaning and expansion of the tests! Thanks a lot.

niekdejonge (commented):
Looks good! I like the testing with parametrize, will start using the myself as well :)

PR #346, opened 2022-04-25
Added option to write appended to `msp` export

### LINKED ISSUE #343: Allow appending to file in export functions
When writing spectra to a file, it is currently only possible in `write` fashion, not in append - therefore appending spectral information to an already existing library is not so easy if one doesn't want to read the library into memory.

This could also be used for streaming output

Add the flags for opening the file as read/write/append as parameter to the export function (write as default).

hechth:
The `mgf` export already writes by default in append mode - should we change it so that the behaviour is consistent, meaning that the `mgf` export deletes file content if the `w` flag is specified and acts as before if the `a` flag is specified, or do we leave it as it is for the moment (always append?).

`msp` now offers write and append options.

Save as `json` is trickier as writing in two operations in append mode will result in two `json` objects stored in the same file - this can't be avoided without reading the file contents and then combining the object and dumping it. For this, the `save` functions needs to call the `load` function, which is not ideal IMHO.

Closes #343


hechth:
@florian-huber What would be your suggestion regarding the `json` format? Should we try to create some solution or just ignore it for the moment? I think few people work with json, so I'm not sure whether this is really necessary. I wanted to implement this feature for all exports but internally we'd only require the `msp` version of it to work.

florian-huber:
@hechth ... sorry, bit of delay from my side.
I forgot about the append mode for mgf.
I would suggest:
- Have the same default for msp and mgf (eighter append or write as default, no strong preference from my side).
- Skip json for the reasons you mention!

florian-huber (approved):
Looks good, thanks a lot @hechth !
Only thing I would change is to have the same default for save_to_mgf and save_to_msp ("w" or "a", choice is up to you 😅 ).

Once this is done and tests pass, this will be ready to be merged,

florian-huber:
This PR became stale. So I would now say we only add the functionality "mode" to the msp export and leave the rest for future PRs!

PR #353, opened 2022-06-24
Add export/import for Scores objects

### LINKED ISSUE #351: [feature request] Add exporting Scores to JSON
Currently, it's only possible to export [Spectrum](https://github.com/matchms/matchms/blob/master/matchms/Spectrum.py) objects. It would be nice if there was a functionality to also export [Scores](https://github.com/matchms/matchms/blob/master/matchms/Scores.py) objects to JSON files for later use, sharing, etc. The advantage of such an approach would be that the scores can be stored together with the individual spectra and related information.

A custom JSON encoder akin to [`SpectrumJSONEncoder`](https://github.com/matchms/matchms/blob/922f5afaef123a793194bdd74391027477cbb844/matchms/exporting/save_as_json.py#L46) seems like an optimal solution.

maximskorik:
* added functionality to import and export Scores objects to/from `json` and `pickle` files.
* added `ScoresBuilder` class to `Scores.py`. The class can be extended if more file formats are added. If anything changes in the way the data is stored, e.g., new scores implementation, the builder can be adjusted without breaking the `Scores` interface.
* defined equality operator `__eq__` for `Scores` class. The operator compares all `Scores` attributes; for `similarity_function` only the class name is compared. I think comparing the parameters of the similarity metrics is not vital in this case, as it seems unlikely someone would want to recalculate scores with the same function and parameters after loading scores.

### Usage example (**UPDATED 8.7.**):
```python
from matchms.importing import scores_from_json, scores_from_pickle
from matchms import calculate_scores, Scores
from matchms.similarity import CosineGreedy

scores = calculate_scores(references, queries, CosineGreedy())

# write a Scores object to disk as json or pickle file
scores.to_json(filename="scores.json")
scores.to_pickle(filename="scores.pkl")

# use readers from matchms.importing to read files
scores_json = scores_from_json("scores.json")
scores_pickle = scores_from_pickle("scores.pkl")

print(scores == scores_json == scores_pickle)
>>> True
```

Closes #351 

### **UPD 8.7.:**
* added test coverage for all metrics from `similarity` subpackage
* found and fixed a bug when deserializing scores with `FingerprintSimilarity`
* move `Scores` readers to `importing` subpackage
* splitted `Scores` writer into `to_json` and `to_pickle` methods

### **UPD 24.7.:**
* fixed unexpected behavior of tests caused by `test_missing_rdkit_module_error()` (1d43add & 110a173) (test-interdependency)
   - because `metadata_utils` module was never reloaded after `rdkit` dependency was set to `None`, all downstream tests that relied on rdkit-dependent methods from `metadata_utils` would fail.

### DIFF
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -217,7 +217,7 @@ class Scores:
                             self._scores[references_idx_sorted, selected_idx].copy()))
         return list(zip(self.references, self._scores[:, selected_idx].copy()))
 
-    def export_to_json(self, filename: str):
+    def to_json(self, filename: str):
         """Export the scores to a file.
 
         Parameters
@@ -228,7 +228,7 @@ class Scores:
         with open(filename, "w", encoding="utf-8") as f:
             json.dump(self, f, cls=ScoresJSONEncoder)
 
-    def export_to_pickle(self, filename: str):
+    def to_pickle(self, filename: str):
         """Export the scores to a file.
 
         Parameters
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -199,6 +199,30 @@ class Scores:
                             self._scores[references_idx_sorted, selected_idx].copy()))
         return list(zip(self.references, self._scores[:, selected_idx].copy()))
 
+    @classmethod
+    def import_from_file(cls, file_path: str) -> Scores:
+        """Import scores from a file.
+
+        Parameters
+        ----------
+        file_path
+            Path to the scores file.
+        """
+        raise NotImplementedError("Import from file is not implemented yet.")
+
+    def export_to_file(self, filename: str, file_format: str = "json"):
+        """Export the scores to a file.
+
+        Parameters
+        ----------
+        filename
+            Path to file to write to
+        file_format
+            File format to write to.
+            Default is "json".
+        """
+        raise NotImplementedError("Not implemented yet.")
+
     @property
     def scores(self) -> numpy.ndarray:
         """Scores as numpy array
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -223,7 +223,7 @@ class Scores:
         file_path
             Path to the scores file.
         """
-        return ScoresBuilder.from_json(file_path).build()
+        return ScoresBuilder().from_json(file_path).build()
 
     def export_to_file(self, filename: str, file_format: str = "json"):
         """Export the scores to a file.
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -225,7 +225,18 @@ class Scores:
         with open(file_path, 'rb') as f:
             scores_dict = json.load(f, object_hook=scores_json_decoder)
 
-        return scores_dict
+        if scores_dict["is_symmetric"] and scores_dict["queries"] is None:
+            scores_dict["queries"] = scores_dict["references"]
+
+        cls._validate_json_input(scores_dict)
+
+        scores = cls(references=numpy.array(scores_dict['references']),
+                     queries=numpy.array(scores_dict['queries']),
+                     similarity_function=BaseSimilarity(),
+                     is_symmetric=scores_dict['is_symmetric'])
+        scores._scores = scores_dict['scores']
+
+        return scores
 
     def export_to_file(self, filename: str, file_format: str = "json"):
         """Export the scores to a file.
@@ -276,3 +287,10 @@ class Scores:
               [0.2 1. ]]
         """
         return self._scores.copy()
+
+    @staticmethod
+    def _validate_json_input(scores_dict: dict):
+        if {"__Scores__", "similarity_function", "is_symmetric", "references", "queries", "scores"} != scores_dict.keys():
+            raise ValueError("Scores JSON file does not match against the schema.\n\
+                             Make sure JSON file contains the following keys:\n\
+                             ['__Scores__', 'similarity_function', 'is_symmetric', 'references', 'queries', 'scores']")
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -81,6 +81,17 @@ class Scores:
         self._scores = numpy.empty([self.n_rows, self.n_cols], dtype="object")
         self._index = 0
 
+    def __eq__(self, other):
+        if isinstance(other, Scores):
+            return self._scores == other._scores and\
+                self.similarity_function == other.similarity_function and\
+                self.is_symmetric == other.is_symmetric and\
+                self.n_rows == other.n_rows and\
+                self.n_cols == other.n_cols and\
+                self.references == other.references and\
+                self.queries == other.queries
+        return False
+
     def __iter__(self):
         return self
 
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -230,7 +230,11 @@ class Scores:
 
     @classmethod
     def import_from_pickle(cls, file_path: str) -> Scores:
-        """Import scores from a pickle file."""
+        """Import scores from a pickle file.
+
+        WARNING: Pickle files are not secure and may execute malicious code. Make sure that you are importing a pickle
+        file from a trusted source.
+        """
         with open(file_path, "rb") as f:
             return pickle.load(f)
 
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -218,7 +218,7 @@ class Scores:
         return list(zip(self.references, self._scores[:, selected_idx].copy()))
 
     def to_json(self, filename: str):
-        """Export the scores to a file.
+        """Export :py:class:`~matchms.Scores.Scores` to a JSON file.
 
         Parameters
         ----------
@@ -229,7 +229,7 @@ class Scores:
             json.dump(self, f, cls=ScoresJSONEncoder)
 
     def to_pickle(self, filename: str):
-        """Export the scores to a file.
+        """Export :py:class:`~matchms.Scores.Scores` to a Pickle file.
 
         Parameters
         ----------
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -257,6 +257,15 @@ class Scores:
         else:
             raise ValueError(f"File format '{file_format}' is not supported.")
 
+    def to_dict(self) -> dict:
+        """Return a dictionary representation of scores."""
+        return {"__Scores__": True,
+                "similarity_function": self._encode_similarity_function(self.similarity_function),
+                "is_symmetric": self.is_symmetric,
+                "references": [reference.to_dict() for reference in self.references],
+                "queries": [query.to_dict() for query in self.queries] if not self.is_symmetric else None,
+                "scores": self.scores.tolist()}
+
     @property
     def scores(self) -> numpy.ndarray:
         """Scores as numpy array
--- a/matchms/Scores.py
+++ b/matchms/Scores.py
@@ -5,6 +5,7 @@ from deprecated.sphinx import deprecated
 from matchms.importing.load_from_json import scores_json_decoder
 from matchms.exporting.save_as_json import ScoresJSONEncoder
 from matchms.similarity.BaseSimilarity import BaseSimilarity
+from matchms.similarity import _get_similarity_function_by_name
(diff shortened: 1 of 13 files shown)


matchms/Scores.py
```diff
@@ -1,6 +1,12 @@
 from __future__ import annotations
+import json
+import pickle
```
hechth, line 3: I think these should also be included in the dependencies in the `setup.py` file?

matchms/Scores.py
```diff
@@ -233,3 +291,76 @@ def scores(self) -> numpy.ndarray:
...
+        """
+        similarity_function_class = get_similarity_function_by_name(similarity_function_dict.pop("__Similarity__"))
+        return similarity_function_class(**similarity_function_dict)
+
+    @staticmethod
+    def _validate_json_input(scores_dict: dict):
+        if {"__Scores__", "similarity_function", "is_symmetric", "references", "queries", "scores"} != scores_dict.keys():
+            raise ValueError("Scores JSON file does not match the expected schema.\n\
+                             Make sure the file contains the following keys:\n\
+                             ['__Scores__', 'similarity_function', 'is_symmetric', 'references', 'queries', 'scores']")
```
hechth, line 357: @florian-huber should this move to a separate file (since it is a standalone class)?

matchms/exporting/save_as_json.py
```diff
@@ -39,20 +40,47 @@ def save_as_json(spectrums: List[Spectrum], filename: str):
...
+    with open(filename, "w", encoding="utf-8") as fout:
         json.dump(spectrums, fout, cls=SpectrumJSONEncoder)
 
 
+def _convert_spectrum_into_dict(spectrum: Spectrum):
+    """Convert :py:class:`~matchms.Spectrum.Spectrum` into a dictionary"""
+    peaks_list = numpy.vstack((spectrum.peaks.mz, spectrum.peaks.intensities)).T.tolist()
+    spectrum_dict = {key: spectrum.metadata[key] for key in spectrum.metadata}
+    spectrum_dict["peaks_json"] = peaks_list
+    return spectrum_dict
```
hechth: Maybe this should rather be a method of the `Spectrum` class than an external method manipulating the class?

matchms/exporting/save_as_json.py
```diff
@@ -39,20 +40,47 @@ def save_as_json(spectrums: List[Spectrum], filename: str):
...
+        # do isinstance(o, Scores) without importing matchms.Scores
+        if class_name == "Scores":
+            scores = copy.deepcopy(o)
+
+            scores_dict = {"__Scores__": True,
+                           "similarity_function": self._encode_similarity_function(scores.similarity_function),
+                           "is_symmetric": scores.is_symmetric,
+                           "references": [_convert_spectrum_into_dict(reference) for reference in scores.references],
+                           "queries": [_convert_spectrum_into_dict(query) for query in scores.queries] if not scores.is_symmetric else None,
+                           "scores": scores.scores.tolist()}
```
hechth: Again - shouldn't this be a method of the scores class? If the scores class will change, this method will have to change - this would introduce some coupling between this method and the scores class. Also, the scores class is the only reason for this method to change - given that the `dict` representation is quite constant.

matchms/Scores.py
```diff
@@ -199,6 +217,46 @@ def scores_by_query(self, query: QueriesType, sort: bool = False) -> numpy.ndarr
...
+            File format to write to. Supported formats are: "json", "pkl". Default is "json".
+        """
+        if file_format == "json":
+            with open(filename, "w", encoding="utf-8") as f:
+                json.dump(self, f, cls=ScoresJSONEncoder)
+        elif file_format == "pkl":
+            with open(filename, "wb") as f:
+                pickle.dump(self, f)
+        else:
+            raise ValueError(f"File format '{file_format}' is not supported.")
```
hechth: I think these functions should move to the import and export modules - the `Scores` class doesn't need to know about how it is exported - neither does the `Spectrum` class. It only needs to know how it is represented as a dict - then there can be a function `scores_to_pickle` or so which actually performs the output.

The actual `Scores` class need not know the details of how it is stored - this is what the exporter should know, given the interface of the `Scores`, so also not knowing all details of the `Scores` class.

matchms/importing/load_from_json.py
```diff
@@ -112,3 +112,12 @@ def parse_fieldname(key):
...
     return None
+
+
+def scores_json_decoder(dct):
+    """
+    Object_hook function to convert JSON dictionary with :py:class:`~matchms.Score.Score` object into a python dictionary.
+    """
+    if "__Scores__" not in dct and "__Similarity__" not in dct:
+        return dict2spectrum(dct, metadata_harmonization=False)
+    return dct
```
hechth, line 123: I think `dct` is a parameter name that is too error prone!

matchms/similarity/__init__.py
```diff
@@ -40,3 +40,23 @@
...
+    """
+    names = __all__
+    functions = [CosineGreedy, CosineHungarian, FingerprintSimilarity, IntersectMz, MetadataMatch, ModifiedCosine,
+                 NeutralLossesCosine, ParentMassMatch, PrecursorMzMatch]
+
+    assert similarity_function_name in names, f"Unknown similarity function: {similarity_function_name}"
+    assert len(names) == len(functions), "Number of similarity functions and names do not match"
+
+    mapper = dict(zip(names, functions))
+    return mapper[similarity_function_name]
```
hechth, line 62: Is this function actually used somewhere?

hechth (changes requested):
I think the structure of where which function is located can be improved to reduce the introduced coupling and keep the classes flexible and independent.

matchms/Scores.py
maximskorik: To my understanding, `setup.py` should include packages that have to be installed for a package to run properly. Since `json` and `pickle` are in Python's standard library, I don't think they belong in `setup.py`.

matchms/exporting/save_as_json.py
maximskorik: Good point. I strived to minimum changes in the original code, so I just wrapped what already was in `save_as_json.py` into a function. @florian-huber what do you think of moving the method to the `Spectrum`?

matchms/importing/load_from_json.py
maximskorik: The Python's documentation uses that name in the example, so it doesn't occur to me as erroneous:
https://docs.python.org/3/library/json.html?highlight=json#module-json
```python
import json
def as_complex(dct):
    if '__complex__' in dct:
        return complex(dct['real'], dct['imag'])
    return dct
```

matchms/similarity/__init__.py
maximskorik: Here: https://github.com/maximskorik/matchms/blob/cd9cd7247e84fc3af96018ca3382f0f5e5370c41/matchms/Scores.py#L358

```python
    @staticmethod
    def _construct_similarity_function(similarity_function_dict: dict) -> BaseSimilarity:
        """
        Construct similarity function from its serialized form.
        """
        similarity_function_class = get_similarity_function_by_name(similarity_function_dict.pop("__Similarity__"))
        return similarity_function_class(**similarity_function_dict)
```

matchms/exporting/save_as_json.py
maximskorik: Resolved in efc55e1

matchms/exporting/save_as_json.py
maximskorik: I agree. Making it into a Scores method is much better. Implemented that in 1613f8f and 626aa5a.

matchms/importing/load_from_json.py
hechth: Oh wow :D Different people have different opinions - but okay, then this seems to be generally accepted.

maximskorik: I get the point. I'd probably come up with another name hadn't I seen this example.

matchms/exporting/save_as_json.py
florian-huber: I would agree that this belongs to `Spectrum`.

matchms/exporting/save_as_json.py
florian-huber: Nice, making this part of `Scores` would be my preference, too!

matchms/Scores.py
```diff
@@ -199,6 +217,55 @@ def scores_by_query(self, query: QueriesType, sort: bool = False) -> numpy.ndarr
...
+
+    def to_dict(self) -> dict:
+        """Return a dictionary representation of scores."""
+        return {"__Scores__": True,
+                "similarity_function": self.similarity_function.to_dict(),
+                "is_symmetric": self.is_symmetric,
+                "references": [reference.to_dict() for reference in self.references],
+                "queries": [query.to_dict() for query in self.queries] if not self.is_symmetric else None,
+                "scores": self.scores.tolist()}
```
florian-huber, line 249: This will work for small and medium `Scores` objects, so fine for now I would say.
But I expect this to be slow for huge arrays.

More importantly, as soon as we (finally...) manage to move to sparse arrays this will need substantial changes.
We already know that this step will break things, and it will break here, too (Scores saved with the here presented version will not import in a future version of matchms with sparse representations).

matchms/Scores.py
florian-huber: Not sure either. In principle I am fine with keeping it here since it belongs to `Scores`. But we decided differently for `SpectrumBuilder`... so a bit inconsistent.

matchms/Scores.py
florian-huber: Well, quick afterthought: We could of course implement a `.tolist()` method in our sparse scores array...

matchms/Scores.py
hechth: I mean the SpectrumBuilder is part of the tests and not really meant to be used in the normal code - the ScoresBuilder though, is.

matchms/Scores.py
hechth: True - but I think this is something we can address when we get to this point.

florian-huber: True, was more a note to myself... :)

matchms/Scores.py
florian-huber: That's right. I would be fine either way and have no clear preferences here. If you have a preference for a separate file, please go ahead.

maximskorik: My main objection against placing the builder into a separate file was that I did not want to pollute matchms core directory with more files. One solution may be to create a `builders` directory, which for now will host `ScoresBuilder` and `SpectrumBuilder` classes. Any thoughts on that?

matchms/Scores.py
maximskorik: I moved `Scores` reading into `importing` subpackage in 21c28d9.

As to writing, I think both options are acceptable and used (e.g., Pandas uses `object.persist_to_disk()` approach, whereas Numpy uses `disk.persist(object)`). I initially tried to mimic Pandas' approach and still think it's a valid option. Nonetheless, after your suggestion I reconsidered my implementation and decided to stick to more simplistic `scores.to_json()` and `scores.to_pickle()` methods (fc04f0c) as they better resemble those of Pandas. Another point that led me to this decision is that `Spectrum's` `save_as_json` can be used to write multiple spectra to a single file. Contrary to that I don't think scores of different spectra combinations belong to the same file, hence the method will always be used on a single object.

Unless there are more objections or intricacies that I'm not aware of, I'd leave it as it is :)

hechth (approved):
@florian-huber IMHO this can be merged :).

Great work @maximskorik !

hechth:
@florian-huber Can we merge this?

florian-huber:
@hechth and @maximskorik : great work! Looks good to me, so I will merge it.
(sorry for the delay... I was on some fully offline holidays ;) )

florian-huber:
I was too fast, this should have been merged to a local branch first to see if all tests pass (they now fail). So this needs some further fixing...

maximskorik:
> I was too fast, this should have been merged to a local branch first to see if all tests pass (they now fail). So this needs some further fixing...

I see. We could revert the PR and I'll work on this if you point me how to reproduce the error(s).

Issue #356, opened 2022-08-11
Tests failing because rdkit import


florian-huber:
After the recent merge #353 the tests fail. Main reason seems to be that rdkit is used which is not included in pip installation (so tests should skip if rdkit is not precent). The tests will then run later in the CI when the full conda installation is used.

florian-huber:
Fixed with #357

PR #357, opened 2022-08-11
Skip test using rdkit for first CI round

florian-huber:
--> issue #356


maximskorik:
@florian-huber, thank you for taking care of that!

florian-huber:
This was only tiny linting to make the tests pass. Should work now.

PR #368, opened 2022-10-02
Pipeline (using sparsestack)

florian-huber:
Pipeline project became too big. So sparsestack was moved to its own package.

- sparsestack is now own package --> https://github.com/florian-huber/sparsestack
- MetadataMatch score was adapted to sparse style
-

TODOs:
- [x] ParentmassMatch needs adaptation to sparse
- [x] PrecursorMzMatch needs adaptation to sparse

PR #370, opened 2022-10-14
Pipeline merge

florian-huber:
#368 has become too big to merge easily.
This PR is hence an in-between step to move Pipeline and sparsity into master.


matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+            spectrums_queries += _spectrum_importer(query_file)
+        self.spectrums_queries += spectrums_queries
+        if reference_files is None:
+            self.is_symmetric = True
+            self.spectrums_references = self.spectrums_queries
+        else:
+            spectrums_references = []
+            for reference_file in reference_files:
+                spectrums_references += _spectrum_importer(reference_file)
+            self.spectrums_references += spectrums_references
```
hechth: This piece of code together with the map to the individual import functions could be moved to a dedicated SpectrumImporter class or so that would wrap the whole import process, independent from the format - this would also make it easier for scripts which accept different file formats as this responsibility - deciding which import function to use based on the filetype - would be handled by the importer.

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+from matchms import calculate_scores
+from matchms.logging_functions import (add_logging_to_file,
+                                       set_matchms_logger_level)
+
+
+_importing_functions = {"json": msimport.load_from_json,
+                        "mgf": msimport.load_from_mgf,
+                        "msp": msimport.load_from_msp,
+                        "mzml": msimport.load_from_mzml,
+                        "mzxml": msimport.load_from_mzxml}
```
hechth, line 20: This piece of could code also then move to the importer class.

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+                self.workflow = ordered_load(file, yaml.SafeLoader)
+            if self.workflow["filtering_refs"] == "filtering_queries":
+                self.workflow["filtering_refs"] = self.workflow["filtering_queries"]
+        if "logging" not in self.workflow:
+            self.workflow["logging"] = {}
+        if self.logging_level is None:
+            self.logging_level = "WARNING"
+
+    def import_workflow_from_yaml(self, config_file):
+        self._initialize_workflow_dict(config_file)
```
hechth: Could possibly move to a `ConfigLoader` class or interface.

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+        # Processing
+        self.write_to_logfile("--- Processing spectra ---")
+        for step in self.filter_steps_queries:
+            self.write_to_logfile(f"-- Processing step: {step} --")
+        for spectrum in tqdm(self.spectrums_queries,
+                             disable=(not self.progress_bar),
+                             desc="Processing query spectrums"):
+            for step in self.filter_steps_queries:
+                spectrum = self.apply_filter(spectrum, step)
+            self.spectrums_queries = [s for s in self.spectrums_queries if s is not None]
```
hechth, line 147: Could be nicely packed into some utility function like `filter_none` as this is very generic functionality.

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def _apply_score_masking(self, computation):
+        if len(computation) == 1:
+            name = self.scores.score_names[-1]
+            self.scores.filter_by_range(inplace=True, name=name)
+        elif "name" not in computation[1]:
+            name = self.scores.scores.score_names[-1]
+            self.scores.filter_by_range(inplace=True, name=name, **computation[1])
+        else:
+            self.scores.filter_by_range(inplace=True, **computation[1])
```
hechth: I think this function definitely needs some explanation in form of a docstring or better function names - why does the string have to contain "name"?

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+        else:
+            new_scores = similarity_measure.sparse_array(references=self.spectrums_references,
+                                                         queries=self.spectrums_queries,
+                                                         idx_row=self.scores.scores.row,
+                                                         idx_col=self.scores.scores.col,
+                                                         is_symmetric=self.is_symmetric)
+            self.scores.scores.add_sparse_data(self.scores.scores.row,
+                                               self.scores.scores.col,
+                                               new_scores,
+                                               similarity_measure.__class__.__name__)
```
hechth: Does this mean that if `i==0`, then the non-sparse computation is used, otherwise the sparse one?

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+        for computation in self.score_computations:
+            if not isinstance(computation, list):
+                computation = [computation]
+            if isinstance(computation[0], str) and computation[0] in _masking_functions:
+                continue
+            if isinstance(computation[0], str) and computation[0] in _score_functions:
+                continue
+            if callable(computation[0]):
+                continue
+            raise ValueError(f"Unknown score computation: {computation[0]}.")
```
hechth: Could be moved to potential `ConfigLoader` class as well

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+def ordered_dump(data, stream=None, dumper=yaml.SafeDumper, **kwds):
+    class OrderedDumper(dumper):
+        pass
+
+    def _dict_representer(dumper, data):
+        return dumper.represent_mapping(
+            yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG,
+            data.items())
+    OrderedDumper.add_representer(OrderedDict, _dict_representer)
+    return yaml.dump(data, stream, OrderedDumper, **kwds)
```
hechth, line 404: This seems to be also related to the config reading part.

matchms/Scores.py
```diff
@@ -247,15 +294,53 @@ def to_pickle(self, filename: str):
...
+            Define operator to be used to compare against `low`. Default is '>'.
+            Possible choices are '>', '<', '>=', '<='.
+        below_operator
+            Define operator to be used to compare against `high`. Default is '<'.
+            Possible choices are '>', '<', '>=', '<='.
+        """
+        if inplace is True:
+            self._scores = self._scores.filter_by_range(**kwargs)
+            return None
+        return self._scores.filter_by_range(**kwargs)
```
hechth, line 355: Does the inplace option allow for specific steps or things within the pipeline to be possible, or is there no specific reason why this option is possible?

hechth (commented):
This brings a lot of changes and I'm very excited to try the new functionality! All my comments are more on the nitpicking and possible future improvements side of things, so overall I think this looks great. Is there anything that we still have to develop or fix or something that still has to be covered with tests?

## Pipeline Class
It seems like this class currently has 3 responsibilities
* Loading the data, independently from the format
* Reading and parsing the config file -> instantiating the function objects required for the steps
* Executing the function from the workflow

Maybe these responsibilities could be divided into 3 classes, a `SpectrumImporter`, a `ConfigLoader` and a `PipelineRunner` - the `Pipeline` class being the one that controls those. This would make it easier in the future to add more complexity and functionality.

But overall the class looks really good, just needs a bit more documentation.

## Tests
The tests for the similarity scores are now covered in the tests for the `spectrum_similarity_functions`?

matchms/Pipeline.py
```diff
@@ -0,0 +1,367 @@
...
+    The second way to define a pipeline is via a Python script. The following code is an
+    example of how this works:
+
+    .. code-block:: python
+
+        from matchms import Pipeline
+
+        pipeline = Pipeline("my_config_file.yaml")
+        pipeline.run()
+
```
maximskorik: There seems to be a duplicated example in docstrings.

matchms/Pipeline.py
florian-huber: It means that for i==0 a new array is computer (would also be sparse usually). For `i > 0` the new data will be stacked into the same sparse array.

matchms/Pipeline.py
florian-huber: Thanks, should be fixed now.

PR #371, opened 2022-10-19
Refactor metadata harmonization to improve msp/mgf reading performance

### LINKED ISSUE #367: [performance improvement] Reduce spectra-loading time by slight refactoring of Spectrum class
## Problem Description
In light of the recent discussions regarding large reading time for big spectra files, I've profiled `.msp` and `.mgf` spectra readers to see if there are bottlenecks or parts that can be improved.

Unsurprisingly, I've found out that reading spectra with **metadata harmonization** is slower than without it. Upon inspecting some profiles I found a peculiar step that takes a lot of time and seems unnecessary, that being the reinitialization of `Metadata` class: once during `Spectrum` initialization, another time just after that in `_apply_metadata_harmonization()` if `harmonization==True` (see the pic below).

image
Profiles of reading Wiley spectral library from `.msp` file. On the left pic, metadata harmonization was disabled; the profile on the right shows reading with metadata harmonization. Notice, that when harmonization is enabled `__init__` in `Metadata.py` gets called twice. 

As evident from the picture, refactoring `Spectrum` class to construct `Metadata` object only once could in theory **save up to 25% of reading time** (`load_from_msp` took 290 seconds of which the second construction of `Metadata` took 70.5 seconds).

## Possible Solution
The most simplistic solution would be to slightly alter `Spectrum` constructor and `_apply_metadata_harmonization` method:
```python
def __init(..., metadata, metadata_harmonization):
   ...
    if metadata_harmonization is True:
        self._apply_metadata_harmonization(metadata)
    else:
        self._metadata = Metadata(metadata)
   ...

def _apply_metadata_harmonization(self, metadata):
    metadata_filtered = _interpret_pepmass_metadata(metadata)
    if metadata_filtered.get("ionmode") is not None:
        metadata_filtered["ionmode"] = metadata.get("ionmode").lower()
    metadata_filtered = _add_precursor_mz_metadata(metadata_filtered)
    ...
    self._metadata = Metadata(metadata_filtered)
```
However, with this exact approach, some tests currently fail so either the implementation has to be polished or changed, the tests have to be adjusted, or both.

The `integration-tests` are passing with this implementation though. I've tried reading the data with these changes and the improvements in reading speed were as shown below:

profiles_comparison

## Starting points
If we agree that this is worth implementing I could work on this sometime next month or would be happy if someone else wants to take care of that.

## Additional context
Jupyter notebook with profiles analysis: https://github.com/maximskorik/matchms/blob/readers_profiling/benchmarking/profile_analysis.ipynb
The profile data and all code that I used for analysis can be found in `benchmarking` folder of `readers_profiling` branch on my MatchMS fork: https://github.com/maximskorik/matchms/tree/readers_profiling

maximskorik:
* moved `_apply_metadata_harmonization` to the `Metadata` class since it's only bound to that class' data
* renamed and slightly refactored `_apply_metadata_harmonization`
* renamed `harmonize_metadata` method

### Testing

* no new features have been added, so there are no new tests
* `integration_tests` and `tests` passed locally

### Performance improvement

Below is a comparison of refactored data reading compared to the current `master` branch on my machine (mac air m1):
performance
*The error bars correspond to deviation in mgf-reading (lower bound) and msp-reading (upper bound). Time in seconds.*

Closes #367


maximskorik:
@florian-huber cloud you check the CI, please? I'm not sure what's wrong with the SonarCloud check.

florian-huber (approved):
Looks very good, thanks a lot @maximskorik. Makes indeed more sense to have this functionality in `Metadata` I believe.

PR #378, opened 2022-12-09
Configure SonarCloud CI to scan PRs from forks

maximskorik:
### Problem
The current implementation of SonarCloud job in the `CI Build` workflow disallows running scans on PRs from forks. The CI fails because it runs in an **unprivileged** mode and has no access to the secrets. 

### Workaround
Split build and scan jobs into two separate workflows **unprivileged** build and **privileged** scan:
   1. The build workflow (`CI_build.yml`) is triggered with `push` and `pull_request` events, which run in **unprivileged** mode and thus have no access to secrets. The workflow generates coverage data and PR metadata and stores it as an artifact for the following workflow.
   2. The `SonarCloud.yml` workflow uses the [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) trigger and runs upon the successful finish of the build workflow. `workflow_run` event is **privileged** and thus has access to the secrets, but it will only run the CI actions from the **default** branch of the repo where it's executed, thus disallowing gaining access to secrets via changing the CI code. The workflow will fetch coverage data and PR metadata from the build workflow and run SonarCloud scan.

### Sources
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/30

### DIFF
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -49,12 +49,6 @@ jobs:
             xunit-result.xml
       - name: Check whether import statements are used consistently
         run: isort . --check-only --diff
-      - name: SonarCloud Scan
-        if: github.repository == 'matchms/matchms'
-        uses: sonarsource/sonarcloud-github-action@master
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
 
   build_pypi:
     name: Test pypi build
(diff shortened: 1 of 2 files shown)


hechth (approved):
@florian-huber looks good to me - do you see any security related issues here that would have to be resolved first?

florian-huber (approved):
Looks good to me. Thanks a lot @maximskorik !

Issue #379, opened 2022-12-19
Make Scores JSON-serializer work with with ML-based similarity metrics


maximskorik:
### Problem
[JSON encoder](https://github.com/matchms/matchms/blob/1284caab52f948115b740b9dd1a73f6c5c5d1d21/matchms/exporting/save_as_json.py#L56) for objects of `Scores` class fails to serialize scores computed with ML-based similarity metrics such as Spec2Vec or MS2DeepScore because the models lack dictionary representation.

### Possible Solution
Adjust the encoder to handle such cases by not attempting to encode entire models but only capturing the models' metadata instead.

florian-huber:
Sounds like a good solution.
It would in principle also be possible to store model weights as json, but probably more work.

MS2DeepScore --> Export of architecture as json: https://www.tensorflow.org/guide/keras/save_and_serialize
or maybe also here for model and weights: https://towardsdatascience.com/saving-and-loading-keras-model-42195b92f57a

However, for Gensim (Spec2Vec) I wouldn't know how to do that.

hechth:
I think @maximskorik already implemented a JSON serialization for Gensim models for Spec2Vec if I am not mistaken? At least the metadata part.

maximskorik:
I started working on this issue and realized it seems to have been fixed in #370. The problem shouldn't arise as the new JSON representation of `Scores` doesn't include a similarity function, but delegates this to stacked sparse array objects. The stacked array only stores the names of the similarity score methods, which is how I'd have approached the issue too. I tried exporting and importing scores computed with Spec2Vec and it worked.

hechth:
@florian-huber @maximskorik so I think this can then be closed?

maximskorik:
I don't see anything else to do regarding this issue but to use/test the functionality. Also, per #380, MS2DeepScore has been updated to ensure compatibility. I think it's safe to close the issue and reopen it if the problem reoccurs.

florian-huber:
Thanks @hechth and @maximskorik for taking care of this!

PR #381, opened 2023-01-20
Update docstrings in Scores class

maximskorik:
* small linting and fixed docstrings in `Scores.py`

I did some linting as I started working on #379. I soon realised the issue had been fixed at that point, hence that's all there is in this PR.


florian-huber (approved):
Thanks for linting @maximskorik

PR #384, opened 2023-01-27
Add updated tutorial

florian-huber:


hechth:
@florian-huber what would still be needed to update this tutorial?

niekdejonge:
@florian-huber What is the status of the tutorial. Does this already use the new Pipeline? Is it worth starting from this tutorial, or do you think it is best to just start over from scratch to make a pipeline based tutorial?

florian-huber:
Outdated, will be closed.

PR #393, opened 2023-03-28
Implemented MSP reading for GOLM style MSP and MSL style files

### LINKED ISSUE #392: Support reading old NIST and GOLM MSP formats
The old NIST and AMDIS software sometimes uses a different MSP format which puts the peak tuples into braces like this: `(55 102) (104 100)` - this is currently not supported using matchms.

This also applies for the GOLM metabolome database which has a `55:102` notation for the peak tuples.

Update the MSP reader to use regular expressions or other methods to detect the style and read the data accordingly - one way would also be to just try after removing comments to extract all numerical values and if they are present in an even number, assume that each pair is an m/z intensity pair representing a peak.

Example spectra:

```
Name: M000000_A097001-101-xxx_NA_959,45_PRED_VAR5_ALK_Unknown#bth-pae-001
Num Peaks: 50
70:10 76:35 77:1000 78:110 79:42 
80:4 81:7 86:6 87:5 88:21 
90:42 91:40 97:15 100:31 103:213 
104:27 105:10 107:48 108:6 110:41 
112:14 118:39 120:79 121:8 122:4 
130:33 134:589 135:55 136:25 137:3 
138:2 140:83 141:4 143:104 144:8 
160:4 184:610 185:58 186:24 187:2 
198:7 199:3 214:294 215:33 216:12 
217:1 228:3 229:109 230:12 231:5 
```

```
NAME:G3P
CASNO:TUBE2_~1-N1010
RI:1586.2
RW:
RT:9.648
COMMENT: RI=1586.2,   9.6481 min TUBE2_28-01-2020_18-48-00|RI:1586.2
SOURCE:C:\USERS\UTILISATEUR\DESKTOP\METABOLOME\calib\standard.msl
NUM PEAKS:  110 
( 52    3) ( 55  136) ( 56   14) ( 59   96) ( 61   31) 
( 62    2) ( 63    1) ( 66    1) ( 68    4) ( 70   11) 
( 73 1000) ( 75  303) ( 76   22) ( 77   12) ( 80    4) 
( 82   19) ( 84   31) ( 85    7) ( 86    7) ( 87    3) 
( 89   84) ( 90    8) ( 91    8) ( 92    1) ( 95    1) 
( 96    1) ( 98    4) (100    6) (101    4) (102    4) 
(103    4) (112   33) (113    4) (115    7) (116    2) 
(117    4) (118    1) (119    2) (126   19) (127    5) 
(128    5) (129    7) (131    7) (133   22) (134    3) 
(135    3) (138    1) (139    1) (140    1) (142    3) 
(143    2) (144    8) (147  172) (148   27) (149   17) 
(150    2) (154   14) (156   65) (157    9) (158    6) 
(159    2) (160    2) (161    1) (163    3) (164    1) 
(168    1) (170   29) (171    5) (172    9) (173    4) 
(174    3) (175    1) (176    1) (182    4) (183    3) 
(184    2) (186   30) (187    4) (188    2) (189    1) 
(190    1) (198   90) (199   13) (200    7) (202   20) 
(203    3) (204    1) (214    1) (216    1) (217    1) 
(229   17) (230    3) (231    1) (244   12) (245    3) 
(246    2) (260    1) (272    3) (273    2) (274    1) 
(288   28) (289    6) (290    3) (291    1) (304   23) 
(305    5) (306    3) (307    1) (319    2) (320    1) 
```

hechth:
Reworked the MSP reader to have a bit more generic logic whoch complies with all tests but also covers 2 new use cases.

We can now read the GOLM database MSP and MSL files created with AMDIS.

Closes #392


hechth:
@florian-huber seems like the windows CI is somehow broken because of PIL - any idea where this comes from? Or should we just ignore it?

matchms/importing/load_from_msp.py
```diff
@@ -140,11 +146,14 @@ def parse_metadata(rline: str, params: dict):
...
+                       ] = splitted_line[1].strip()
     else:
         params[splitted_line[0].lower()] = splitted_line[1].strip()
 
 
 def contains_metadata(rline: str) -> bool:
     """ Check if line contains Spectrum metadata."""
-    return ':' in rline
+    has_colon = ':' in rline
+    is_golm_peaks = re.match(r"(\d+:{1}\d+)", rline) is not None
```
niekdejonge: Could you add a short description above this line? If I understand correctly the golm format can have ":" in a line which is a peak and not metadata right?

matchms/importing/load_from_msp.py
```diff
@@ -76,58 +76,64 @@ def parse_msp_file(filename: str) -> Generator[dict, None, None]:
...
 
-def get_peak_tuples(rline: str) -> Iterator[str]:
-    """ Splits line at ';' and performs additional string cleaning. """
-    tokens = filter(None, rline.split(";"))
-    peak_pairs = map(lambda x: x.lstrip().rstrip(), tokens)
-    return peak_pairs
 
 
-def get_peak_comment(rline: str) -> str:
+def _parse_line(rline: str):
```
niekdejonge: Maybe rename to _parse_line_with_peaks. (this line should always contain mz and intensities right?

niekdejonge (approved):
Looks good. I left some small comments on documentation and naming clarity.

matchms/importing/load_from_msp.py
hechth: Yeah exactly. GOLM specifies parks as mz:int, so for example 55:1290. It conflicts with the previous recognition of a metadata line. The regex though should only find number:number sections in strings, which are more than unlikely to occur in comments I think.

matchms/importing/load_from_msp.py
hechth: True - will address that. A line without any mz and intensities should throw an error I guess, also to avoid silently ignoring peaks or not reading spectra properly.

florian-huber:
> @florian-huber seems like the windows CI is somehow broken because of PIL - any idea where this comes from? Or should we just ignore it?

I not sure either where this comes from. I will have a look at it!

florian-huber (approved):
Looks good to me @hechth !
(+ thanks @niekdejonge, too)

I will have another look at the CI runs that failed for Windows.

florian-huber:
@hechth the issue with the Windows CI had to do with `pillow`. I tried a few things and the easiest seemed to simply exclude the version that caused issues (9.4.0) --> #394
Should hence be fixed by now in the master branch.

matchms/importing/load_from_msp.py
hechth: Adressed in https://github.com/matchms/matchms/pull/393/commits/7ef2c57896b08fa74c8128d038f65eed9657c820

matchms/importing/load_from_msp.py
hechth: Adressed in https://github.com/matchms/matchms/pull/393/commits/7ef2c57896b08fa74c8128d038f65eed9657c820

matchms/importing/load_from_msp.py
```diff
@@ -76,58 +76,72 @@ def parse_msp_file(filename: str) -> Generator[dict, None, None]:
...
-    """ Get the m/z and intensity value from the line containing the peak information. """
-    splitted_line = peak.split(maxsplit=2)
-    mz = float(splitted_line[0].strip())
-    intensity = float(splitted_line[1].strip())
-    return mz, intensity
+                continue
+
+            mz, ints, comment = _parse_line_with_peaks(rline)
+
+            masses = np.append(masses, mz)
```
mobiusklein, line 83: Random kibitzing: `np.append` reallocates the array on every single call because `np.ndarray`'s buffer must exactly match its shape for all the fancy striding and indexing tricks to work efficiently. Most linear containers use an amortized pessimistic resizing algorithm, reserving more memory than they need to.

You want to use Python's `array` module to make `array.array('d')` which does use the amortized resizing rule but also supports the buffer protocol. You would grow the buffer in the `array.array`, and then at the very end pass it to `np.frombuffer` or `np.asarray` with dtype specified and `copy=False`. This will avoid re-copying those arrays and give you a NumPy array over that same memory.

hechth: @mobiusklein This is very interesting, I didn't know that! Might be good to change it here since we kind of know the number of peaks to be read in advance given the num_peaks variable. So I think already allocating the right sized array would be the fastest?

Thank you very much for the hint!

PR #396, opened 2023-04-12
Load all file types

niekdejonge:
Added a function to infer filetype when loading in functions. 

For most other tools using Matchms a function is needed that just uses the extension to assume file type. This has now been added. 

@florian-huber @hechth Do you agree with this implementation?


hechth (changes requested):
I think this is a cool feature to have a central loading function - maybe we can add the extension as an optional parameter? Galaxy for example renames everything as .dat and then you can get the filetype from the metadata, I think some databases might work in a similar fashion - so this would greatly improve usability and only add a little bit of complexity.

hechth:
@niekdejonge I made some small updates to your PR that contains the changes I mentioned - please let me know if you agree with those :)

@florian-huber otherwise I think this is good and can be merged. We could also some point implement a wrapper around the export process.

niekdejonge:
@hechth Sorry for the later reply! Thanks a lot for already implementing the changes, I will check them now.

matchms/importing/__init__.py
```diff
@@ -30,6 +31,8 @@
     "load_from_mzml",
     "load_from_mzxml",
     "load_from_usi",
+    "load_spectra",
```
niekdejonge, line 34: I was afraid this might cause some circular import issues, but if the tests pass that is a good addition indeed!

tests/test_load_spectra.py
```diff
@@ -0,0 +1,30 @@
...
+from matchms.importing.load_spectra import load_spectra
+
+
+def test_load_spectra_unknown_file(tmp_path):
+    """Tests if unknown file raises an Assertion error"""
+    with pytest.raises(AssertionError):
+        load_spectra(os.path.join(tmp_path, "file_that_does_not_exist.json"))
+
+
+@pytest.mark.parametrize("filename, ftype, expected_num_spectra", [
```
niekdejonge, line 13: Great that is a lot cleaner!

matchms/importing/load_spectra.py
```diff
@@ -0,0 +1,60 @@
...
+    -----
+    file:
+        Path to file containing spectra, with file extension "mzML", "json", "mgf", "msp",
+        "mzxml", "usi" or "pickle"
+    ftype:
+        Optional. Filetype
+    """
+    assert os.path.exists(file), f"The specified file: {file} does not exists"
+
+    if ftype is None:
```
niekdejonge, line 26: Good addition!

niekdejonge:
@florian-huber @hechth Thanks Helge, good suggestions! I will merge the pull request.

PR #398, opened 2023-04-17
Improve MSP output keys

### LINKED ISSUE #300: Improve MSP output keys
We have a function `save_to_msp` which works, in principle.

But it will simply take all keys and make them upper-case which is usually *very* different from the msp-style at MoNA or MassBank.
Related to #294 we might want to define a mapping dictionary with the most common key and how we want matchms to translate them from internal style (e.g. `precursor_mz`) to external/MSP style (e.g. `PrecursorMZ`).

zargham-ahmad:
This PR contains the following:

- added functions to translate internal style keys to external/MSP style keys #300 
- added export_key_conversions.csv 
- updated known_key_conversions.csv

Limitations: 

- As the export_key_conversions.csv have empty keys, therefore If the internal style key map to an empty key it will be removed
- while reading the file if the metadata contains a key which does not have matchms default key, it will be exported as it is to the output file. This case can be seen in the test_save_as_msp_export_style function.


matchms/exporting/save_as_msp.py
```diff
@@ -56,6 +59,9 @@ def save_as_msp(spectrums: List[Spectrum], filename: str,
                        filename.split(".")[-1])
     spectrums = _ensure_list(spectrums)
 
+    if style != "matchms":
+        list(map(lambda x: x.clone().update_keys(style), spectrums))
```
florian-huber:
```suggestion
        spectrums = [x.clone().update_keys(style) for x in spectrums]

```

florian-huber: Just wondering if that is maybe more readable.

florian-huber:
@zargham-ahmad @hechth Thanks for the PR!
In general, this looks very good.

One thing I wasn't sure about is the use of the `update_keys()` method.
To me, this makes sense if users also want to work (within matchms) with the other style keys.
If we just want to use the other style keys for export/import of files, we could maybe try to avoid changing the spectrum objects. For instance if we add a `to_dict()` method to `Metadata`, e.g.
```python
def to_dict(self, key_style="matchms"):
    if key_style == "matchms":
        return dict(self._data)
    else:
        ... # what's now in update_keys()
```

In `save_as_msp` we could then replace `for key, value in metadata.items()` with
`for key, value in metadata.to_dict("riken").items()`

... just an idea.

matchms/exporting/save_as_msp.py
hechth: linter actually complains about not using generator expressions I assume.

hechth:
@florian-huber your suggestion sounds good - but one thing we should maybe think about is also the handling of the different "styles" in general, not only for export. I was thinkin that actually something like a multi-dict would be interesting.

hechth:
For now I moved the key updating to the spectrum export part, this could theoretically also be used for mgs keys (but since I don't really work a lot with mgf maybe someone else can comment on that?).

Now, the keys which are not present in a specific format are not exported. The assumption is that other parsers which expect this format will fail if they encounter this MSP style, so better not include anything unspecified in the format.

florian-huber:
I will merge this to a local branch so that tests will run and that we can work on it together.

florian-huber:
Working on it in #401 .

PR #399, opened 2023-04-17
Handled sparse array_type fingerprint similarity

### LINKED ISSUE #389: Implement `sparse` array_type computation for fingerprint similarity
A naive implementation computing the dense matrix and then just returning it as sparse could be sufficient - or compute dense and then filter out 0 values.

zargham-ahmad:

- Handled sparse array_type fingerprint similarity #389 
- Updated fingerprint_similarity test

closes #389


hechth (approved):
Looks good - pretty straightforward. In a next iteration I think we can make an actual sparse implementation for the fingerprint similarity.

florian-huber (approved):
Thanks a lot for the changes. Looks good to me. I will first move it to a new local branch to run the CI.

PR #400, opened 2023-04-17
Improve MSP output keys

florian-huber:
PR from @zargham-ahmad:

This PR contains the following:

    added functions to translate internal style keys to external/MSP style keys 

    https://github.com/matchms/matchms/issues/300
    added export_key_conversions.csv
    updated known_key_conversions.csv

Limitations:

    As the export_key_conversions.csv have empty keys, therefore If the internal style key map to an empty key it will be removed
    while reading the file if the metadata contains a key which does not have matchms default key, it will be exported as it is to the output file. This case can be seen in the test_save_as_msp_export_style function.


hechth:
This is continued in #401

PR #401, opened 2023-04-17
switch to to_dict() method

florian-huber:
@hechth @zargham-ahmad
I added some changes. Let me know what you think of this.

I also checked the failing exporting_metadata test which happend because "accession" and "db#" are now both being mapped to "spectrum_id" according to your key_conversion table update (seems fine to, but I am not the best person to judge this).

If the changes look OK, the `update_keys()` method is in principle no longer needed.


tests/test_exporting_metadata.py
```diff
@@ -30,8 +30,8 @@ def test_get_metadata_dict():
 def test_get_metadata_as_array(spectra):
     actual, colnames = get_metadata_as_array(spectra)
     assert len(actual) == 10
-    assert len(colnames) == 24
-    
+    assert len(colnames) == 23
+
```
hechth, line 34: This is because the two keys are mapped to the same key in matchms now?

matchms/exporting/save_as_msp.py
hechth, line 1: Yeah this is actually way more elegant now.

matchms/Spectrum.py
```diff
@@ -182,7 +182,10 @@ def to_dict(self) -> dict:
...
             spectrum_dict["fingerprint"] = spectrum_dict["fingerprint"].tolist()
         return spectrum_dict
-    
+
+    def metadata_dict(self, export_style: str = "matchms") -> dict:
+        return self._metadata.to_dict(export_style)
+
     def update_keys(self, key_conversions: dict):
         self._metadata.update_keys(key_conversions)
         return self
```
hechth: So I think this is now obsolete.

hechth (approved):
Looks good to me! Thanks for polishing this.

tests/test_exporting_metadata.py
florian-huber: Yes, "accession" and "db#" are now mapped to "spectrum_id". I guess that was intended?

hechth:
@florian-huber this looks good to me and I think it can be merged.

matchms/exporting/save_as_msp.py
```diff
@@ -61,17 +59,16 @@ def save_as_msp(spectrums: List[Spectrum], filename: str,
...
-            _write_spectrum(spectrum, outfile, write_peak_comments)
+            _write_spectrum(spectrum, outfile, write_peak_comments, style)
 
 
-def _write_spectrum(spectrum: Spectrum, outfile: IO, write_peak_comments: bool):
-    _write_metadata(spectrum.metadata, outfile)
+def _write_spectrum(spectrum: Spectrum,
+                    outfile: IO,
+                    write_peak_comments: bool,
+                    export_style: str = "matchms"):
```
zargham-ahmad, line 70: Do we really need to have a default value for export_style in every function if we will not be calling the function individually?

matchms/importing/load_spectra.py
```diff
@@ -0,0 +1,60 @@
...
+        "mzxml", "usi" or "pickle"
+    ftype:
+        Optional. Filetype
+    """
+    assert os.path.exists(file), f"The specified file: {file} does not exists"
+
+    if ftype is None:
+        ftype = os.path.splitext(file)[1].lower()[1:]
+
+    if ftype == "mzml":
```
zargham-ahmad, line 31: I think it would be good if we lower() the ftype before the comparison if this function will be used by the user

zargham-ahmad (approved):
Looks good to me!

matchms/importing/load_spectra.py
```diff
@@ -0,0 +1,60 @@
...
+    file:
+        Path to file containing spectra, with file extension "mzML", "json", "mgf", "msp",
+        "mzxml", "usi" or "pickle"
+    ftype:
+        Optional. Filetype
+    """
+    assert os.path.exists(file), f"The specified file: {file} does not exists"
+
+    if ftype is None:
+        ftype = os.path.splitext(file)[1].lower()[1:]
```
hechth, line 27:
```suggestion
    if ftype is None:
        ftype = os.path.splitext(file)[1].lower()[1:]
    else:
        ftype = ftype.lower()
```

matchms/exporting/save_as_msp.py
hechth: it doesn't hurt I'd say

PR #402, opened 2023-04-17
Consider move to Python 3.8 and higher

florian-huber:
- CI runs for Python 3.8, 3.9, 3.10
- since `rdkit` is now on pypi, tests for full functionality including rdkit now run faster when using pip instead of conda.

Remaining open question: drop Python 3.7 ? yes/no?

Fixes #395

PR #403, opened 2023-04-18
Try mamba for installing rdkit

florian-huber:


florian-huber:
went for #402 instead.

PR #404, opened 2023-04-19
Issue300

florian-huber:
Code was not yet merged into master branch.

@zargham-ahmad thanks for contributing!
Please feel free to add your name to `CITATION.cff` and `zenodo.json`.


matchms/Metadata.py
```diff
@@ -144,6 +144,26 @@ def items(self):
...
+        Parameters
+        ----------
+        export_style:
+            Specifies the naming style of the metadata fields.
+            Default is "matchms".
+        """
+        if export_style == "matchms":
+            return dict(self._data)
+        key_conversions = load_export_key_conversions(export_style=export_style)
+        keep_list = [x for x in self._data.keys() if x in key_conversions]
```
hechth, line 159: This should actually have been checking if it is in the values, not the keys, as the keys are actually the keywords from matchms, not the style to export to

PR #405, opened 2023-04-20
Handled sparse array_type fingerprint similarity

florian-huber:
Local branch containing #399 (to trigger CI).


florian-huber:
Thanks a lot @hechth and @zargham-ahmad !

PR #409, opened 2023-04-27
Updated CI to trigger on all pushes onto a PR branch

hechth:


florian-huber:
@hechth I wasn't sure what the status is (due to failing tests). Let me know if you want any code reviews.

hechth:
This will actually pass and only makes changes to the CI

PR #410, opened 2023-05-10
Updated load_from_mgf reader

### LINKED ISSUE #407: save_as_msp fails if "write_peak_comments" is true and peak comments are empty
**To Reproduce**
1. Load spectra without peak comments
2. Save spextra to msp with "write_peak_comments" as true
3. Observe bug

**Expected behavior**
A warning should be written to the logfile and the program should just not write the peak comments.

zargham-ahmad:

- updated load_from_mgf reader to convert peak_comments from string to dict. The issue was that the peaks comments returned by the load_from_mgf were in string and save_as_msp expects them to be in dict #407 
- added test and testdata 
- updated condition to check if peak_comments are dict in save_as_msp 
- updated changelog

closes #407


hechth (approved):

florian-huber (approved):
Nice, thanks for this update!

PR #418, opened 2023-05-19
Updated harmonize_values function to remove invalid metadata

### LINKED ISSUE #344: Empty `inchi` doesn't get removed but results in empty string -> remove invalid metadata after reading the spectrum
**Describe the bug**
Currently, an `inchi` entry that is empty will result in a metadata entry that holds an empty string.
Ideally, if spectra are loaded with `harmonize_metadata` as true, _invalid_ or empty metadata values should be removed from the spectrum metadata.

zargham-ahmad:

- Updated harmonize_values function to remove invalid metadata #344 
- added test
- updated changelog

closes #344


florian-huber (approved):
Looks like this does the job! Thanks.
(we might want to have something like this in PickyDict, too, but given how small the change is I would leave it as it is here for now)

hechth (approved):
Looks good - yeah we can at some point move this into pickydict but for now I think it is fine to be here

PR #419, opened 2023-05-22
add min_mz and max_mz

florian-huber:
add parameters `min_mz` and `max_mz` (such that they can also be used within the array plots).


florian-huber:
I consider these very minor changes, so I will merge this right away.

PR #426, opened 2023-05-30
Loading spectra in pipeline

### LINKED ISSUE #425: Pipeline should use load_spectra function
A new function was added to load in spectra from any filetype (based on the file extension). 
Pipeline has a separate approach for doing the same. For maintainability and allowing the use of pickled spectra the load spectra should be reused.

niekdejonge:
Fixes #425


hechth (approved):
Looks good

PR #427, opened 2023-06-01
added get compound from puchem

niekdejonge:


niekdejonge:
@florian-huber I refactored the pubchem lookup and also added tests. The approach is different from matchmsextras, since it first collects all unique compound names, which drastically reduces the number of lookups that have to happen. Now finishes in < 1 hour.
I have added tests, do you think this fits in matchms? Or should it stay in matchmsextras? 
It is different in approach, since it runs on a list of spectra, instead of per spectrum. This would be an argument for leaving it in matchmsextras. In that case we could update matchms extras with this version.

niekdejonge:
@florian-huber I also added the pytest.importorskip("rdkit") to one of the tests that uses rdkit. Does this test still run in the current workflow or is it skipped in all tests?

hechth:
@niekdejonge please have a look at this package: https://github.com/RECETOX/MSMetaEnhancer It basically already does all of this and more.

niekdejonge:
@hechth Thanks! That looks really interesting indeed! And it already has some compatibility with matchms. Might be interesting to make part of matchms pipeline at some point as well.

hechth:
the tool is built on top of matchms and is already available on bioconda and Galaxy - it is already fully compatible with the pipeline :)

niekdejonge:
@hechth I noticed, that is great! I meant specifically the Pipeline class in matchms, but that might also already work :)

hechth:
@niekdejonge and @florian-huber I'd vote for keeping this in matchms extras and for now out of the core package, as it is a bit out of scope IMHO. Also, the functionality is already provided by MSMetaEnhancer. As that one depends on matchms, it isn't easy to combine it with the pipeline, but as it is a python package, people can easily use it in their own notebooks etc.

Sorry to be a bit of a showstopper here but I just think it is a lot cleaner to keep this separated.

niekdejonge:
@hechth No I agree! I also was not sure if we wanted to add it here already. I will make a PR in matchmsextras for this.

hechth:
Then let's keep this open here for now and we can close it once the PR is merged in matchmsextras. Maybe just xref the PR here so we can keep an eye on it :)

niekdejonge:
Has been merged in matchmsextras https://github.com/matchms/matchmsextras/pull/18

PR #436, opened 2023-06-06
New filter functions for parent mass

niekdejonge:
Parent mass does not always match smiles match in library spectra from GNPS.
To make sure the annotations are correct, we will add filter functions that can remove incorrect matches.

In addition functions will be added that can repair the parent mass, to make sure not too many (almost correct) spectra are removed.


niekdejonge:
Some stats about the filtering of the GNPS library:
Running with mass tolerance = 0.4
require_parent_mass_match_smiles without first repairing removes 24879 spectra.

repair_adduct_based_on_smiles repairs: 8175/24879 spectra
repair_precursor_is_parent_mass repairs: 964/16704 
repair_smiles_salt_ions: 783/1496

After repairs:
14957 are removed by require_parent_mass_match_smiles

hechth:
@niekdejonge could you elaborate a bit what you mean with the parent mass match not matching the smiles match? Do you mean that the mass derived from SMILES doesn't match with the parent mass given in the spectrum metadata?

niekdejonge:
@hechth Yes exactly. The functions I am adding in this PR are meant to solve some of these issues that can be relatively easily resolved. For instance if the wrong adduct is given.

PR #439, opened 2023-06-15
Updated regex in get_peak_values function

### LINKED ISSUE #421: Parsing issues with matchms 0.19
Hello !

There is some issus about "bad peaks pattern" and "could not convert string to float (for metadata_filtering.get("parent"))" with the 0.19 which worked with the 0.18

zargham-ahmad:

- Updated the regex in the get_peak_values function to fix the parsing issue for scientific notation as a float #421 
- added test
- updated changelog

closes #421


hechth (commented):
please add a test with the spectrum from the issue, after that I think this can be merged!

zargham-ahmad:
> please add a test with the spectrum from the issue, after that I think this can be merged!

done in 150daef

hechth (approved):

PR #447, opened 2023-06-23
rdkit log level implementation to supress warnings

alpi314:
I had a problem that when using matchms metadata filter functions, lower level rdkit logs would be shown.

Examples:
`[11:57:51] WARNING: Omitted undefined stereo`
`[12:31:27] WARNING: Charges were rearranged; Omitted undefined stereo`

To solve this I have implemented a function on matchms level that mimics [rdkit's set log level](https://github.com/rdkit/rdkit/blob/master/rdkit/RDLogger.py). The functionality of this is a bit above just warning supression since it allows you to set any rdkit log level you want. Avaliable log levels are read from rdkit so that they remain consistent.

I have also added a test in tests/test_metadata_utils.py that checks if all log messages are propagated correctly.

### DIFF
--- a/matchms/__init__.py
+++ b/matchms/__init__.py
@@ -11,7 +11,7 @@ from .Spectrum import Spectrum
 
 
 _init_logger()
-set_rdkit_logger_level(logging_functions.RDKIT_ERROR)
+set_rdkit_logger_level('rdApp.error')
 
 
 __author__ = "Netherlands eScience Center"
--- a/matchms/__init__.py
+++ b/matchms/__init__.py
@@ -11,7 +11,22 @@ from .Spectrum import Spectrum
 
 
 _init_logger()
-set_rdkit_logger_level('rdApp.error')
+
+try:  # rdkit is not included in pip package
+    from rdkit import Chem
+    set_rdkit_logger_level('rdApp.error')
+except ImportError:
+    _has_rdkit = False
+    from collections import UserString
+
+    class ChemMock(UserString):
+        def __call__(self, *args, **kwargs):
+            return self
+
+        def __getattr__(self, key):
+            return self
+
+    Chem = AllChem = ChemMock("")
 
 
 __author__ = "Netherlands eScience Center"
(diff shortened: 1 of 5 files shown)


alpi314:
I have just realized that there is a [matchms/logging_functions.py](https://github.com/matchms/matchms/blob/master/matchms/logging_functions.py). It might make more sense for the function to be in that file.

niekdejonge:
@al-pi314 Thanks a lot! Looks good.
I think we can go one step further and by default already set the warnings to rdApp.error, since these warnings about omitted stereo are generally not useful and clutter the logging. @florian-huber do you agree?

florian-huber:
Yes @niekdejonge, silencing them by default would be fine for me as well.

hechth:
Would it be also possible to tie the logging level of RDkit to the one of matchms? So that setting loglevel to warning will also set the one of RDkit to warning?

alpi314:
I was also thinking of connecting the log levels, but I believe there would be two issues:

- matchms logging can initalize multiple independednt loggers (main one is named "matchms") with different logging levels, which could lead to ambiguity to which logger the rdkit's level (which is global) is tied to.
- matchms uses warning level as default, but to suppress (not so important) rdkit's warnings the default for rdkit log level would need to be "rdkit.error"

hechth:
> I was also thinking of connecting the log levels, but I believe there would be two issues:
> 
>     * matchms logging can initalize multiple independednt loggers (main one is named "matchms") with different logging levels, which could lead to ambiguity to which logger the rdkit's level (which is global) is tied to.
> 
>     * matchms uses warning level as default, but to suppress (not so important) rdkit's warnings the default for rdkit log level would need to be "rdkit.error"

Fair enough, these are actually very good points!

niekdejonge:
@al-pi314 Thanks for the Changes! I made some small changes to make all test pass and will merge the PR once they finish.

florian-huber:
Thanks for the contribution @al-pi314 !

Please also feel free to add your name to:
- https://github.com/matchms/matchms/blob/master/.zenodo.json
- https://github.com/matchms/matchms/blob/master/CITATION.cff

alpi314:
@florian-huber Thanks, I would be grateful. Since the pull request is merged and I would probably need to open a new one could I ask you, when able, to make the change directly?

First name: Aleksander
Last name: Piciga
Affiliation (for the related project): Jožef Stefan Institute

niekdejonge:
@al-pi314 I have just added you to the citation files in #468 Thanks again for your contribution!

PR #440, opened 2023-06-16
Interpret adducts

niekdejonge:
Added functionality to interpret adducts if they are not in known adducts list.

PR #448, opened 2023-06-27
Add smiles from compound name

niekdejonge:
Added a function to add annotations based on compound names. 
The functionality to get the annotations from pubchem can be found in matchmsextras pubchem_lookup. 
This function just uses a csv file with annotated compound names to add annotations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New filter functions to repair a smiles that mismatches parent mass
-- Added repair_smiles_from_compound_name
+- Added repair_smiles_from_compound_name [#448](https://github.com/matchms/matchms/pull/448)
 ### Changed
 - Moved functions related to adducts and parent mass
 - Updated reiterate_peak_comments function to convert the peak_comments keys to float [#437](https://github.com/matchms/matchms/pull/437)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New filter functions to repair a smiles that mismatches parent mass
-
+- Added repair_smiles_from_compound_name
 ### Changed
 - Moved functions related to adducts and parent mass
 - Updated reiterate_peak_comments function to convert the peak_comments keys to float [#437](https://github.com/matchms/matchms/pull/437)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New filter functions to repair a smiles that mismatches parent mass
-
+- Added repair_smiles_from_compound_name
```
florian-huber:
```suggestion
- Added repair_smiles_from_compound_name [#448](https://github.com/matchms/matchms/pull/448)
```

matchms/filtering/repair_smiles_from_compound_name.py
```diff
@@ -0,0 +1,72 @@
...
+logger = logging.getLogger("matchms")
+
+
+def repair_smiles_from_compound_name(spectrum_in: Spectrum,
+                                     annotated_compound_names_file,
+                                     mass_tolerance=0.1):
+    """Adds annotations (smiles, inchi, inchikey) based on compound name
+
+    The compound name smiles matches should already be stored in a csv file.
+    This csv file can be made using pubchem_lookup.py from matchmextras
```
florian-huber:
```suggestion
    Based on a table of compound names and smiles matches (stored in a csv file) this function
    adds the new annotations to the input spectrums if the smiles seem consistent with the available
    spectrum metadata (e.g., parent mass).
    This function can be used with csv files that are returned by the pubchem_lookup.py from matchmextras.
```

florian-huber (approved):
Looks good to me. The main thing to discuss (@hechth and @niekdejonge ) is if we want to add pandas to the dependencies. I would be up for it, since we can use it in many other places, too.

niekdejonge:
> Looks good to me. The main thing to discuss (@hechth and @niekdejonge ) is if we want to add pandas to the dependencies. I would be up for it, since we can use it in many other places, too.

I think it will not cause any problems, since most of the scientific community will already use pandas. 
The only problem would be if we restrict the specific pandas version too much. It might be worth adding some tests for different pandas versions. (at least 1.0 and 2.0)
I now also use it for working with the adduct table in #440

hechth:
The function name is a bit very specific and this could be generalized as a metadata import. I think for now it is okay to keep it in this specialized way, but how it could work in the future is to just have one function to import metadata which will then be joined based on a specific key and to have a filter which checks for the integrity of metadata.

hechth:
About introducing pandas - I think it makes sense if we consider rewriting the metadata and spectral library handling. If we introduce a spectral library class, I think it makes sense to keep all metadata in a pandas dataframe. This spectral library implementation could also be based on mzSpecLib to make it more modular and make it work with the open standard.

PR #449, opened 2023-06-27
Add filter ionmode

niekdejonge:
Added a filter function to remove the wrong ionmode


matchms/filtering/require_correct_ionmode.py
```diff
@@ -0,0 +1,27 @@
...
+
+def require_correct_ionmode(spectrum_in: Spectrum,
+                            ion_mode_to_keep):
+    """Removes spectra that are not in the correct ionmode"""
+    if spectrum_in is None:
+        return None
+    spectrum = spectrum_in.clone()
+
+    assert ion_mode_to_keep in {"positive", "negative", "both"}, \
+        "ion_mode should be set to 'positive', 'negative' or 'both'"
```
hechth: assert is generally not advised to use in normal code I think - the recommended thing is to throw an exception. (https://stackoverflow.com/questions/944592/best-practice-for-using-assert) asserts might also get removed for optimization, so they are not a very stable way of handling errors

hechth: We could also start having a collection of constants as enums or so for these categorical variables.

tests/filtering/test_require_corect_ionmode.py
hechth, line 1: There is a typo in the filename!

tests/filtering/test_require_corect_ionmode.py
```diff
@@ -0,0 +1,23 @@
...
+    ("positive", "positive", False),
+    ("negative", "negative", False),
+    ("positive", "both", False),
+    ("negative", "both", False),
+    ("positive", "negative", True),
+    ("negative", "positive", True),
+    ("n/a", "both", True),
+])
+def test_clean_compound_name(ionmode, ionmode_to_keep, spectrum_removed):
+    builder = SpectrumBuilder()
```
hechth: I think this test should be renamed as it doesn't test cleaning the compound name?

hechth (changes requested):
Please see my comments about the test names and filenames, otherwise seems good.

tests/filtering/test_require_corect_ionmode.py
niekdejonge: Well spotted :)

tests/filtering/test_require_corect_ionmode.py
niekdejonge: Yes! Copied an old test as scaffold...

matchms/filtering/require_correct_ionmode.py
niekdejonge: The collection of constants might be good to add at some point indeed. I will leave it like this for now.

PR #451, opened 2023-06-28
Require annotation

niekdejonge:
@florian-huber In addition to checking for valid_inchi, valid_inchikey, valid_smiles I also check if both the smiles and inchi match the inchikey. On the GNPS set this removes maybe a few hundred structures. I don't plan to write a repair function, since these are so little spectra and it will be hard to determine which one of the two is correct.


matchms/filtering/require_annotation.py
```diff
@@ -0,0 +1,39 @@
...
+                       "inchi = %s, inchikey = %s, expected_inchikey = %s",
+                       inchi, inchikey, convert_inchi_to_inchikey(inchi))
+        return None
+    # check if smiles matches the inchikey (first convert to inchi followed by converting to inchikey)
+    if not inchikey[:14] == convert_inchi_to_inchikey(convert_smiles_to_inchi(smiles))[:14]:
+        logger.warning("Removed spectrum since smiles does not match the inchikey. "
+                       "inchikey = %s, smiles = %s, expected_inchikey = %s",
+                       inchikey, smiles, convert_inchi_to_inchikey(convert_smiles_to_inchi(smiles)))
+        return None
+    return spectrum
```
hechth: this could be extracted to a private function to check inchikey as it is a bit more complex.

matchms/filtering/require_annotation.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def require_annotation(spectrum_in: Spectrum):
+    """Removes spectra that are not fully annotated (correct and matching, smiles, inchi and inchikey)"""
+    if spectrum_in is None:
+        return None
+    spectrum = spectrum_in.clone()
+    smiles = spectrum.get("smiles")
+    inchi = spectrum.get("inchi")
+    inchikey = spectrum.get("inchikey")
```
hechth: I think we don't need to clone here as the spectrum variable is never used except when returning.

hechth (commented):
LGTM - some minor comments on functions. Would be good to add this addition of the filter to the changelog.

florian-huber (approved):
Nice, saves me to use my custom code snippet each time I do this in a notebook :)

Thanks @hechth and @niekdejonge !

PR #454, opened 2023-07-02
refactor filter functions and their path structure

florian-huber:
- Divide filter functions into three folders: `filter_utils`, `metadata_processing`, and `peak_processing`.
- Removal of the filters `make_ionmode_lowercase` and `set_ionmode_na_when_missing`
(their functionality should meanwhile be completely be covered by our `Metadata` class.
- Adjust imports and __init__.py files accordingly
- Some code lintings
- Add missing docstrings.


tests/filtering/test_derive_adduct_from_name.py
```diff
@@ -1,7 +1,8 @@
 import pytest
```
niekdejonge, line 1: The tests do not yet follow the same structure as the filter functions. For consistency this might be good to change as well.

niekdejonge:
Looks good!

matchms/filtering/SpeciesString.py
```diff
@@ -2,8 +2,32 @@
 
 
 class SpeciesString:
```
hechth, line 4: I find this class name rather non-intuitive, maybe we can use something like StructureRepresentation or so?

hechth (approved):

Issue #432, opened 2023-06-05
Group filtering functions


niekdejonge:
The matchms filtering functions are now all in one folder, which makes it hard to select the filtering functions needed for your issue.

Suggestion:
Group the filter functions by functionality. E.g.:
- Peak filtering
  - Removing peaks
  - Checking peak quality
- Metadata cleaning
  - Harmonizing metadata names
  - Adding missing metadata
  - Annotation cleaning (smiles, inchi key, inchi)

__init__.py in filtering should still contain all the links to the filter functions, to make it flexible to change the sub(folder) in which these functions are stored.

florian-huber:
This underwent a few changes already with the latest PRs (#440, # 448), but we should revise this more carefully. Right now this produces issues with the Conda CI --> https://github.com/bioconda/bioconda-recipes/pull/41784#issuecomment-1615769099

I suggest the following:
- Add a `SpectrumProcessing` class that will take care of the different cleaning, filtering etc. steps in a more user-friendly way.
- In the back of this we will keep all the different filtering functions we have, but using them directly will be meant for developers mostly.
- We can also add a bit more structure (though not over-engineer this), e.g., a folder for `utils`, a folder for peak related actions (`peak_processing`) and a folder for metadata related functions (`metadata_processing`).

What do you think @hechth @niekdejonge ?

florian-huber:
--> working on it in #454

florian-huber:
--> and #455

niekdejonge:
I agree that the effort needed for converting all functions to classes might not be worth it. And a SpectrumProcessing class might indeed be a lot easier to implement. 

How do you plan to let SpectrumProcessing differ from Pipeline? Would you like to separate the cleaning and the scores calculations into SpectrumProcessing and Pipeline? Or would SpectrumProcessing be part of Pipeline?

I have a few things that might reconsidering the Filter classes worth it would be: 
- How would you store which functions should be run before the other? With the Filter classes, I would like to add a method that returns required_filter_steps. This method can be called by Pipeline to check if the previous steps have already been performed/or to reorder the filter functions. For the SpectrumProcessing you might store the required order in SpectrumProcessing itself as a list of rules, but this might become confusing quickly. 
[...]

PR #455, opened 2023-07-03
Add filter class

florian-huber:
Adding a new `SpectrumProcessing` class. Would be used like this:

```python
processor = SpectrumProcessor("default")
spectrum_processed = processor.process_spectrum(spectrum)
```

Or, to build something custom-like
```python
processor = SpectrumProcessor("basic")
processor.add_matchms_filter((
    "require_minimum_number_of_peaks",
    {"n_required":  5}
    ))
processor.add_filter("require_valid_annotation")
spectrum_processed = processor.process_spectrum(spectrum)
```

Some open ends:
- [x] add `process_spectrums()` method (with progress bar)
- [x] add processing report --> what was changed etc.
- [x] Integrate with `Pipeline` class
- [x] Allow to add custom-made filter functions, too.

In addition, this PR removes a long obsolete filter (`make_charge_scalar()`).


matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,287 @@
...
+        Returns
+        -------
+        Spectrum
+            The processed spectrum.
+        """
+        if self.filters == []:
+            raise TypeError("No filters to process")
+        for filter_func in self.filters:
+            spectrum = filter_func(spectrum)
+            if spectrum is None:
```
niekdejonge: I would like to keep track of the number of changes each filter step did, instead of what fields were changed, to get a bit more insight into what is happening. This can be implemented by calling processing_report.add_to_report() here. 

Add_to_report does not have to find exactly what was changed in this case. I would instead just return, was changed, or was not changed. Alternatively, we could extend this to spectrum was removed, a field was added, or was changed, but this is often already clear from the function name. 

This can be implemented without changing the filtering functions to classes, but it might be causing some extra overhead since all fields have to be checked again after each filter function. A class approach would make this faster and cleaner, but I agree that this is probably not worth the time investment by us, so this is a nice in-between solution.

niekdejonge: I will make a separate PR into this branch to illustrate what I had in mind.

matchms/Pipeline.py
```diff
@@ -95,18 +93,40 @@ def _initialize_workflow_dict(self, config_file):
             self.workflow = OrderedDict()
             self.workflow["importing"] = {"queries": None,
                                           "references": None}
-            self.workflow["filtering_queries"] = ["default_filters"]
-            self.workflow["filtering_refs"] = ["default_filters"]
+            self.workflow["predefined_processing_queries"] = "default"
```
niekdejonge, line 114: I did not directly get that refs stood for reference. So maybe change that to additional_processing_references

matchms/Pipeline.py
```diff
@@ -43,8 +41,8 @@ class Pipeline:
 
         pipeline = Pipeline()
         pipeline.query_files = "spectrums_file.msp"
-        pipeline.filter_steps_queries = [
-            ["default_filters"],
+        pipeline.predefined_processing_queries = "basic"
+        pipeline.additional_processing_queries = [
             ["add_parent_mass"],
```
niekdejonge, line 46: The Pipeline class became quite complex by now. It did cost me some time to familiarize myself again after my holidays. This is not really due to this PR. But still might be worth making some changes to make it more modular. Let me know what you think about the suggested changes below:

An option that might make sense would be splitting pipeline into two parts. 1 which will just take in a yaml file and run this and a separate function that can be used to create new Yaml files. 

I think currently Yaml files also contain the spectrum files location. To me, this does not really make sense, since you would like to share Yaml files with others. I would argue using a Yaml file only for the pipeline and specifying the query files when running pipeline. Your logfile on the other hand should store the filenames.

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,316 @@
...
+from functools import partial
+from typing import Optional
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+import matchms.filtering as msfilters
+from matchms import Spectrum
+
+
+class SpectrumProcessor:
```
niekdejonge, line 11: Looks good to me! I will later try running the new pipeline on some larger test cases as well.

niekdejonge (approved):
To me it looks good and can be merged. When going over this PR I noticed that the Pipeline class became quite bulky and confusing to me, so a left a few comments about this, but this should (if we want to) be changed in a future PR.

matchms/Pipeline.py
florian-huber: Thanks. I will change this.

matchms/Pipeline.py
florian-huber: That's a good remark and I am not sure what I would prefer here.
But about file location and what to include into Pipeline (or the yaml) file and what not, I would prefer making a separate issue/PR to not make this one even more complex...

I did a bit of code cleaning to the Pipeline class, but it obviously remains a complex piece of code nonetheless.

niekdejonge: Yes fully agree that this should not be added to this PR. I will make an issue. I am also not sure if that would really make things better, so I will think about it a bit more.

matchms/Pipeline.py
```diff
@@ -86,32 +84,51 @@ def __init__(self, config_file=None, progress_bar=True):
...
+        """Define Pipeline workflow based on a yaml file (config_file).
+        """
         if config_file is None:
             self.workflow = OrderedDict()
             self.workflow["importing"] = {"queries": None,
                                           "references": None}
-            self.workflow["filtering_queries"] = ["default_filters"]
-            self.workflow["filtering_refs"] = ["default_filters"]
+            self.workflow["predefined_processing_queries"] = "default"
+            self.workflow["predefined_processing_references"] = "default"
```
niekdejonge, line 115: @florian-huber I just realized that the yaml file will now store "default" instead of all the filter steps in default. I would remove the "predefined_processing_queries" in the yaml file and just combine the "predefined_processing_queries" and the addition_processing_queries, to make sure that changes in what we define under default will not affect the results of an analysis, when redone.

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,316 @@
...
+class ProcessingReport:
+    """Class to keep track of spectrum changes during filtering.
+    """
+    def __init__(self):
+        self.counter_changed_field = defaultdict(int)
+        self.counter_added_field = defaultdict(int)
+        self.counter_removed_spectrums = defaultdict(int)
+        self.counter_number_processed = 0
+
+    def add_to_report(self, spectrum_old, spectrum_new: Spectrum,
```
niekdejonge, line 278: @florian-huber Filter steps that do not do any filtering do not end up in the report currently. This might be wanted behaviour, since it keeps the list short, however, you could also argue that it might be nice to have a complete overview of all the filter steps, to easily pick out the filter steps that do not have any effect.

matchms/Pipeline.py
```diff
@@ -132,25 +149,22 @@ def run(self):
...
-                spectrum = self.apply_filter(spectrum, step)
-            self.spectrums_queries[i] = spectrum
-        self.spectrums_queries = [s for s in self.spectrums_queries if s is not None]
+        # Process query spectra
+        spectrums, report = self.processing_queries.process_spectrums(
+            self.spectrums_queries,
+            create_report=True,
+            progress_bar=self.progress_bar)
+        self.spectrums_queries = spectrums
+        self.write_to_logfile(str(report))
```
hechth, line 158: I think the spectrum processor should handle the logging and should eventually be passed a logger or so, which can then come from the pipeline?

matchms/Pipeline.py
```diff
@@ -346,20 +344,40 @@ def logging_level(self, log_level):
...
+    def predefined_processing_references(self):
+        return self.workflow.get("predefined_processing_references")
 
-    @filter_steps_queries.setter
-    def filter_steps_queries(self, files):
-        self.workflow["filtering_queries"] = files
+    @predefined_processing_references.setter
+    def predefined_processing_references(self, files):
+        self.workflow["predefined_processing_references"] = files
+        self._initialize_spectrum_processor_references()
```
hechth, line 362: This just seems very bulky. Is there a use case where we want different filters for the references or for the queries? I think in any case, we would only want a single processor and then run it with different filters.

matchms/Pipeline.py
```diff
@@ -86,32 +84,51 @@ def __init__(self, config_file=None, progress_bar=True):
...
+        """Initialize spectrum processing workflow for the reference spectra."""
+        self.processing_references = initialize_spectrum_processor(
+            self.predefined_processing_references,
+            self.additional_processing_references
+            )
+        self.write_to_logfile(str(self.processing_references))
+
+    def import_workflow_from_yaml(self, config_file):
+        """Define Pipeline workflow based on a yaml file (config_file).
+        """
```
hechth, line 109: Missing docstring and type hints

matchms/Pipeline.py
```diff
@@ -346,20 +344,40 @@ def logging_level(self, log_level):
...
     @property
-    def filter_steps_queries(self):
-        return self.workflow.get("filtering_queries")
+    def predefined_processing_queries(self):
+        return self.workflow.get("predefined_processing_queries")
+
+    @predefined_processing_queries.setter
+    def predefined_processing_queries(self, files):
+        self.workflow["predefined_processing_queries"] = files
+        self._initialize_spectrum_processor_queries()
```
hechth, line 353: missing docstring and type hints

matchms/Pipeline.py
```diff
@@ -346,20 +344,40 @@ def logging_level(self, log_level):
...
+
+    @property
+    def predefined_processing_references(self):
+        return self.workflow.get("predefined_processing_references")
```
hechth, line 357: missing return type annotation

matchms/Pipeline.py
hechth, line 362: missing docstring and type hint

matchms/Pipeline.py
```diff
@@ -346,20 +344,40 @@ def logging_level(self, log_level):
...
+    def additional_processing_queries(self):
+        return self.workflow.get("additional_processing_queries")
 
-    @filter_steps_refs.setter
-    def filter_steps_refs(self, filter_list):
-        self.workflow["filtering_refs"] = filter_list
+    @additional_processing_queries.setter
+    def additional_processing_queries(self, files):
+        self.workflow["additional_processing_queries"] = files
+        self._initialize_spectrum_processor_queries()
```
hechth, line 371: missing docstring and type hint

matchms/Pipeline.py
```diff
@@ -346,20 +344,40 @@ def logging_level(self, log_level):
...
+
+    @property
+    def additional_processing_references(self):
+        return self.workflow.get("additional_processing_references")
```
hechth, line 375: missing return type annotation

matchms/Pipeline.py
```diff
@@ -346,20 +344,40 @@ def logging_level(self, log_level):
...
+
+    @additional_processing_references.setter
+    def additional_processing_references(self, files):
+        self.workflow["additional_processing_references"] = files
+        self._initialize_spectrum_processor_references()
```
hechth, line 380: missing docstring and type hint

matchms/Pipeline.py
```diff
@@ -395,3 +413,13 @@ def _dict_representer(dumper, data):
             data.items())
     OrderedDumper.add_representer(OrderedDict, _dict_representer)
     return yaml.dump(data, stream, OrderedDumper, **kwds)
+
+
+def initialize_spectrum_processor(predefined_workflow, additional_filters):
+    """Initialize spectrum processing workflow."""
+    processor = SpectrumProcessor(predefined_workflow)
```
hechth, line 420: missing docstrings, type hint and return type annotation

matchms/Pipeline.py
```diff
@@ -395,3 +413,13 @@ def _dict_representer(dumper, data):
...
+    for step in additional_filters:
+        if isinstance(step, (tuple, list)) and len(step) == 1:
+            step = step[0]
```
hechth, line 423: I think this defensive programming makes the code more obscure - we should just agree on expecting a list and with the type hints, users should be made aware that they need to pass a list, even if it has only a single entry

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+                raise ValueError("Predefined pipeline parameter should be a string")
+            if predefined_pipeline not in PREDEFINED_PIPELINES:
+                raise ValueError(f"Unknown processing pipeline '{predefined_pipeline}'. Available pipelines: {list(PREDEFINED_PIPELINES.keys())}")
+            for fname in PREDEFINED_PIPELINES[predefined_pipeline]:
+                self.add_matchms_filter(fname)
+
+    def add_filter(self, filter_function):
+        """Add a filter to the processing pipeline. Takes both matchms filter names (and parameters)
+        as well as custom-made functions.
+        """
```
hechth, line 39: missing type hints and argument documentation

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+        if isinstance(filter_function, str):
+            self.add_matchms_filter(filter_function)
+        elif isinstance(filter_function, (tuple, list)) and isinstance(filter_function[0], str):
+            self.add_matchms_filter(filter_function)
+        else:
+            self.add_custom_filter(filter_function[0], filter_function[1])
+
+    def add_matchms_filter(self, filter_spec):
```
hechth, line 47: missing type hint

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+            filter_func = partial(FILTER_FUNCTIONS[filter_name], **filter_args)
+            filter_func.__name__ = FILTER_FUNCTIONS[filter_name].__name__
+        else:
+            raise TypeError("filter_spec should be a string or a tuple or list")
+
+        self.filters.append(filter_func)
+        # Sort filters according to their order in self.filter_order
+        self.filters.sort(key=lambda f: self.filter_order.index(f.__name__))
+
+    def add_custom_filter(self, filter_function, filter_params=None):
```
hechth, line 74: missing type hints

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+        self.filter_order.insert(filter_position + 1, filter_function.__name__)
+        if filter_params is None:
+            self.filters.append(filter_function)
+        else:
+            filter_func = partial(filter_function, **filter_params)
+            filter_func.__name__ = filter_function.__name__
+            self.filters.append(filter_func)
+
+    def process_spectrum(self, spectrum,
+                         processing_report: Optional["ProcessingReport"] = None):
```
hechth, line 104: missing return type and type hints

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+    def __init__(self, predefined_pipeline='default'):
+        self.filters = []
+        self.filter_order = [x.__name__ for x in ALL_FILTERS]
+        if predefined_pipeline is not None:
+            if not isinstance(predefined_pipeline, str):
+                raise ValueError("Predefined pipeline parameter should be a string")
+            if predefined_pipeline not in PREDEFINED_PIPELINES:
+                raise ValueError(f"Unknown processing pipeline '{predefined_pipeline}'. Available pipelines: {list(PREDEFINED_PIPELINES.keys())}")
+            for fname in PREDEFINED_PIPELINES[predefined_pipeline]:
+                self.add_matchms_filter(fname)
```
hechth, line 34: I think the filters should be passed into the constructor or the process spectrums function and the predefined sets be available as global variables - this would make using the processor a lot easier and it has to store less information.

matchms/SpectrumProcessor.py
hechth, line 1: If the filters are passed on initialization or when processing the spectra, the whole adding filters logic is not required.

matchms/Pipeline.py
niekdejonge: This has been fixed in #480

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+                processing_report.add_to_report(spectrum, spectrum_out, filter_func.__name__)
+            if spectrum_out is None:
+                break
+            spectrum = spectrum_out
+        return spectrum_out
+
+    def process_spectrums(self, spectrums: list,
+                          create_report=False,
+                          progress_bar=True,
+                          ):
```
hechth, line 136: Add return type hint - list can be specfied, also not sure if python < 3.11 accepts those styles of type hints (list)

matchms/SpectrumProcessor.py
```diff
@@ -0,0 +1,329 @@
...
+
+        Returns
+        -------
+        Spectrums
+            List containing the processed spectrums.
+        """
+        if create_report:
+            processing_report = ProcessingReport()
+        else:
+            processing_report = None
```
hechth, line 158: just make creating the report non optional - users can ignore the returned value from it.

this makes the function have a constant return type and reduces the number of parameters for this function. Also, then the spectrum processor doesn't have to deal with None values.

matchms/SpectrumProcessor.py
niekdejonge: I created an issue about this #485

matchms/filtering/metadata_processing/clean_compound_name.py
```diff
@@ -69,11 +25,58 @@ def remove_misplaced_mass(name):
...
     if name_cleaned != name:
         spectrum.set("compound_name", name_cleaned)
         logger.info("Added cleaned compound name: %s", name_cleaned)
 
     return spectrum
+
+
+def _remove_parts_by_regular_expression(name):
+    """Clean name string by removing known parts that don't belong there."""
+    name = name.strip()
```
hechth, line 40: add missing type hints

hechth (changes requested):
I personally would prefer moving to a object oriented system for processing the spectra and using the filters to be able to share functionality and reduce code duplication.

florian-huber:
> I personally would prefer moving to a object oriented system for processing the spectra and using the filters to be able to share functionality and reduce code duplication.

@hechth do you mean to have all filters in a OO design? 
Niek and I also discussed this a bit (see #432).

matchms/SpectrumProcessor.py
florian-huber: You are right, that would make the method much cleaner.
I was implement as it is now to have the option to skip reporting under the assumption that this consumes compute time. All the in-out spectrum comparisons sum up to many operations in total... but I haven't done any benchmarking yet!

If the difference in computing time is small, I would fully agree to make this non optional.

matchms/filtering/metadata_processing/clean_compound_name.py
florian-huber: Done (will be added in separate PR)

matchms/SpectrumProcessor.py
florian-huber: Done (will be added in separate PR)

matchms/Pipeline.py
florian-huber: @niekdejonge see Helges comments on the Pipeline part here and below. Probably best if you check what applied to your latest PRs on the Pipeline part?

matchms/Pipeline.py
florian-huber: In principle that makes sense. But I do not have a quick fix for this because the Pipeline logging file has a somewhat different logic than the normal matchms logger (for me at least). With the normal logger I have to decide the logging level, but while I would under all circumstances add the created report to the logging file it somehow doesn't fit to the level descriptions such as "warning" or higher. With the method called here, I am sure it ends up in the logging file no matter if user set the threshold to info, warning, etc.

I guess it best is made into a new issue?

hechth:
Yes, since some filters share quite a lot of functionality which could be put into a single place this way - specifically about things like logging or exception handling - some of this can also be put into the `SpectrumProcessor` class but I think we should first actually sketch out how to implement it. All existing functions can also be mapped to the object oriented design and marked as deprecated to not break the current API.

As of now, with all filters being individual functions, I find it quite hard to harmonize the behaviour, logging messages etc. - these things could then be handled by the base classes and on the highest layer, everything is just implementing the `Filter` interface which has a single `apply` function or so, and then the specialized base classes can implement the specific behaviour - like the require filters either return the same object or None - this could then even be checked in a `RequireFilter` class where `apply` runs the implementation of the specific filtering method (let's call it `require` in this case) and then checks whether the spectrum is still the same or None - else it should throw an exception or so. Overall, the `Template method` pattern can be used here really nicely.

@zargham-ahmad and me have been sketching out an implementation based on this design choice - maybe we can move this implementation to a specific feature branch for the filters before moving it into the main codebase?

Issue #434, opened 2023-06-06
Convert filtering functions to classes


niekdejonge:
The filtering steps now are functions with similar code structure. 

It would make sense to create a parent class for the FilterMethod, to formalize what is expected of these filtering functions. Just like we did with the MetadataFeatureGenerator in MS2Deepscore.

Benefit would be cleaner methods and easier to add general new functionality.
A few things that are now already standard are:
input is type Spectrum.
Always run:
if spectrum_in is None:
    return None

spectrum = spectrum_in.clone()

Output is of type Optional[Spectrum]

We could also make more specific Classes. Some ideas:
CleanMetadata which can only access metadata of spectra
CleanPeaks which can only change peaks and intensities
Require which can not change spectra, but only checks if a requirement is met and returns None or the Spectrum. 
This is closely related to #432 

It would make it easier to implement future changes that apply to every filtering function. It would also make it possible to more safely use your own filtering methods in pipeline, since they would require to be of type FilterMethod. 

Downside might be that not all people in the community are very familiar with using classes, with might make it less accessible to contribute. 

What do you think @florian-huber @hechth

niekdejonge:
Also code like in add_parent_mass (and many other filtering functions): 
logger = logging.getLogger("matchms")

_default_key = "parent_mass"
_accepted_keys = ["parentmass", "exact_mass"]
_accepted_types = (float, str, int)
_accepted_missing_entries = ["", "N/A", "NA", "n/a"]

can be included in init.

niekdejonge:
It might also be possible to add a method for required pre filtering for each filter step, that specify which other filter steps are expected to already have been done. This could be used in Pipeline, during the precheck if the order of filtering steps makes sense. For instance you should not run add_parent_mass before derive_ionmode or correct_charge, since it depends on the entry of charge and ionmode.

florian-huber:
As discussed in person, I would prefer to stick to the filter functions. But as outlined in #432, I would rather try adding a `SpectrumProcessor` class that will take care of running those functions. This could include:
- A few default settings, e.g., `minimal`, `default`, `full` processing options (or also a flag for `GCMS` or `LCMS`?)
- Also the option to list all filters that should be applied, the class will then take care of the right order of those filters (e.g., add, repair etc. parent mass before running a filter relying on it).

--> See #455.

florian-huber:
[...]

PR #472, opened 2023-08-09
Change processing report to store filter step instead of field

niekdejonge:
Instead of reporting the fields that were changed, this PR will result in returning the changes made by each filter step. 

----- Spectrum Processing Report -----
Number of spectrums processed: 3
Number of spectrums removed: 0
Changes during processing:
removed spectra changes additions
filter
make_charge_int 0 2 0
interpret_pepmass 0 0 3
derive_ionmode 0 0 3

I have implemented this now and will make a PR after I fixed the tests

### DIFF
--- a/matchms/SpectrumProcessor.py
+++ b/matchms/SpectrumProcessor.py
@@ -1,10 +1,11 @@
-from typing import Optional, Tuple, List
 from collections import defaultdict
 from functools import partial
+from typing import List, Optional, Tuple
 import pandas as pd
 from tqdm import tqdm
 import matchms.filtering as msfilters
 from matchms import Spectrum
+
 
 class SpectrumProcessor:
     """
--- a/matchms/SpectrumProcessor.py
+++ b/matchms/SpectrumProcessor.py
@@ -115,12 +115,13 @@ class SpectrumProcessor:
         if processing_report is not None:
             processing_report.counter_number_processed += 1
         for filter_func in self.filters:
-            spectrum = filter_func(spectrum)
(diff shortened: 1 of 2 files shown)


matchms/SpectrumProcessor.py
```diff
@@ -91,32 +92,41 @@ def add_custom_filter(self, filter_function, filter_params=None):
...
             The processed spectrum.
         """
         if self.filters == []:
             raise TypeError("No filters to process")
+        if processing_report is not None:
+            processing_report.counter_number_processed += 1
         for filter_func in self.filters:
-            spectrum = filter_func(spectrum)
-            if spectrum is None:
+            spectrum_out = filter_func(spectrum.clone())
```
niekdejonge: Here I added .clone() By adding it here it will not be necessary anymore to do this in each filter function. Which I think would be a nice and saver way. However, with the current filter functions this will result in cloning twice, which is time consuming. We can remove this here, remove the spectrum.clone() from the filter functions, or accept the extra runtime for extra safety. What do you think?

florian-huber: Even though it looks like a massive code duplication, I would leave the clone() part in each filter because that is what I would expect an individual filter function to do (and not to change the spectrum object itself). 
There is still the option to streamline this more in the near future as you suggested Niek. I also just added some ideas to your issue #434 on this.

florian-huber: Here, we could hence skip it (to avoid cloning twice).

Issue #476, opened 2023-08-11
Add test that checks that ALL_FILTERS contains all filter functions


niekdejonge:
Since this list is used to determine order, it is important that this is complete. When adding new tests it is easy to forget, therefore it is important to add a test that checks if all functions available are in the ALL_FILTERS list. This is hard to implement, since there are also helperfunctions in the folder filtering. 

This can probably be implemented after adding a decorator or creating classes, see #455

niekdejonge:
A quick solution is to check if the length of ALL_FILTERS is equal to the number of scripts in the folders metadata_processing and peak_processing (excluding init)

PR #477, opened 2023-08-11
Complete filter order

niekdejonge:
Solves #476 

Added a test to check if ALL_FILTERS actually contains all the available filters.

The missing filters have been added and to helper functions (in the filter folders) a "_" has been added to the function name.


florian-huber (approved):
Good addition and nice linting! Thanks @niekdejonge !

Issue #475, opened 2023-08-10
Pipeline refactoring


niekdejonge:
The Pipeline class became quite complex by now. The following two changes might make using Pipeline a bit easier: 

An option that might make sense would be splitting Pipeline into two parts. The first Class will just take in a Yaml file and run this and a separate function that can be used to create new Yaml files. 

I think currently Yaml files also contain the spectrum files location. To me, this does not really make sense, since you would like to share Yaml files with others. I would argue using a Yaml file only for the pipeline and specifying the query files when running pipeline. Your logfile on the other hand should store the filenames.

PR #479, opened 2023-08-14
Move yaml file creation and loading outside Pipeline.py

niekdejonge:
This should solve #475 
The PR's #482 #481 #480 build further onto this PR. 

Pipeline became quite a complicated class. I wanted to implement some changes to the way the yaml file was stored, but this was a bit challenging due to the fact that yaml file/workflow is used although Pipeline, making it harder to make changes.

As a solution the yaml file creation is moved outside of Pipeline. 

These changes will remove the option to modify an already existing Pipeline object. So you cannot run Pipeline and after add some steps and than run again. I think this was unwanted behaviour anyway, since this can result in having a difference between your Yaml file and the actual pipeline that was run, which is unwanted.

### DIFF
--- a/integration-tests/test_user_workflow.py
+++ b/integration-tests/test_user_workflow.py
@@ -9,18 +9,15 @@ from matchms.similarity import CosineGreedy
 def test_user_workflow():
     module_root = os.path.join(os.path.dirname(__file__), "..")
     spectrums_file = os.path.join(module_root, "tests", "testdata", "pesticides.mgf")
-    workflow = create_workflow()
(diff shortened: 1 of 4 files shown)


niekdejonge:
@florian-huber and @hechth I have done quite some refactoring of Pipeline and the Yaml file format in the PR #479 #480 #481 and #482 . Let me know if you agree with these changes, or if you would prefer the old functionality. Before merging I would suggest to first merge the add_filter_class PR #455, since this PR builds on that.

florian-huber (approved):
Nice refactoring @niekdejonge !

Issue #478, opened 2023-08-11
Store all filters in yaml file instead of predefined_processing_queries


niekdejonge:
Currently in #455 the yaml files created will look something like:
``` yaml
# Matchms pipeline config file 
# Change and adapt fields where necessary 
# ==================== 
importing:
  queries: ../tests/testdata/pesticides.mgf
  references: ../tests/testdata/pesticides.mgf
predefined_processing_queries: minimal
predefined_processing_references: minimal
additional_processing_queries:
- - add_parent_mass
- - normalize_intensities
additional_processing_references: []
score_computations: []
logging:
  logging_file: ./local_data/my_pipeline.log
  logging_level: WARNING
```

Instead it would be better to store every individual filter step instead of the predefined processing_queries. This would look like:

``` yaml
# Matchms pipeline config file 
# Change and adapt fields where necessary 
# ==================== 
importing:
  queries: ../tests/testdata/pesticides.mgf
  references: ../tests/testdata/pesticides.mgf
filters_queries:
-- make_charge_int
-- interpret_pepmass
-- derive_ionmode
-- correct_charge
- - add_parent_mass
- - normalize_intensities
score_computations: []
logging:
  logging_file: ./local_data/my_pipeline.log
  logging_level: WARNING
```

This ensures that a yaml file always does the same analysis even if we later change the filters in "minimal". It also gives a clearer overview of what filters are used and in which order they are run.

florian-huber:
@niekdejonge this is now taken care of, right?

PR #480, opened 2023-08-14
Add individual filters in yaml

niekdejonge:
Solves #478 

Instead of storing "default" in the pipeline we will store the actual filters called in the order they are run in a yaml file. 

This makes it possible to run a yaml file of an older version of matchms (with different default filters) and still getting the same results. 
It also gives me a quick check that the default filters actually do what I expect them to do as an user, without having to dive into the matchms code.

### DIFF
--- a/tests/test_pipeline.yaml
+++ b/tests/test_pipeline.yaml
@@ -4,13 +4,27 @@
 importing:
   queries: ./tests/testdata/massbank_five_spectra.msp
   references:
-predefined_processing_queries: default
-predefined_processing_references: default
-additional_processing_queries:
+query_filters:
+- - make_charge_int
+- - interpret_pepmass
+- - derive_ionmode
+- - correct_charge
+- - add_compound_name
+- - derive_adduct_from_name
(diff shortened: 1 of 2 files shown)


florian-huber (approved):
@niekdejonge I am completely fine with this (but have no strong opinion on default vs having the individual filters). It probably is the cleaner, certainly the more reproducible solution!

tests/test_pipeline.yaml
```diff
@@ -4,13 +4,27 @@
...
+- - derive_formula_from_name
+- - clean_compound_name
+- - add_precursor_mz
+- - require_precursor_mz
+- - add_parent_mass
+- - harmonize_undefined_inchikey
+- - harmonize_undefined_inchi
+- - harmonize_undefined_smiles
+- - repair_inchi_inchikey_smiles
+- - repair_parent_mass_match_smiles_wrapper
```
hechth, line 23: Why the double dashes?

hechth: @niekdejonge

PR #481, opened 2023-08-14
Remove file names from yaml

niekdejonge:
Currently the yaml file stores the query and reference files.
This is not intuitive to me when sharing pipelines, since a path will always be different. To me it would make more sense to have a pipeline file store all the steps, but specifying the query files and reference files in the python script were Pipeline is defined. Instead the log file should store the query file and reference file location. 

- [x] Remove spectrum files from yaml file
- [x] Add spectrum files to logging
- [x] Specify file names in run, instead of as Pipeline attributes

(Docstring will be updated, once the final setup is selected)


florian-huber (approved):
Good changes. Removing the filenames from the yaml files makes a lot of sense to me!

PR #482, opened 2023-08-14
Add check for correct workflow

niekdejonge:
Made changes not implemented yet in #475, since it was easier to make implementations in #481 #479 #480 first.

To do:
- [x] Add extra checks for correct workflow
- [x] Remove logging and logging file from yaml
- [x] Upgrade docstring (I will first wait for review, before finalizing the dosctrings)
- [x] Add changes to changelog

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -148,33 +148,46 @@ Introduction
 
 To get started with matchms, we recommend following our `matchms introduction tutorial <https://blog.esciencecenter.nl/build-your-own-mass-spectrometry-analysis-pipeline-in-python-using-matchms-part-i-d96c718c68ee>`_.
 
-Below is a small example of using matchms to calculate the Cosine score between mass Spectrums in the `tests/testdata/pesticides.mgf <https://github.com/matchms/matchms/blob/master/tests/testdata/pesticides.mgf>`_ file.
+Below is an example of using default filter steps for cleaning spectra, 
+followed by calculating the Cosine score between mass Spectrums in the `tests/testdata/pesticides.mgf <https://github.com/matchms/matchms/blob/master/tests/testdata/pesticides.mgf>`_ file.
 
 .. code-block:: python
 
-    from matchms import Pipeline
-    
-    pipeline = Pipeline()
-    
-    # Read spectrums from a MGF formatted file, for other formats see https://matchms.readthedocs.io/en/latest/api/matchms.importing.html 
-    pipeline.query_files = "tests/testdata/pesticides.mgf"
-    pipeline.prefedined_processing_queries = "basic"
-    pipeline.additional_processing_queries = [
-        ["add_parent_mass"],
-        ["normalize_intensities"],
-        ["select_by_intensity", {"intensity_from": 0.001, "intensity_to": 1.0}],
-        ["select_by_mz", {"mz_from": 0, "mz_to": 1000}],
-        ["require_minimum_number_of_peaks", {"n_required": 5}]
-    ]
-    pipeline.score_computations = [["precursormzmatch",  {"tolerance": 100.0}],
-                                   ["cosinegreedy", {"tolerance": 1.0}],
-                                   ["filter_by_range", {"name": "CosineGreedy_score", "low": 0.2}]]
+    from matchms.Pipeline import Pipeline, create_workflow
 
+    workflow = create_workflow(
+        yaml_file_name="my_config_file.yaml", # The workflow will be stored in a yaml file, this can be used to rerun your workflow or to share it with others.
+        score_computations=[["cosinegreedy", {"tolerance": 1.0}]],
+        )
+    pipeline = Pipeline(workflow)
     pipeline.logging_file = "my_pipeline.log"  # for pipeline and logging message
-    pipeline.logging_level = "INFO"
-    pipeline.run()
+    pipeline.run("tests/testdata/pesticides.mgf")
+    
+Below a more advanced code example showing how you can make a specific pipeline for your needs.
 
+.. code-block:: python
 
+    from matchms.Pipeline import Pipeline, create_workflow
+
+    workflow = create_workflow(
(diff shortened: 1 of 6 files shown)


matchms/Pipeline.py
```diff
@@ -326,7 +324,7 @@ def write_to_logfile(self, line):
             with open(self.logging_file, "a", encoding="utf-8") as f:
                 f.write(line + '\n')
 
-    def import_data(self, query_files, reference_files=None):
+    def import_spectra(self, query_files, reference_files=None):
```
florian-huber: Somehow it became a matchms&co tradition to use "spectrums" in such cases. Since we also use it elsewhere in matchms like this I would change it to `import_spectrums`.

matchms/Pipeline.py
```diff
@@ -378,28 +376,16 @@ def check_files_exist(filenames):
 
     # Getter & Setters
     @property
-    def logging_file(self):
-        return self.workflow["logging"].get("logging_file")
-
-    @logging_file.setter
-    def logging_file(self, file):
-        self.workflow["logging"]["logging_file"] = file
```
florian-huber, line 385: Why remove the getters/setter?

They seem verbose, but also are a robust way to define interactions with a Pipeline object. If it's about lines of code, there are other ways to shorten this part.

florian-huber: E.g, something like:
```python
_workflow_keys_mapping = {
        'query_files': ("importing", "queries"),
        'reference_files': ("importing", "references"),
        'logging_file': ("logging", "logging_file"),
        'logging_level': ("logging", "logging_level"),
        'predefined_processing_queries': ("predefined_processing_queries",)
    }

    def __getattr__(self, name):
        if name in self._workflow_keys_mapping:
            pointer = self.workflow
            for key in self._workflow_keys_mapping[name]:
                pointer = pointer.get(key, {})
            return pointer
        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")

    def __setattr__(self, name, value):
        if name in self._workflow_keys_mapping:
            pointer = self.workflow
            keys = self._workflow_keys_mapping[name]
            for key in keys[:-1]:
                pointer = pointer.setdefault(key, {})
            pointer[keys[-1]] = value
        else:
            super().__setattr__(name, value)
```

florian-huber:
@niekdejonge the checks for correct workflow seemed fine to me!
I added a few comments and questions on other points.

Plus: Since this seems to be the last of the PRs you just added it would be good to add all PR main changes to the Changelog.

matchms/Pipeline.py
niekdejonge: @florian-huber My motivation was not so much the number of lines of code.

The setter/getter for logging was removed since it is not an attribute of workflow anymore, instead, it is now an attribute of Pipeline, so the setter and getter were now redundant. 

In addition, the setters were removed for all workflow settings, since I did not want it to be possible to change a workflow after a pipeline object is created. Since the SpectrumProcessor objects are initialized in __init__ of workflow, the problem with later changing anything in workflow is that the SpectrumProcessor objects (self.processing_queries) will not be updated. Creating unexpected behavior when running pipeline. 

A solution could be still creating setters, that after setting call _initialize_spectrum_processor_queries. That would look like:
```python
@query_filters.setter
def query_filters(self, query_filters):
    self.workflow["query_filters"] = query_filters
    self._initialize_spectrum_processor_queries
```

niekdejonge: Will be implemented in #483

README.rst
```diff
@@ -146,33 +148,46 @@ Introduction
...
-        ["add_parent_mass"],
-        ["normalize_intensities"],
-        ["select_by_intensity", {"intensity_from": 0.001, "intensity_to": 1.0}],
-        ["select_by_mz", {"mz_from": 0, "mz_to": 1000}],
-        ["require_minimum_number_of_peaks", {"n_required": 5}]
-    ]
-    pipeline.score_computations = [["precursormzmatch",  {"tolerance": 100.0}],
-                                   ["cosinegreedy", {"tolerance": 1.0}],
-                                   ["filter_by_range", {"name": "CosineGreedy_score", "low": 0.2}]]
+    from matchms.Pipeline import Pipeline, create_workflow
```
niekdejonge, line 156: @florian-huber I have updated the readme, to work with the new version. 
Feel free to make any changes. 
The tutorials still point to the older version. Maybe you could update the tutorial before or after the workshops?

florian-huber (approved):
Looks good to me Niek. We can simply see what users want and if anything is weird/missing/unclear about the API for future PRs and releases.

matchms/Pipeline.py
```diff
@@ -50,9 +72,11 @@ def initialize_spectrum_processor(predefined_workflow, additional_filters):
     return processor
 
 
-def load_in_workflow_from_yaml_file(yaml_file):
+def load_workflow_from_yaml_file(yaml_file):
```
hechth: add type hints and return type annotations

matchms/Pipeline.py
```diff
@@ -83,61 +107,87 @@ def _dict_representer(dumper, data):
     return yaml.dump(data, stream, OrderedDumper, **kwds)
 
 
+def check_score_computation(score_computations):
```
hechth: add argument type annotation and documentation

matchms/Pipeline.py
```diff
@@ -325,7 +354,7 @@ def write_to_logfile(self, line):
             with open(self.logging_file, "a", encoding="utf-8") as f:
                 f.write(line + '\n')
 
-    def import_data(self, query_files, reference_files=None):
+    def import_spectrums(self, query_files, reference_files=None):
```
hechth: Some of the functionality of this function could be moved to the general load or import spectra function, such as the error handling etc.

niekdejonge: That certainly makes sense, I have moved most of the functionality to load_spectrums

PR #483, opened 2023-08-15
Add workflow setters

niekdejonge:
Solves suggestion of @florian-huber 

Adds setters for workflow attributes back in to prevent changing anything in workflow that will not be passed on the SpectrumProcessor.

### DIFF
--- a/matchms/Pipeline.py
+++ b/matchms/Pipeline.py
@@ -174,7 +174,7 @@ class Pipeline:
         self.logging_level = logging_level
         self.logging_file = logging_file
         self.progress_bar = progress_bar
-        self.workflow = workflow
+        self.__workflow = workflow
         self.check_workflow()
 
         self.write_to_logfile("--- Processing pipeline: ---")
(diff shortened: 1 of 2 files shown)


florian-huber (approved):
Fine with making workflow a hidden attribute! Thanks Niek.

matchms/Pipeline.py
```diff
@@ -377,15 +376,30 @@ def check_files_exist(filenames):
     # Getter & Setters
     @property
     def score_computations(self):
-        return self.workflow.get("score_computations")
+        return self.__workflow.get("score_computations")
```
hechth: please add return type annotations

matchms/Pipeline.py
```diff
@@ -377,15 +376,30 @@ def check_files_exist(filenames):
...
+
+    @score_computations.setter
+    def score_computations(self, computations):
+        self.__workflow["score_computations"] = computations
+        check_score_computation(score_computations=self.score_computations)
 
     @property
     def query_filters(self):
```
hechth: same here - return types would be very helpful!

matchms/Pipeline.py
```diff
@@ -377,15 +376,30 @@ def check_files_exist(filenames):
...
-        return self.workflow.get("query_filters")
+        return self.__workflow.get("query_filters")
+
+    @query_filters.setter
+    def query_filters(self, filters):
```
hechth: type hints for the filters parameter are missing

matchms/Pipeline.py
```diff
@@ -377,15 +376,30 @@ def check_files_exist(filenames):
...
+        self.__workflow["query_filters"] = filters
+        self._initialize_spectrum_processor_queries()
 
     @property
     def reference_filters(self):
-        return self.workflow.get("reference_filters")
+        return self.__workflow.get("reference_filters")
+
+    @reference_filters.setter
+    def reference_filters(self, filters):
```
hechth: type hints for the filters parameter are missing

PR #484, opened 2023-08-15
logging level adjustments

florian-huber:
Whenever people work with unknown / not annotated data they will see warnings saying "No compound name found" all the time... although that is not really a warning sign in many cases. So I "downgraded" it to info-level.


niekdejonge (approved):
Yes, good change. This certainly is more a info than a warning.

PR #487, opened 2023-08-16
remove unnecessary warning

florian-huber:
This is just to remove an unnecessary warning currently happening if a precursor_mz value `None` is replaced by a new value.


florian-huber:
@hechth The CI run suddenly started to fail for all Windows runs and it seems to be related to issues with pep 517. Any ideas what's the matter here? 

I am not yet familiar with the new poetry build, but it seemed related to this: https://github.com/python-poetry/poetry/issues/7611

florian-huber:
OK ... good exercise to get more used to poetry.
Running `poetry update` in a new environment seems to have fixed this.

PR #493, opened 2023-08-21
Fix filter parameter setting

### LINKED ISSUE #492: Filter parameters in default pipelines are not used
When adding filter parameters in the default filters they are not used by SpectrumProcessor. 

For instance in the code below "ion_mode_to_keep" is not set to "both"

```python
FULLY_ANNOTATED_PROCESSING = DEFAULT_FILTERS \
    + ["clean_adduct",
       "derive_inchi_from_smiles",
       "derive_smiles_from_inchi",
       ("require_correct_ionmode", {"ion_mode_to_keep": "both"}),
       ("require_parent_mass_match_smiles", {'mass_tolerance': 0.1}),
       "require_valid_annotation",
    ]
```

niekdejonge:
Fixes #492 
The filter parameters were not passed on when creating a workflow, since they are not returned in the property processing_steps.

### DIFF
--- a/matchms/Pipeline.py
+++ b/matchms/Pipeline.py
@@ -214,24 +214,25 @@ class Pipeline:
         self.__workflow = workflow
         self.check_workflow()
 
-        self.write_to_logfile("--- Processing pipeline: ---")
         self._initialize_spectrum_processor_queries()
         if self.is_symmetric is False:
             self._initialize_spectrum_processor_references()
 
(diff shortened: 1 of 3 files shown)


florian-huber (approved):
Nice catch @niekdejonge !
(I wonder, should there also be a specific test for this?)
Appart from that, feel free to merge!

niekdejonge:
@florian-huber tests were adjusted and added in #496 that should now catch this issue.

PR #496, opened 2023-08-21
Add default settings to processing steps

### LINKED ISSUE #494: Add default filters to yaml file
Any default settings of functions are not shown in the yaml file. This is bad for reproducibility, since changing any default settings in a filter function will currently result in a different pipeline running. 

Suggested change:
Automatically get the default filter settings from each filter and returning them in properties. (this will be used to create yaml files and printing the log file)

niekdejonge:
Fixes #494 
Any default settings of functions are not shown in the yaml file. This is bad for reproducibility, since changing any default settings in a filter function will currently result in a different pipeline running.

Instead we will now automatically get the default filter settings combined with the defined parameters from each filter and returning them in properties. (this will be used to create yaml files and printing the log file)

### DIFF
--- a/matchms/Pipeline.py
+++ b/matchms/Pipeline.py
@@ -214,24 +214,25 @@ class Pipeline:
         self.__workflow = workflow
         self.check_workflow()
 
-        self.write_to_logfile("--- Processing pipeline: ---")
         self._initialize_spectrum_processor_queries()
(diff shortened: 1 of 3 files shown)


florian-huber (approved):
thanks @niekdejonge, looks like good additions to me!

PR #498, opened 2023-08-21
Spectrum processor sorting

niekdejonge:
The sorting algorithm was a bit counter intuitive. 
The custom_filters would be added just after the first of the already added filters.

I have now added the option to specify a position, which will just add it to the end if it is not given.


florian-huber (approved):
I added a tiny change (hope that's fine for you). Apart from that this looks good!

PR #499, opened 2023-08-21
Improve default pipelines

niekdejonge:
Sorry for this large PR, some of the rebasing went a bit wrong, so I combined two PR in one. Hope it is still clear.

The default pipelines for library cleaning are not yet complete and tests for the filter order are limited.

- [x] Add tests for known orders in order filter
- [x] Add tests missing from the predefined_pipelines
- [x] Move ALL_FILTERS to filter_order_and_default_pipelines
- [x] Moved metadata_utils to filter utils

In addition #500 ProgressReport can count one spectrum multiple times is fixed.


matchms/filtering/__init__.py
```diff
@@ -121,8 +121,8 @@
 from .metadata_processing.require_precursor_below_mz import \
     require_precursor_below_mz
 from .metadata_processing.require_precursor_mz import require_precursor_mz
-from .metadata_processing.require_valid_annotation import \
-    require_valid_annotation
+from .metadata_processing.require_valid_annotation import (
+    repair_not_matching_annotation, require_valid_annotation)
```
florian-huber, line 125: I thought those functions should not be in the same file --> see my suggetsions in  #513 (but that can be discussed there)

matchms/filtering/metadata_processing/repair_adduct_based_on_smiles.py
```diff
@@ -11,7 +11,7 @@
 
 def repair_adduct_based_on_smiles(spectrum_in: Spectrum,
                                   mass_tolerance,
-                                  accept_parent_mass_is_mol_wt):
+                                  accept_parent_mass_is_mol_wt = True):
```
florian-huber: add type hints

matchms/filtering/metadata_processing/repair_adduct_based_on_smiles.py
```diff
@@ -11,7 +11,7 @@
...
     """If the parent mass is wrong due to a wrong of is derived from the precursor mz
     To do this the charge and adduct are used"""
```
florian-huber:
```suggestion
    """
    Corrects the adduct of a spectrum based on its SMILES representation and the precursor m/z.

    Given a spectrum, this function tries to match the spectrum's parent mass, derived from its
    precursor m/z and known adducts, to the neutral monoisotopic mass of the molecule derived
    from its SMILES representation. If a match is found within a given mass tolerance, the 
    adduct and parent mass of the spectrum are updated.

    Parameters:
    ----------
    spectrum_in : Spectrum
        The input spectrum whose adduct needs to be repaired.

    mass_tolerance : float
        Maximum allowed mass difference between the calculated parent mass and the neutral 
        monoisotopic mass derived from the SMILES.

    accept_parent_mass_is_mol_wt : bool, optional (default=True)
        Allows the function to attempt repairing the spectrum's parent mass by assuming it 
        represents the molecule's weight. If True, further checks and corrections are made 
        using `repair_parent_mass_is_mol_wt`.
    """
```

niekdejonge:
@florian-huber, this can now be merged. But I will wait until you had a look at restructure_metadata_utils, since that PR will otherwise very hard to follow, since this branch merges into restructure_metadata_utils.

matchms/filtering/__init__.py
florian-huber: Sorry, I am now just a bit confused about the order of the different PRs and branches...
This one should be solved by other PRs, right?

matchms/filtering/metadata_processing/repair_smiles_of_salts.py
```diff
@@ -17,6 +19,9 @@ def repair_smiles_of_salts(spectrum_in,
     spectrum = spectrum_in.clone()
 
     smiles = spectrum.get("smiles")
+    if not is_valid_smiles(smiles):
+        logger.warning("No valid smiles, please run require_valid_annotation before")
```
florian-huber, line 26: I think this is too focused on workflows on annotated data.
In many cases people will run default pipelines for data that is largely not annotated, in which case they will be flooded with warnings :)

Since we also offer no filter tree structure (such as "if valid smiles run this filter..."), we should be less strict here and also accept that there is simply no smiles --> `if smiles is not None and not is_valid_smiles(smiles):`

florian-huber: Or, better I think to add:
```python
if smiles is None:
    return spectrum
```

matchms/filtering/metadata_processing/require_valid_annotation.py
```diff
@@ -1,13 +1,63 @@
...
+    convert_smiles_to_inchi, is_valid_inchi, is_valid_inchikey,
+    is_valid_smiles)
+from matchms.filtering.metadata_processing.require_parent_mass_match_smiles import \
+    _check_smiles_and_parent_mass_match
 
 
 logger = logging.getLogger("matchms")
 
 
+def repair_not_matching_annotation(spectrum_in: Spectrum):
```
florian-huber, line 14: This will have changed with #513, right?

florian-huber (changes requested):
I would allow spectra without smiles to skip the warning (see comment in code).

matchms/filtering/metadata_processing/repair_smiles_of_salts.py
florian-huber: @niekdejonge I added this bit of code. Hope that works.

PR #505, opened 2023-08-23
Improve adding filters in SpectrumProcessor

niekdejonge:
Improved the adding of spectra to SpectrumProcessor:
- Add custom filter, now adds at the end of the filter list, unless a filter_position is applied.
- Added a check that all needed parameters are given for a filter.
- Tests were updated accordingly

Repair not matching annotation was moved to a separate function and refactored.

### DIFF
--- a/matchms/filtering/metadata_processing/require_valid_annotation.py
+++ b/matchms/filtering/metadata_processing/require_valid_annotation.py
@@ -1,8 +1,8 @@
 import logging
 from matchms import Spectrum
 from matchms.filtering.filter_utils.smile_inchi_inchikey_conversions import (
-    convert_inchi_to_inchikey, convert_smiles_to_inchi, is_valid_inchi, is_valid_inchikey,
-    is_valid_smiles)
+    convert_inchi_to_inchikey, convert_smiles_to_inchi, is_valid_inchi,
+    is_valid_inchikey, is_valid_smiles)
 
 
 logger = logging.getLogger("matchms")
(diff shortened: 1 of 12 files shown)


matchms/filtering/metadata_processing/require_valid_annotation.py
```diff
@@ -1,13 +1,63 @@
...
+    convert_smiles_to_inchi, is_valid_inchi, is_valid_inchikey,
+    is_valid_smiles)
+from matchms.filtering.metadata_processing.require_parent_mass_match_smiles import \
+    _check_smiles_and_parent_mass_match
 
 
 logger = logging.getLogger("matchms")
 
 
+def repair_not_matching_annotation(spectrum_in: Spectrum):
```
florian-huber: This should be in a separate file `require_not_matching_anotation` like the other filters. According to the other files I was not looking in this python file for it.

matchms/filtering/metadata_processing/require_valid_annotation.py
```diff
@@ -1,13 +1,63 @@
...
+    """When smiles and inchi do not match the annotation that matches parent mass will be kept"""
```
florian-huber: This needs a more extensive docstring, since the name of the function "repair ... annotation" promises quite a lot ;)

florian-huber:
Hi @niekdejonge, here some suggestions to refactor the code a bit --> #513 
(should address my comments above)

florian-huber (approved):
Looks good @niekdejonge !

hechth:
@niekdejonge would it be possible to add a description for the PR? It is very difficult to retrospectively find out what was done from the code.

niekdejonge:
@hechth Sorry about that. I have added it to the description

PR #509, opened 2023-08-25
Add pubchem lookup

niekdejonge:
Previously in #427 this was implemented in a style that did not match the matchms style, so it better fitted in matchmsextras. 
However, when working with Pipeline, I noticed this was not user friendly. 
The problem is that the pubchem annotation needs to be done, after the first metadata cleaning, but before the final require filters. This would mean switching from Pipeline, to matchmsextras to a final Pipeline, which is unnecesary complex. 

The previous version required a file that already contained compound annotations. Now this is optional, it can be run with and without csv file. If you run it without csv file it will cache previously looked up compound_names reducing the number of times pubchem lookup has to be performed (since this is an expensive step). If a csv file is supplied, lookup will still be performed if the csv file does not contain the compound name and this will be added to the csv file, which will make future pipeline cleaning runs faster, since this csv file can be used as starting point. 

This means pubchempy has to be added as a dependency. However, I don't think this is a limitation, since it only depends on pandas, which is already a requirement of matchms. 

I tried specifying version of pubhempy to 1.0.4 in the poetry file, but I did not yet get this to work. Does one of you know how to get this to work?


florian-huber:
> I tried specifying version of pubhempy to 1.0.4 in the poetry file, but I did not yet get this to work. Does one of you know how to get this to work?

@niekdejonge this no longer applied, right?

matchms/filtering/metadata_processing/repair_smiles_from_compound_name.py
```diff
@@ -10,8 +15,8 @@
 
 
 def repair_smiles_from_compound_name(spectrum_in: Spectrum,
-                                     annotated_compound_names_file,
-                                     mass_tolerance=0.1):
+                                     annotated_compound_names_file: Optional[str] = None,
+                                     mass_tolerance: float = 0.1):
     """Adds annotations (smiles, inchi, inchikey) based on compound name
 
     Based on a table of compound names and smiles matches (stored in a csv file) this function
```
florian-huber: The docstring is outdated/not correct. It should mention more clearly that it also contains a PubChem library search.

That being said, it might also be better to make PubChem more prominent here. Maybe `_get_pubchem_compound_name_annotation` or similar. Plus, we might (now or in the future) also add a flag to specify different databases?

niekdejonge: Helge also mentioned that they already have a package that has that kind of functionality. Since it builds on matchms it was not really easy to build directly into pipeline. Therefore I did in the end still build the pubchem lookup in here, instead of using their package. But it might be better to later use his package for this, to prevent implementing the same thing twice. 

I will update the docstring.

niekdejonge:
> > I tried specifying version of pubhempy to 1.0.4 in the poetry file, but I did not yet get this to work. Does one of you know how to get this to work?
> 
> @niekdejonge this no longer applied, right?

@florian-huber No it works now, I was not able to specify the version though, so it will just use the most up to date version.

niekdejonge:
During the fixing of the Docstring I noticed some mistakes in the filter order, so these have also been fixed.

florian-huber:
@niekdejonge through the first few PRs I updated the `restructure_metadata_utils` branch to the current master branch. Also the PRs introduced several things that seem out of sync. So this and the following PRs need to be adjusted regarding merge conflicts.

niekdejonge:
@florian-huber, the merge conflicts should now be resolved

PR #513, opened 2023-08-30
refactoring

florian-huber:
Attempt to restructure some of the code from #505.


niekdejonge (approved):
Yes, it makes more sense to make a separate script for this!

PR #523, opened 2023-09-06
Improve derive adduct from name

niekdejonge:
Some compound names had multiple adducts in the name. These were previously not removed. Now improve_derive_adduct_form_name does remove every occurance of an adduct and adds only an adduct if it can actually be interpreted to retrieve a mass_multiplier and a charge. 

@florian-huber Did I miss any type of common input for which this approach would not work well?


florian-huber (approved):
Nice @niekdejonge ! That solves an issue that I had only marked as an inline comment in the filters list (which you now correctly removed).

PR #524, opened 2023-09-06
Remove duplicated filters

niekdejonge:
Duplicated filters will now not be added twice to the pipeline, instead they are added only once. And the settings of the last added filter are used. This would be expected behaviour for me, since sometimes you might want to use the default filters, but use different settings. This makes it possible to just use the default pipeline and only change the parameters for the filter that has to be changed.


florian-huber (approved):
Looks good to me. Also agree with you @niekdejonge that this is what I would expect as behavior from the pipeline.

PR #525, opened 2023-09-06
Complete library cleaning pipeline

niekdejonge:
After a second round of testing pipeline on an entire library a few small changes were made:

- Added filters that were missing from default pipelines.
- Also added a small function that prints filter functions available that are currently not used in a yaml file. This makes it easy to find filters that you might have forgotten in your pipeline. 
- Only check if the 2D structures match between smiles and inchi instead of 3D structure. 
- Change warning to logging for changed sign of given charge.
- Change default minimum number of peaks in reduce to number of peaks, since this is unexpected behaviour to remove spectra by default. The removal of spectra with too little peaks should be done by the filter made for that.


matchms/filtering/filter_order_and_default_pipelines.py
```diff
@@ -75,12 +76,27 @@
...
+                         ("reduce_to_number_of_peaks", {"n_max": 1000}),
+                         ("require_minimum_of_high_peaks", {"no_peaks": 5, "intensity_percent": 2.0}),
+                         ]
+
 PREDEFINED_PIPELINES = {
     "minimal": MINIMAL_FILTERS,
     "basic": BASIC_FILTERS,
     "default": DEFAULT_FILTERS,
     "fully_annotated": FULLY_ANNOTATED_PROCESSING,
+    "ms2deepscore": MS2DEEPSCORE_TRAINING,
```
florian-huber, line 101: I am fine with the addition of `ms2deepscore`. But that also reminded me that we could also reconsider the other ones. I assembled them fairly quickly when introducing the SpectrumProcessor, and then you made some edits. But let's better discuss this elsewhere --> #528

florian-huber (approved):

PR #526, opened 2023-09-07
Fix adduct missing charge

niekdejonge:
Still many adducts are not recognized since they do not have a charge e.g. [M+H+Na] would not be used, since the charge is unknown. Before some of these were repaired, because we stored common adduct types. In addition now the charge stored in spectrum.get("charge") is used to repair an adduct that misses the overall charge.

### DIFF
--- a/matchms/filtering/metadata_processing/derive_ionmode.py
+++ b/matchms/filtering/metadata_processing/derive_ionmode.py
@@ -30,8 +30,6 @@ def derive_ionmode(spectrum_in: SpectrumType) -> SpectrumType:
     spectrum = spectrum_in.clone()
 
     ionmode = spectrum.get("ionmode")
-    if ionmode:
-        assert ionmode == ionmode.lower(), "Ionmode field not harmonized."
     if ionmode in ["positive", "negative"]:
         return spectrum
 
--- a/matchms/filtering/metadata_processing/derive_ionmode.py
+++ b/matchms/filtering/metadata_processing/derive_ionmode.py
@@ -44,7 +44,7 @@ def derive_ionmode(spectrum_in: SpectrumType) -> SpectrumType:
     if adduct in list(known_adducts["adduct"]):
         ionmode = known_adducts.loc[known_adducts["adduct"] == adduct, "ionmode"].values[0]
     else:
-        ionmode = "n/a"
+        ionmode = None
 
     spectrum.set("ionmode", ionmode)
     logger.info("Set ionmode to %s.", ionmode)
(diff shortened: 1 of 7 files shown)


matchms/filtering/metadata_processing/derive_ionmode.py
```diff
@@ -29,20 +29,20 @@ def derive_ionmode(spectrum_in: SpectrumType) -> SpectrumType:
...
-
-    adduct = spectrum.get("adduct", None)
-    # Harmonize adduct string
-    if adduct:
-        adduct = _clean_adduct(adduct)
-
     ionmode = spectrum.get("ionmode")
     if ionmode:
         assert ionmode == ionmode.lower(), ("Ionmode field not harmonized.",
                                             "Apply 'make_ionmode_lowercase' filter first.")
```
florian-huber:
```suggestion
        assert ionmode == ionmode.lower(), "Ionmode field not harmonized."
```

florian-huber: `make_ionmode_lowercase` no longer exists since this is now part of the `Metadata` class.

niekdejonge: I guess we can remove this assert statement altogether in this case.

matchms/filtering/metadata_processing/derive_ionmode.py
```diff
@@ -30,8 +30,6 @@ def derive_ionmode(spectrum_in: SpectrumType) -> SpectrumType:
     spectrum = spectrum_in.clone()
 
     ionmode = spectrum.get("ionmode")
-    if ionmode:
-        assert ionmode == ionmode.lower(), "Ionmode field not harmonized."
```
florian-huber: You are right, can be removed entirely.

matchms/filtering/metadata_processing/derive_ionmode.py
```diff
@@ -51,5 +48,4 @@ def derive_ionmode(spectrum_in: SpectrumType) -> SpectrumType:
```
florian-huber, line 48: I think ionmode should be set to `None` instead of `"n/a"`.

florian-huber: (we haven't touched this filter in a while it seems...)

florian-huber (approved):
Looks good. `derive_ionmode` has long not been touched it seems. I would say we should remove missing entries, while this currently still sets ionmode to "n/a".

But OK, data without known ionmode is still not a problem that we should spend too much on in `matchms` 😄

PR #527, opened 2023-09-08
Fix processing report

niekdejonge:
Two small fixes of old PR's and improved the printing of the ProcessingReport (pandas has a default of max 2 columns when printing)
Hmm, there seems to be something weird going on with the commits. It seems like it is showing the commits from the previous commits as well. This might have to do with me doing a gitpull, resulting in a rebase for this branch. The actual changes were only a few, so most changes have already been merged. I do not really understand why this is happening, so I hope everything went fine in the end and we don't lose anything. @florian-huber


niekdejonge:
Something weird happened, so I just created a new branch that actually matches the master branch. #529

PR #531, opened 2023-09-12
Convert filtering functions to classes

zargham-ahmad:
As per our discussion, Opening the draft PR for discussion. 

This implementation is based on the template design pattern, containing a base class that contains two functions process() and abstract method apply_filter(). Then it has further sub base class for filters that share quite a lot of similar code e.g. select_by filter and harmonize_undefined filters. There is not really a change in the current filter function usage to keep usage the same for now, but internally they now use class implementation.

### DIFF
--- a/matchms/filtering/filters/add_compound_name.py
+++ b/matchms/filtering/filters/add_compound_name.py
@@ -1,0 +1,21 @@
+import logging
+from matchms.filtering.filters.base_spectrum_filter import BaseSpectrumFilter
+from matchms.typing import SpectrumType
+
+
+logger = logging.getLogger("matchms")
+
+class AddCompoundName(BaseSpectrumFilter):
+    def apply_filter(self, spectrum: SpectrumType) -> SpectrumType:
+        if spectrum.get("compound_name", None) is None:
+            if isinstance(spectrum.get("name", None), str):
+                spectrum.set("compound_name", spectrum.get("name"))
+                return spectrum
+
+            if isinstance(spectrum.get("title", None), str):
+                spectrum.set("compound_name", spectrum.get("title"))
+                return spectrum
+
+            logger.info("No compound name found in metadata.")
+
+        return spectrum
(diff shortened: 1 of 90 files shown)


niekdejonge:
@zargham-ahmad I hoped to give a code review before my holidays, but I became ill and won't be able to check it before my holidays anymore. What we saw during our meeting looked good! I will pick this up after my holidays again.

matchms/filtering/filters/add_compound_name.py
```diff
@@ -0,0 +1,21 @@
+import logging
+from matchms.filtering.filters.base_spectrum_filter import BaseSpectrumFilter
+from matchms.typing import SpectrumType
+
+
+logger = logging.getLogger("matchms")
+
+class AddCompoundName(BaseSpectrumFilter):
```
niekdejonge, line 8: It might make more sense, to have the Classes in the same file that has the functional equivalent.

niekdejonge: So below the add_compound_name function

niekdejonge: Something that is hard to judge now is if the same functionality is still in the classes that was previously in the function, since we made quite some updates to matchms in the last few weeks. There is a risk that these changes are undone. I am not sure how we could best check this. Anyone any suggestions?

matchms/filtering/filters/harmonize_undefined_template.py
```diff
@@ -0,0 +1,22 @@
+from typing import List
+from matchms.filtering.filters.base_spectrum_filter import BaseSpectrumFilter
+from matchms.typing import SpectrumType
+
+
+class HarmonizeUndefinedTemplate(BaseSpectrumFilter):
```
niekdejonge, line 6: Maybe good for file organizations to make a separate folder with these kinds of parent classes. The folder could be called FilterTypes for instance.

florian-huber:
We might at some point come back to this approach. The current implementation, however, took a different route (see `development` branch)

PR #535, opened 2023-10-09
Make initialize_spectrum_processor part of SpectrumProcessor init

niekdejonge:
I noticed we almost always used initialize_spectrum_processor to initialize SpectrumProcessor. This is now part of SpectrumProcessor init instead.


florian-huber (approved):
Looks good!

PR #536, opened 2023-10-09
Store cleaning settings

niekdejonge:
Currently the cleaning settings are not stored in the log file. 
Also we have a different way of generating a yaml file and generating the string format of cleaning. I will try to combine this into the same funciton in this PR

### DIFF
--- a/matchms/filtering/SpectrumProcessor.py
+++ b/matchms/filtering/SpectrumProcessor.py
@@ -1,6 +1,6 @@
 import inspect
 import logging
-from collections import defaultdict, OrderedDict
+from collections import OrderedDict, defaultdict
 from functools import partial
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 import numpy as np
--- a/matchms/filtering/SpectrumProcessor.py
+++ b/matchms/filtering/SpectrumProcessor.py
@@ -297,13 +297,11 @@ class ProcessingReport:
     def __str__(self):
         pd.set_option('display.max_columns', 4)
         pd.set_option('display.width', 1000)
-        report_str = f"""\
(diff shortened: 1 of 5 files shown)


matchms/filtering/SpectrumProcessor.py
```diff
@@ -305,12 +298,12 @@ def __str__(self):
...
-Number of spectrums removed: {sum(self.counter_removed_spectrums.values())}
-Changes during processing:
-{str(self.to_dataframe())}
-"""
+            ----- Spectrum Processing Report -----
+            Number of spectrums processed: {self.counter_number_processed}
+            Number of spectrums removed: {sum(self.counter_removed_spectrums.values())}
+            Changes during processing:
+            {str(self.to_dataframe())}
+            """
```
florian-huber: This didn't work for me and will create a string with many tabs in front of each line.

florian-huber: If you want to avoid having the string content without indentations a better solution would be
```
def test():
    lines = [
        "abc",
        "def"
    ]
    return "\n".join(lines)
```

or (my preference):
```
def test():
    my_str = (
        "abc\n"
        "def\n"
    )
    return my_str
```

florian-huber:
Looks good @niekdejonge.
Only the report string part needs to be changed.

matchms/filtering/SpectrumProcessor.py
niekdejonge: Ah good catch! Now I understand why it was weirdly formatted before...

niekdejonge: And the second solution is nice, did not know that that was possible in python

Issue #441, opened 2023-06-19
Implement `save_spectra` method


hechth:
Similar to `load_spectra` it might be nice to have a function to `save_spectra` in any format from a single function.

niekdejonge:
Will be implemented in #543

hechth:
This has been implemented by now.

PR #543, opened 2023-10-19
Prevent accidental file overwriting

### LINKED ISSUE #555: Load from mgf does not close the file after the generator is empty.
When creating a test for save_spectra I noticed that load_from_mgf does not correctly close the file after completely loading, while other loading methods do.

The test I wrote did run this. 
```python
spectrum_list = load_test_spectra_file()
    with tempfile.TemporaryDirectory() as temp_dir:
        filename = os.path.join(temp_dir, file_name)
        save_spectra(spectrum_list, filename)
        if not os.path.exists(filename):
            FileNotFoundError("No file was created")
        reloaded_spectra = list(load_spectra(filename))
```
And resulted in the error:
E           NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\jonge094\\AppData\\Local\\Temp\\tmpv9hy574j\\spectra.mgf'
The issue only occurs on windows and the problem is that the file cannot be removed before being closed, see https://python-forum.io/thread-33169.html 

This indicates that the file is still kept open after everything was loaded, which is not ideal. I will solve this by passing an opened file to the MGF class in #543

niekdejonge:
@hechth @florian-huber I agreed with your comments that the initial solution was not ideal. Therefore I reimplemented this in a better way. The saving now happens within spectrumprocessor, to make this possible I also added an save_spectra function, that saves as the filetype specified in the file name.
- [x] Add save_spectra, see #441 
- [x] Fix #555 (loading from mgf should use file object instead of file name)
- [x] Check if file exists before running pipeline
- [x] Save file auotmatically in pipeline
- [x] Add to changelog
- [x] Add tests
- [x] linting


matchms/Pipeline.py
```diff
@@ -57,6 +58,10 @@ def create_workflow(yaml_file_name: Optional[str] = None,
     workflow["reference_filters"] = reference_processor.processing_steps
     workflow["score_computations"] = score_computations
     if yaml_file_name is not None:
+        assert not os.path.exists(yaml_file_name), \
+            "This yaml file name already exists. " \
+            "To use the settings in the yaml file, please use the load_workflow_from_yaml_file function " \
+            "in yaml_file_functions.py or check the tutorial."
```
hechth, line 64: Ideally this would be covered by a test case

matchms/utils.py
```diff
@@ -73,3 +73,25 @@ def _load_key_conversions(file: str, source: str, target: str) -> dict:
...
+        new_file_name = f"{file_name_base}({i}){file_extension}"
+    print(f"Instead the file will be stored in {new_file_name}")
+    return new_file_name
+
+
+def create_dir_if_missing(folder):
+    """Checks if the folder already exists otherwise create the folder"""
+    if not os.path.isdir(folder):
+        assert not os.path.isfile(folder), f"{folder} is expected to be a folder but a file was found"
+        os.mkdir(folder)
```
hechth: The regular mkdir function already throws an exception if the folder exists - see https://docs.python.org/3/library/os.html#os.mkdir

matchms/utils.py
```diff
@@ -73,3 +73,25 @@ def _load_key_conversions(file: str, source: str, target: str) -> dict:
...
+        return file_name
+    print(f"The file name already exists: {file_name}")
+    file_name_base, file_extension = os.path.splitext(file_name)
+    i = 1
+    new_file_name = f"{file_name_base}({i}){file_extension}"
+    while os.path.exists(new_file_name):
+        i += 1
+        new_file_name = f"{file_name_base}({i}){file_extension}"
+    print(f"Instead the file will be stored in {new_file_name}")
+    return new_file_name
```
hechth: I find this highly problematic - either a user passes a valid path or should generate termporary filenames based on a timestamp or so ... but

matchms/utils.py
niekdejonge: My issue with os.mkdir, was actually that it throws an exception when the folder with that exact name already exists. When running the pipeline myself, I did often rerun previous run pipelines so sometimes the folder was already created and sometimes it was not yet there, making the code crash. 

But the assert statement to check if it is a file is not needed indeed. 

Not sure if it should be added to matchms, but, since I found it useful from a user perspective, I thought it might be more user friendly for other users as well.

matchms/utils.py
niekdejonge: The reason I added this, is that I sometimes rerun a pipeline and forget to remove the already existing file. It is really frustrating if you have a pipeline running overnight, finishing and not saving the file with the spectra, because it already existed. This always worked well for me as a quick fix.. But, using timestamps (or checking at the start) would of course work as well. 

Again my motivation was, since I find that it makes running more user friendly for me this probably is the case for other users as well and would therefore be good to add to matchms.

However, if this gives any issues for other pipelines, there is no need to add this to matchms of course and it is fine for me to remove it.

matchms/utils.py
hechth: I mean you can use try - catch to deal with the exception - an assert is actually going to do exactly the same, and also assert is very slow and only supposed to be used in testing and debugging, so I'd like to keep this out of running production code.

florian-huber:
```suggestion
    try:
        os.makedirs(folder, exist_ok=True)
    except FileExistsError:
        raise FileExistsError(f"{folder} is expected to be a folder but a file was found") from None
```

florian-huber: Would something like this work @niekdejonge ?

matchms/utils.py
florian-huber: I agree that this is not ideal but the pain Niek describes is probably even harder :)
Ideally, we still would like to avoid this I think and test before any extensive computation if file name issues are expected to arise.

If that is not easily possible I would suggest to keep such a solution for now, we make a clear issue that this should be avoided if possible as soon as we find better ways.
In that case, I would still change the function to make clearer (I hope) what it does. Maybe something like:
```python
def get_unique_filename(file_name: str):
    """
    Returns a unique filename, appending an index if the provided filename already exists.

    Parameters
    ------------
    file_name: Desired file name.

    Returns:
    - str: Unique file name.
    """
    if not os.path.exists(file_name):
        return file_name
    
    file_name_base, file_extension = os.path.splitext(file_name)
    i = 1

    while os.path.exists(f"{file_name_base}({i}){file_extension}"):
        i += 1
        
    return f"{file_name_base}({i}){file_extension}"
```

florian-huber: What happens if this function is used could be defined at the location where it is in applied. 

```python
original_name = "sample.txt"
new_name = get_unique_filename(original_name)

if original_name != new_name:
    print(f"The file name '{original_name}' already exists.")
    print(f"Instead, the file will be stored as '{new_name}'.")
    # or do some logging etc...
```

niekdejonge: That is a nice cleanup indeed Florian. But I do agree that this is a bit too hacky, therefore I will just use this only locally if I have this issue when rerunning many times (mostly during testing). 
A better solution is to just raise an error, before the pipeline starts running. This was not very straightforward to implement, since the saving now happens outside of pipeline. But we can easily move this inside the Pipeline class and add a check if the file actually exists. I will remove these functions and implement the check and saving in pipeline instead.

matchms/utils.py
niekdejonge: Oh yes that is perfect. I think the except statement is not even necessary in this case, so in that case we do not need this extra function and I will just replace it with os.mkdirs(folder, exist_ok=True).

tests/importing/test_load_from_mgf.py
```diff
@@ -11,14 +11,3 @@ def test_load_from_mgf_using_filepath():
 
     assert len(spectra) > 0
     assert isinstance(spectra[0], Spectrum)
-
-
-def test_load_from_mgf_using_file():
-    module_root = os.path.join(os.path.dirname(__file__), "..")
-    spectra_filepath = os.path.join(module_root, "testdata", "pesticides.mgf")
-
-    with open(spectra_filepath, "r", encoding="utf-8") as spectra_file:
```
niekdejonge, line 20: This is now the standard way of loading, so this test does not work anymore. See #555 Do you agree that this is expected behavior? @florian-huber @hechth

tests/test_pipeline.py
```diff
@@ -216,3 +218,14 @@ def test_pipeline_changing_workflow():
     expected = np.array([[1., 0.30384404],
                          [0.30384404, 1.]])
     assert np.allclose(all_scores["ModifiedCosine_score"][3:, 3:], expected)
+
+
+def test_save_spectra_spectrum_processor():
```
fabianegli: You could use the tmp_path fixture from pytest to simplify the setup. See https://docs.pytest.org/en/7.4.x/how-to/tmp_path.html

tests/importing/test_load_from_mgf.py
florian-huber: Yes, this should indeed be closed as you pointed out in #555.

matchms/exporting/save_spectra.py
```diff
@@ -0,0 +1,60 @@
+import os
+import pickle
+from typing import List
+from matchms.exporting import save_as_json, save_as_mgf, save_as_msp
+from matchms.Spectrum import Spectrum
+from matchms.typing import SpectrumType
+
+
+def save_spectra(spectrums: List[SpectrumType],
+                 file: str) -> None:
```
florian-huber: According to #557 the export_style should also be added as a parameter to this function.

matchms/exporting/save_spectra.py
```diff
@@ -0,0 +1,60 @@
...
+    Args:
+    -----
+    spectrums:
+        The spectra that are saved.
+    file:
+        Path to file containing spectra, with file extension "json", "mgf", "msp"
+    ftype:
+        Optional. Filetype
+    """
+    assert not os.path.exists(file), f"The specified file: {file} does already exist"
```
florian-huber:
```suggestion
    if os.path.exists(file):
        raise FileExistsError(f"The specified file: {file} already exists.")
```

tests/test_pipeline.py
florian-huber: Indeed. We still have several of those older "with tempfile... " constructions in our tests. Probably can simply be replaced by `tmp_dir` (tryout #566).

tests/filtering/test_spectrum_processor.py
```diff
@@ -261,3 +264,16 @@ def test_add_filter_twice():
     processor.add_filter(("derive_adduct_from_name", {"remove_adduct_from_name": False}))
     processor.add_filter("derive_adduct_from_name")
     assert processor.processing_steps == [("derive_adduct_from_name", {"remove_adduct_from_name": True})]
+
+
+def test_save_spectra_spectrum_processor(spectrums):
+    processor = SpectrumProcessor("default")
+    with tempfile.TemporaryDirectory() as temp_dir:
```
florian-huber: Same here --> switch to `tmp_path` (#566)

florian-huber (approved):
Nice additions @niekdejonge!
Added a few comments. The only notable change that would be nice to have before merging is to add `export_style` to the save function. Minor code change in the tests on top of this PR is in #566.

matchms/exporting/save_spectra.py
fabianegli: Assert should not be used in this way. One reason is seen in the suggested change which raises a more specific error. The second is that assert will not be run if Python is invoked with optimization. Biopython had the same issue https://github.com/biopython/biopython/issues/3600

niekdejonge: Thanks @fabianegli Before, I never really understood why assert should not be used in production code. Thanks for clarifying!

tests/test_pipeline.py
niekdejonge: Thanks. Not sure why I did not go for tmp_path this time...

PR #547, opened 2023-10-25
Enable setting key conversion with static method

### LINKED ISSUE #546: Enable setting key conversion with static method
Currently it is impossible to avoid the key conversion. It should be possible to set them to a user defined value with a static method.

hechth:
This allows setting a custom key conversion and changes the `_key_replacements` and `_key_regex` variables to be class variables from global.

This allows removing all key conversion if desired to allows reading and storing files in their original way if desired.

Fixes #546

### DIFF
--- a/matchms/Metadata.py
+++ b/matchms/Metadata.py
@@ -11,11 +11,6 @@ from .filtering.metadata_processing.make_charge_int import \
 from .utils import load_export_key_conversions, load_known_key_conversions
 
 
-_key_regex_replacements = {r"\s": "_",
-                           r"[!?.,;:]": ""}
-_key_replacements = load_known_key_conversions()
-
-
 class Metadata:
     """Class to handle spectrum metadata in matchms.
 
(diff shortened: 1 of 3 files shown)


zargham-ahmad (approved):
Looks good to me!

matchms/Metadata.py
```diff
@@ -195,3 +195,14 @@ def data(self, new_dict):
                 self.harmonize_keys()
         else:
             raise TypeError("Expected input of type dict or PickyDict.")
+    
+    @staticmethod
```
niekdejonge, line 199: What is the expected use case? Is is to remove all the known_key_conversions, or to replace/add just a few? In the last case, it might be better to replace set_key_replacements with change_key_replacements, which should add the given keys to the known_key_conversions , or will overwrite if it was already given in the known_key_conversions.

niekdejonge (approved):
Looks good, one comment with a suggestion, but feel free to implement in the way you see fit.

florian-huber (approved):
Looks fine to me. Within the class make this changable from outside which could become handy. Thanks for the PR!

matchms/Metadata.py
hechth: The use case is kind of flexible - can be disabling the key conversions or just setting it to a user defined value - since it is such a minimal and non-invasive change I didn't give that much thought to be fair.

PR #552, opened 2023-11-01
Update require_parent_mass_match_smiles.py

niekdejonge:
Fixes an issue when parent mass is None


florian-huber (approved):
Looks good to me!

Issue #551, opened 2023-10-31
WARNINGs and KeyError when `load_from_msp` MoNA LC-MS/MS Positive Mode MSP


kozo2:
When I execute the code below,
```
# Get MoNA LC-MS/MS Positive Mode MSP zip file
wget https://mona.fiehnlab.ucdavis.edu/rest/downloads/retrieve/873fbe29-4808-46d1-a4a3-a4134ac8c755 
mv 873fbe29-4808-46d1-a4a3-a4134ac8c755 mona_msp.zip
unzip mona_msp.zip
```
```
from matchms.importing import load_from_msp

# Download msp file from MassBank of North America repository at https://mona.fiehnlab.ucdavis.edu/
file_msp = "MoNA-export-LC-MS-MS_Positive_Mode.msp"
tmp = load_from_msp(file_msp)
foo = list(tmp)
```

I get the following warnings and errors.
```
...
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
2023-10-31 08:59:20,349:WARNING:matchms:add_retention:0.58-0.89 min can't be converted to float.
WARNING:matchms:0.58-0.89 min can't be converted to float.
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
2023-10-31 08:59:20,357:WARNING:matchms:add_retention:0.58-0.89 min can't be converted to float.
WARNING:matchms:0.58-0.89 min can't be converted to float.
WARNING:pickydict:Key 'exact_mass' will be interpreted as 'parent_mass'. But this entry already exists. Please use 'parent_mass' if you want to replace the entry.
[...]

PR #553, opened 2023-11-09
Fix retention time for msp

niekdejonge:
Solves #551 

The issue was that the regex would search for patterns like:

100 s
100 min

But it would use a split function later to find these cases. This resulted in 100 sec, which matches 100 s so passed the regex test, but sec would be used instead.

Fixes:

- [x] Change regex to require end of line as well.
- [x] Added sec as option
- [x] Renaming of some unclear variables
- [x] Added test
- [x] In metadata harmonization all common keys are used, instead of only "retention_time"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [unreleased]
 ### Added
 - added option to set custom key replacements [#547](https://github.com/matchms/matchms/pull/547)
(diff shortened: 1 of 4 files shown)


niekdejonge:
@florian-huber and @hechth Do you agree that during metadata harmonization that we should use all common retention_time_keys: ["retention_time", "retentiontime", "rt", "scan_start_time", "rt_query", "rtinseconds"] instead of only using "retention_time"?

florian-huber (approved):
Looks good for me. But I never worked with all those retention time namings. Maybe @hechth can have a quick look, too.

hechth (approved):

PR #554, opened 2023-11-10
Fix issue with parse metadata recognizing empty metadata as well.

### LINKED ISSUE #548: Metadata processing of some msp files fails fatally
**Describe the bug**
When using some MoNA MassBank msp files, the metadata extraction fails on line 157 of load_from_msp.py. 

**To Reproduce**
Attempt to load the MSP file here: https://mona.fiehnlab.ucdavis.edu/downloads, specifically the LC-MS/MS positive mode dataset. 

**Expected behavior**
This is probably due to one or more files not complying with the metadata format specified or expected by the parser. This in and of itself is not a problem since non-compliant files are not matchms problem per se; however, I would expect the inability to parse a metadata field to be handled more gracefully. Currently when this issue occurs, the entire extraction fails. On my system I have worked around this by modifying the function and wrapping lines 155-157 in a try catch. 

 - OS: MacOS
 - Browser: N/A
 - Version 0.22.0

**Additional context**
I'd be happy to submit a pull request or to make some tweaks to the metadata parsing to make it more robust to these sorts of issues. I am only working with MoNA currently so I'm not sure how widespread the problem could be. I do have metadata harmonization disabled in the call to load_from_msp.

niekdejonge:
Fixes #548 

This was a good lesson for me in the complexity of different metadata formats accepted in msp ...

What caused #548 is that cases like 'Comments: "compound class=" 'Comments: "compound class=" 'will be recognized by the regex, but will return '' for the seconds group, which makes match[1] not work. 
Simply adding if len(match) != 2 solves this issue.

The function was quite complex, so I wanted to simplify this at first. But I have now figured out why this was not possible, since it would not recognize all the possible formats anymore. I have added tests for these cases and added documentation to the function to clarify the reason for the steps.

### DIFF
--- a/matchms/importing/load_from_msp.py
+++ b/matchms/importing/load_from_msp.py
@@ -154,6 +154,9 @@ def parse_metadata(rline: str, params: dict):
         # The first checks for compound="caffeine", the second and third "compound=caffeine and the last for parent_mass=12.1
         # The second and third almost match the same pattern, but the second is there to ensure tha the pattern
         # "SMILES=CC(O)C(O)=O" is recognized as 'smiles': 'CC(O)C(O)=O' instead of 'smiles=cc(o)c(o)': 'O'
+        # Cases like "Smiles=" will not be stores (since len(match) = 1) The reason that we did not change the * to +
+        # in the regex, is that we want to still match to these cases, so that the if len(matches) == 0: below is not
+        # called in cases like comments: "smiles=".
         pattern = r'(\S+)="([^"]*)"|' \
                   r'"(\w+)=([^"]*)"|' \
                   r'"([^"]*)=([^"]*)"|' \
--- a/matchms/importing/load_from_msp.py
+++ b/matchms/importing/load_from_msp.py
@@ -172,7 +172,7 @@ def parse_metadata(rline: str, params: dict):
                     params[key.lower().strip()] = value.strip()
     # msp files can have the format comments: smiles="CC=O" but also the format smiles: CC=O.
     # The latter is captured by these lines.
-    if len(matches) == 0:
+    if len(matches) == 0 and splitted_line[0] != "comments":
         params[splitted_line[0].lower()] = splitted_line[1].strip()
 
 
--- a/matchms/importing/load_from_msp.py
+++ b/matchms/importing/load_from_msp.py
@@ -154,6 +154,7 @@ def parse_metadata(rline: str, params: dict):
         # The first checks for compound="caffeine", the second and third "compound=caffeine and the last for parent_mass=12.1
         # The second and third almost match the same pattern, but the second is there to ensure tha the pattern
         # "SMILES=CC(O)C(O)=O" is recognized as 'smiles': 'CC(O)C(O)=O' instead of 'smiles=cc(o)c(o)': 'O'
+        # The third pattern is needed since some keys like DB# should be stored, which is not recognized by the \w regex
         # Cases like "Smiles=" will not be stores (since len(match) = 1) The reason that we did not change the * to +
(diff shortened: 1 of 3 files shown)


matchms/importing/load_from_msp.py
```diff
@@ -149,16 +147,20 @@ def parse_metadata(rline: str, params: dict):
...
-                params[key.lower().strip()] = value.strip()
+            if len(match) == 2:
+                key = match[0]
+                value = match[1]
+                if key.lower().strip() in params.keys() and key.lower().strip() == 'smiles':
+                    params[key.lower()+"_2"] = value.strip()
+                else:
+                    params[key.lower().strip()] = value.strip()
     if len(matches) == 0:
         params[splitted_line[0].lower()] = splitted_line[1].strip()
```
niekdejonge, line 166: I think the len(matches) == 0 part, ensures that lines that contain something of the form compound:some name, is stored. This seems weir to me. But it might be something we want. To me it would make sense to remove this. What do you think?

matchms/importing/load_from_msp.py
```diff
@@ -149,16 +147,20 @@ def parse_metadata(rline: str, params: dict):
     matches = []
```
niekdejonge, line 150: Params is not returned, but edited in the function and used outside the function again. This might slightly improve speed, but I think this does not outweight the downside.

Suggestion:
Just return params and use the output in other functions.

matchms/importing/load_from_msp.py
```diff
@@ -149,16 +147,20 @@ def parse_metadata(rline: str, params: dict):
...
     splitted_line = rline.split(":", 1)
     if splitted_line[0].lower() == 'comments' and "=" in splitted_line[1]:
-        pattern = r'(\S+)="([^"]*)"|"(\w+)=([^"]*)"|"([^"]*)=([^"]*)"|(\S+)=(\d+(?:\.\d*)?)'
+        pattern = r'(\S+)="([^"]+)"|' \
```
niekdejonge, line 150: I have changed this to require a symbol in both groups (before and after =) before cases like. "compound_name=" would make the code crash, since the second part is None. But we might want to store this as compound_name: None in the spectrum object. This long regex seems a bit hacky and I think it can be simplified, but I am afraid to remove edge cases that cannot be detected in that case anymore, what do you think?

matchms/importing/load_from_msp.py
```diff
@@ -149,16 +147,20 @@ def parse_metadata(rline: str, params: dict):
...
+                  r'"(\w+)=([^"]+)"|' \
+                  r'"([^"]+)=([^"]+)"|' \
+                  r'(\S+)=(\d+(?:\.\d*)?)'
         matches = re.findall(pattern, splitted_line[1].replace("'", '"'))
         for match in matches:
             match = [i for i in match if i]
```
niekdejonge, line 156: This regex will return a tuple with 8 matches, of which 6 will be empty ''. Only the non empty once are stored. This was what made the code break at first, since cases like "compound_name=" Would have an empty value for 7 tuples, resulting in match[1] not existing. The reason this regex is so complex, is that it tries to cover different formats, to be specific different usages of "". It can be: "compound_name=caffeine" compound_name="caffeine", and parent_mass=12.3 I found this structure a bit hacky, but did not really know a quick fix, so just added the check that len(match) ==2

niekdejonge: I think the pattern "(\w+)=([^"]+)" is always captured by "([^"]+)=([^"]+)" as well. So I did remove this, since this does not add anything.

hechth (approved):

Issue #556, opened 2023-11-14
Add export style to all export formats


florian-huber:
Currently only `save_as_msp()` uses a `style` parameter to choose the style of metadata field names (matchms, massbank, gnps ...).
This should also be implemented for
- `save_as_mgf` 
- `save_as_json`

florian-huber:
Follows from #537

florian-huber:
Will be fixed with #557

PR #557, opened 2023-11-14
add export_style parameter

florian-huber:
Work on #556.

- [x] `save_as_mgf` supports `export_style` parameter
- [x] tests for `save_as_mgf` using other than the default `export_style` parameter (e.g. `gnps`).
- [x] `save_as_json` supports `export_style` parameter
- [x] tests for `save_as_jsonf` using other than the default `export_style` parameter (e.g. `gnps`).

The exporting module could need some refactoring. The focus of this PR is mostly to have a functioning export style option in all export formats.

### DIFF
--- a/matchms/Spectrum.py
+++ b/matchms/Spectrum.py
@@ -187,17 +187,31 @@ class Spectrum:
         self._metadata.set(key, value)
         return self
 
-    def to_dict(self) -> dict:
-        """Return a dictionary representation of a spectrum."""
+    def to_dict(self, export_style: str = "matchms") -> dict:
+        """Return a dictionary representation of a spectrum.
+
+        Parameters
+        ----------
+        export_style:
+            Converts the keys to the required export style. One of ["matchms", "massbank", "nist", "riken", "gnps"].
+            Default is "matchms"
+        """
         peaks_list = np.vstack((self.peaks.mz, self.peaks.intensities)).T.tolist()
-        spectrum_dict = dict(self.metadata.items())
+        spectrum_dict = self.metadata_dict(export_style)  # dict(self.metadata.items())
         spectrum_dict["peaks_json"] = peaks_list
         if "fingerprint" in spectrum_dict:
             spectrum_dict["fingerprint"] = spectrum_dict["fingerprint"].tolist()
         return spectrum_dict
 
     def metadata_dict(self, export_style: str = "matchms") -> dict:
-        """Convert spectrum metadata to Python dictionary."""
+        """Convert spectrum metadata to Python dictionary.
+
+        Parameters
+        ----------
+        export_style:
+            Converts the keys to the required export style. One of ["matchms", "massbank", "nist", "riken", "gnps"].
+            Default is "matchms"
+        """
         return self._metadata.to_dict(export_style)
 
     @property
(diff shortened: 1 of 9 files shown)


matchms/exporting/save_as_json.py
```diff
@@ -42,25 +47,27 @@ def save_as_json(spectrums: List[Spectrum], filename: str):
...
+            """JSON Encoder for a matchms.Spectrum.Spectrum object"""
+            if isinstance(o, Spectrum):
+                spec = o.clone().to_dict(export_style)
+                spec.pop("fingerprint", None)
+                return spec
+            return super().default(o)
+    return CustomSpectrumJSONEncoder
 
 
 class ScoresJSONEncoder(json.JSONEncoder):
```
niekdejonge: Shouldn't this be moved to the Scores class, since it is only used there? Now my first impression was that this was part of the saving a spectrum as JSON, while it is only used when saving the scores as JSON.

florian-huber: I think that this is so far only used in the saving spectra as JSON part. Unless it is needed elsewhere I think we don't have to move it into Spectrum.

matchms/exporting/save_as_json.py
```diff
@@ -42,25 +47,27 @@ def save_as_json(spectrums: List[Spectrum], filename: str):
     fingerprint_export_warning(spectrums)
 
     # Write to json file
+    encoder_class = create_spectrum_json_encoder(export_style)
     with open(filename, "w", encoding="utf-8") as fout:
-        json.dump(spectrums, fout, cls=SpectrumJSONEncoder)
+        json.dump(spectrums, fout, cls=encoder_class)
```
niekdejonge, line 53: Looks good!

matchms/exporting/save_as_json.py
florian-huber: Ah, OK, sorry. I also overlooked this.
The `ScoresJSONEncoder` is indeed used in Scores as well. I am fine either way: we can keep it in the save_as_json part or move this one into Scores. The `create_spectrum_json_encoder` should stay in the save_as_json I believe.

florian-huber: I think we did this at the time to avoid some circular imports... but that might not be an issue anymore.

matchms/Spectrum.py
```diff
@@ -187,17 +187,31 @@ def set(self, key: str, value):
         self._metadata.set(key, value)
         return self
 
-    def to_dict(self) -> dict:
-        """Return a dictionary representation of a spectrum."""
+    def to_dict(self, export_style: str = "matchms") -> dict:
```
niekdejonge, line 190: This is also used in places like scores, for exporting spectra. This works the same now, since the default is set to "matchms". However, we might want to add the export_style function also to Scores.to_dict()

niekdejonge (approved):
I left some small comments, but looks good to me!

matchms/exporting/save_as_json.py
florian-huber: I now moved it to `Scores`.

niekdejonge: I think the ScoresJSONEncoder is only used in Scores, or is it also used in save_as_json? But the circular imports could be an issue indeed. 

Yes, the create_spectrum_json_encoder should certainly stay in the save_as_json

florian-huber:
Thanks a lot @niekdejonge !

matchms/Spectrum.py
florian-huber: Good point. And should be easy to do.

florian-huber: I added a test for the `to_dict()` method. But the export style change is more complex than I thought. I believe, because of the automatically triggered harmonization when using Scores... anyway that will be a new issue.

PR #558, opened 2023-11-14
Fix repair not matching annotation

niekdejonge:
Before this filter would still repair the smiles from the inchi, if both would actually be incorrect (not matching parent mass).


hechth (commented):
Would be great if there was a test case covering this!

PR #559, opened 2023-11-14
Fix wrong pubchem annotations

niekdejonge:
Some annotations on pubchem are not interpretable by rdkit. This resulted in them being deleted by require_valid_annotation later. 

The following changes were made to fix this:
- Only add smiles/inchi if they are considered valid by rdkit. 
- Run the derive_smiles_from_inchi, derive_inchi_from_smiles and derive_inchikey_from_inchi filters after deriving the annotation from pubchem. This ensures that any of the none valid annotations are derived from a valid annotation after.
- Run the filter only if there is no valid smiles or valid inchi. Before we checked for complete annotation, but we only want to run it if there is no other way to derive the annotation from smiles/inchi. 

Also:
- The name was changed to derive_annotation_from_compound_name


matchms/filtering/metadata_processing/derive_annotation_from_compound_name.py
```diff
@@ -10,16 +10,16 @@
...
 
-def derive_smiles_from_pubchem_compound_name_search(spectrum_in: Spectrum,
-                                                    annotated_compound_names_file: Optional[str] = None,
-                                                    mass_tolerance: float = 0.1):
+def derive_annotation_from_compound_name(spectrum_in: Spectrum,
+                                         annotated_compound_names_file: Optional[str] = None,
+                                         mass_tolerance: float = 0.1):
     """Adds smiles, inchi, inchikey based on compound name by searching pubchem
 
     The smiles, inchi and inchikey are repaired if the found smiles is close enough to the parent mass.
```
florian-huber: "repaired" is not how I would call this. Better describe in a bit more detail what is done by the function (including the new change that this function will only be executed if there is no valid annotation.

matchms/filtering/metadata_processing/derive_annotation_from_compound_name.py
```diff
@@ -40,13 +40,13 @@ def derive_smiles_from_pubchem_compound_name_search(spectrum_in: Spectrum,
     if spectrum_in is None:
         return None
     spectrum = spectrum_in.clone()
-    if _check_fully_annotated(spectrum):
+    # Only run this function if it does not yet have a useful annotation
+    if is_valid_inchi(spectrum.get("inchi")) or is_valid_smiles(spectrum.get("smiles")):
         return spectrum
```
florian-huber, line 46: Maybe add a logger.info after those lines, saying that no valid annotation was found and that a pubchem search will be carried out?

matchms/filtering/metadata_processing/derive_annotation_from_compound_name.py
```diff
@@ -55,10 +55,15 @@ def derive_smiles_from_pubchem_compound_name_search(spectrum_in: Spectrum,
             if within_mass_tolerance.shape[0] > 0:
                 # Select the match with the most
```
florian-huber: Incomplete comment.

florian-huber (approved):
Looks good @niekdejonge and the changes in strategy make sense in my view.
Only thing is maybe that there might be cases when people want to trust PubChem more than the current metadata, but probably this could be handled better in a new function. Maybe a function that compares the search results for compound name with the search results for inchi/smiles to check for correctness (but let's keep this for another time...).

matchms/filtering/metadata_processing/derive_annotation_from_compound_name.py
niekdejonge: Good suggestion. I added it at the end of the function, which will only run if it did not find a good match. So we can actually log that pubchem lookup was run but not successfully.

matchms/filtering/metadata_processing/derive_annotation_from_compound_name.py
niekdejonge: This made me realize that I actually did the wrong thing in the line below...

Before I selected the lowest parent mass, instead of the smallest mass difference. In reality this would never give an issue, since we only select cases that are within the smallest mass difference. But it has now been fixed and a test has been added.

matchms/filtering/metadata_processing/derive_annotation_from_compound_name.py
niekdejonge: Good point, this has been fixed

niekdejonge:
> Looks good @niekdejonge and the changes in strategy make sense in my view. Only thing is maybe that there might be cases when people want to trust PubChem more than the current metadata, but probably this could be handled better in a new function. Maybe a function that compares the search results for compound name with the search results for inchi/smiles to check for correctness (but let's keep this for another time...).

Good suggestion. I will create an issue.

hechth:
SMILES handling and creation is notoriously wrong and there are many inconsistencies between software packages. RDKit doesn't take some SMILES from Pubchem, CDK doesn't take some from RDKit but is happy with Pubchem - I think inchified SMILES would be the way forward, as they are canonicalized and unique. Though I think support is overall lacking.

Issue #561, opened 2023-11-14
Remove predefined pipelines from SpectrumProcessor


niekdejonge:
Currently we can separately add predefined pipelines and additional filters.

```python 
class SpectrumProcessor:
    def __init__(self, predefined_pipeline: Optional[str] = 'default',
                 additional_filters: Iterable[Union[str, List[dict]]] = ()):
        self.filters = []
        self.filter_order = [x.__name__ for x in ALL_FILTERS]
        if predefined_pipeline is not None:
            if not isinstance(predefined_pipeline, str):
                raise ValueError("Predefined pipeline parameter should be a string")
            if predefined_pipeline not in PREDEFINED_PIPELINES:
                raise ValueError(f"Unknown processing pipeline '{predefined_pipeline}'. Available pipelines: {list(PREDEFINED_PIPELINES.keys())}")
            for filter_name in PREDEFINED_PIPELINES[predefined_pipeline]:
                self.add_matchms_filter(filter_name)
        for filter_name in additional_filters:
            if isinstance(filter_name, (tuple, list)) and len(filter_name) == 1:
                filter_name = filter_name[0]
            self.add_filter(filter_name)
```

Proposed solution:
Still allow for using the predefined pipelines, but instead of storing them in PREDEFINED_PIPELINES, just use the predefined lists of filters. 
so this would look like:

```python 

from filter_order_and_defaul_pipelines import DEFAULT

class SpectrumProcessor:
    def __init__(self, filters: Iterable[Union[str, List[dict]]] = ()):
        self.filters = []
        self.filter_order = [x.__name__ for x in ALL_FILTERS]
        for filter_name in filters:
            if isinstance(filter_name, (tuple, list)) and len(filter_name) == 1:
                filter_name = filter_name[0]
            self.add_filter(filter_name)

SpectrumProcessor(DEFAULT + ["normalize_intensities])
```
I think that would be a lot cleaner and more straightforward.

niekdejonge:
@florian-huber what do you think?

PR #565, opened 2023-11-15
Remove predefined pipelines

niekdejonge:
Started of with implementing #561, but noticed that SpectrumProcessor had a confusing structure for adding filters and a lot of code repetition. This improves this and also makes the processing of matchms filters or custom filters more consistent. (I probably should not have spent my time on this, since I have a paper to finish... )

- [x] Implement #561 (Remove predefined pipelines as input from SpectrumProcessor)
- [x] Add tests that all the pipelines can actually run
- [x] Refactore SpectrumProcessor to make more Modular and less code repetition
- [x] Allow matchms filters to be passed as filter functions and to order them in a different way than the default order. 
- [x] The default pipelines now store the matchms functions instead of string representation (this makes sure name changes are automatically implemented)


matchms/filtering/SpectrumProcessor.py
```diff
@@ -24,116 +23,106 @@ class SpectrumProcessor:
...
-        filter_spec : str or tuple
-            Name of the filter function to add, or a tuple where the first element is the name of the
-            filter function and the second element is a dictionary containing additional arguments for the function.
+        for filter_name in filters:
+            self.parse_and_add_filter(filter_name)
+
+    def parse_and_add_filter(self, filter_description: Union[Tuple[str, Dict[str, any]],
+                                                             str,
+                                                             Tuple[Callable, Dict[str, any]],
+                                                             Callable],
```
hechth: Maybe it would be worth to look into typedefs for these things: https://stackoverflow.com/questions/69446189/python-equivalent-for-typedef

matchms/filtering/SpectrumProcessor.py
```diff
@@ -24,116 +23,106 @@ class SpectrumProcessor:
...
+        filter_args = None
+        if isinstance(filter_description, (tuple, list)):
+            if len(filter_description) == 1:
+                filter_function = filter_description[0]
+            elif len(filter_description) == 2:
+                filter_function = filter_description[0]
+                filter_args = filter_description[1]
+            else:
+                raise ValueError("The filter_function should contain only two values, "
+                                 "the first should be string or callable and the second a dictionary with settings")
```
hechth, line 69: maybe using typedefs or even a class would then also make constructing these things a lot easier and would get rid of a lot of code.

matchms/filtering/SpectrumProcessor.py
```diff
@@ -24,116 +23,106 @@ class SpectrumProcessor:
...
-                               "check yaml file for details", new_filter_func)
-                self.filters[i] = new_filter_func
+                               "check yaml file for details", new_filter_function.__name__)
+                self.filters[i] = new_filter_function
                 filter_already_added = True
         if not filter_already_added:
-            self.filters.append(new_filter_func)
+            self.filters.append(new_filter_function)
         # Sort filters according to their order in self.filter_order
         self.filters.sort(key=lambda f: self.filter_order.index(f.__name__))
```
hechth: This whole section is a bit convoluted and hiding what it is actually doing. Putting all the checks into functions would make it easier to read.

hechth (approved):
Removing the string definitions is for sure a good thing. Parts of the code could still be segmented finer into functions to improve readability

matchms/filtering/SpectrumProcessor.py
niekdejonge: Ah good suggestion. I did not know/ think about typedefs before. Can certainly be helpfull. I have used it here and also combined the two two Tuple types into one.

matchms/filtering/SpectrumProcessor.py
niekdejonge: I like your suggestion of making a separate class for FilterFunction. However, this filter function would still need exactly the same code in the init. So in that sense not really solving the complexity. The reason we need this code is that we allow for multiple formats as input. But this is actually our design choice and I think we could simply change that. I have created an issue about this: #571

matchms/filtering/SpectrumProcessor.py
niekdejonge: Agreed. i have added a _replace_already_stored_filters function

florian-huber (approved):

PR #566, opened 2023-11-16
Switch to tmp_path

florian-huber:
... just to make it more readable...


niekdejonge:
Thanks Florian, for directly implementing the suggestion :) I quickly changed to using os.path.join() for consistency.

PR #567, opened 2023-11-17
Solve merge conflicts

niekdejonge:
There were quite some merge conflicts, that had to be solved manually. 
Also some of the new tests in main did not adhere to the new way of creating SpectrumProcessor objects, so that was updated here as well. 

(this also already includes #558)


hechth:
@niekdejonge if you have consecutive work, it makes sense to first indicate which PRs should be reviewed and merged first before others, as it makes the reviewing process easier.

niekdejonge:
@hechth Yes agreed. Sorry if that was unclear from the description.

PR #581, opened 2023-11-23
adjust logger message

florian-huber:


matchms/filtering/metadata_processing/repair_not_matching_annotation.py
```diff
@@ -84,9 +84,9 @@ def _check_repairing_is_possible(smiles, inchi, inchikey) -> bool:
     elif valid_smiles or valid_inchi or valid_inchikey:
         # At least one of the annotations, but some are not.
         # Since if valid_smiles and valid_inchi and valid_inchikey was False
-        logger.warning("Please first run repair_inchi_from_smiles, repair_smiles_from_inchi and repair_inchikey. "
-                       "The spectrum had partly valid annotations, "
-                       "this shows that these repair functions were not yet run.")
+        logger.warning("Please first run derive_inchi_from_smiles, derive_smiles_from_inchi and derive_inchikey. "
+                       "The spectrum has partly valid annotations, "
+                       "either these repair functions were not yet run or conversions between the formats failed.")
```
niekdejonge, line 89: Can the conversion between the formats fail? That would also explain the removal of some spectra after repairing. I would expect that every annotation that has at least one valid form of annotation, should be able to derive the inchi and inchikey.

matchms/filtering/metadata_processing/repair_not_matching_annotation.py
```diff
@@ -84,9 +84,9 @@ def _check_repairing_is_possible(smiles, inchi, inchikey) -> bool:
     elif valid_smiles or valid_inchi or valid_inchikey:
```
niekdejonge, line 84: Now I see this, the valid_inchikey is not enough, since we cannot derive smiles and inchi from inchikey alone. So I would only check elif valid_smiles or valid_inchi:

niekdejonge: And above we should also do the same. Instead of if not valid_smiles and not valid_inchi and not valid_inchikey: we should just check if not valid_smiles and not valid_inchi:

niekdejonge (commented):
@florian-huber See my comments :)

PR #582, opened 2023-11-27
Replace 'List' with 'list'

kozo2:
If not, the following error will occur.
```
NameError: name 'List' is not defined
```


florian-huber:
That's of course correct. Thanks for the notification!

florian-huber (approved):

PR #587, opened 2023-12-06
Don't repeatedly open mgf file when exporting

niekdejonge:
The exporting to mgf is very slow. >10 min for a library of 100.000 spectra.
It also gives the warning:
UserWarning: Passing a single spectrum to `write()` is discouraged. To write a set of spectra, pass them to `write()` all at once. For more info, see: https://github.com/levitsky/pyteomics/discussions/109.
  warnings.warn("Passing a single spectrum to `write()` is discouraged.
  
We can just use a generator to create the right format and this should speed up the process. (I did not yet try if this actually speeds things up on large test sets, but at least the warning will not be given anymore)


florian-huber (approved):
Sorry for the slow review and thanks for this addition @niekdejonge .

PR #596, opened 2024-01-11
Stabilize CosineGreedy by using mergesort on matches

### LINKED ISSUE #595: CosineGreedy is unstable
**Describe the bug**
`CosineSimilarity.pair` results (`scores` and `matches`) are unstable and dependent on specific hardware, even when inputs are identical. 

[Here's](https://github.com/matchms/matchms/blob/0864088b7c4e9b0ee133af286e7f1db8a2171590/matchms/similarity/CosineGreedy.py#L93C17-L93C17)  the one line that causes the instability. 

The reason:
- `np.argsort` is relying on quicksort as a default sorting algorithm ([docs](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html#numpy-argsort)). 
- Quicksort [is not stable](https://en.wikipedia.org/wiki/Quicksort). i.e. when sorting `matching_pairs` array like `1*, 3, 2, 1, 5`, we can either get either a) `1*, 1, 2, 3, 5` or b) `1, 1*, 2, 3, 5`.
- Then, `score_best_matches` is invoked we will either include pair 1* or 1 first.
- Each pair is made from two peaks, and every peak can only be used once. If 1* is `[a, b]` and 1 is `[a, c]`, depending on which comes first, we will discard either peak `b` or peak `c`.
- The final score and number of matches will differ based on how much pairs made from peaks `b` or `c` differ. 

**To Reproduce**
The full reproduction is difficult because quicksort is still deterministic on a given machine, with given numpy and numba binaries. 

The [following WIP colab notebook](https://colab.research.google.com/drive/104m3cxngVN_iqWQlIEbFJkYoBJilWyZv?usp=sharing) should at least give you an idea on how the instability can be replicated. 

Note that, due to RNG engine being different on different machines, even if you run this notebook locally, the instability you might get might not be the kind discussed here.

**Expected behavior**
The similarity algorithm should be stable, up to the numerical precision of the machine, and independent from underlying hardware.

 - OS: Ubuntu 22.10
 - Matchms version: 0.23.1 (from pip)

tornikeo:
## Description

Stabilize CosineGreedy by using mergesort on matches. Fixes #595.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Handle missing `precursor_mz` in representation and [#452](https://github.com/matchms/matchms/issues/452) introduced by [#514](https://github.com/matchms/matchms/pull/514/files)[#540](https://github.com/matchms/matchms/pull/540)
 - Fixed retention time harmonization for msp files [#551](https://github.com/matchms/matchms/issues/551)
 - Fix closing mgf file after loading and prevent reopening. [#555](https://github.com/matchms/matchms/issues/555)
+- Fix instability introduced in CosineGreedy by `np.argsort`. [#595](https://github.com/matchms/matchms/issues/595)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


florian-huber (approved):
Looks good to me. Many thanks for this catch @tornikeo .

PR #605, opened 2024-02-26
Fix array truth value error in Metadata tests

tornikeo:
Tests always fail on `tests/networking/test_SimilarityNetwork.py`, due to `ambiguous array truth value`. This was caused by [this commit](https://github.com/matchms/matchms/commit/658861eb5da346c837cff60e7aa6f87ffa9a888a).  In particular, 

```python
        metadata_filtered = {k:v for k,v in metadata_filtered.items() if v not in invalid_entries}
```

Since `metadata_filtered` has attribute `footprint` which is an `np.ndarray`, `not in` operator causes it to become a boolean array, which can't be evaluated without an `any` or `all`. In this case I solved this by only checking that string entries **are not** in invalid values.

### DIFF
--- a/matchms/Metadata.py
+++ b/matchms/Metadata.py
@@ -121,9 +121,14 @@ class Metadata:
             metadata_filtered["charge"] = charge_int
 
         invalid_entries = ["", "NA", "N/A", "NaN"]
-        metadata_filtered = {k:v for k,v in metadata_filtered.items() if v not in invalid_entries}
-
-        self.data = metadata_filtered
+        
+        metadata_filtered_ = {}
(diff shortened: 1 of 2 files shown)


niekdejonge:
@tornikeo For me all tests in test_similarityNetwork already work. Is this PR still relevant, or can it be closed?

tornikeo:
Really? Maybe it's just my python version. If it works for you, I think this can be closed. @niekdejonge

niekdejonge:
Could be, the tests pass for python 3.8 -3.11, do you use an older version?

tornikeo:
Huh... now I can't seem to reproduce the error. Guess it was something to do with my package setup. Closing this now.

PR #607, opened 2024-03-27
fixed non respected specified mz boudaries

lionel42:
The function `plot_spectra_mirror` should behave the same as `plot_spectrum` for the keyword argument `min_mz` and `max_mz`. 

this pull request provide a quick fix for that.

### DIFF
--- a/matchms/plotting/spectrum_plots.py
+++ b/matchms/plotting/spectrum_plots.py
@@ -167,6 +167,13 @@ def plot_spectra_mirror(spec_top,
             np.ceil(spec_bottom.peaks.mz[-1] / 100 + 1) * 100,
         ]
     )
+    
+    # Enforce min_mz and max_mz if provided 
+    if 'max_mz' in spectrum_kws:
+        max_mz = spectrum_kws['max_mz']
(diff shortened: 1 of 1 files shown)


lionel42:
Just adding a small test for the behaviour 

```
from matchms import Spectrum
from matchms.plotting import plot_spectra_mirror
import numpy as np

# Create two random spectrums
spec_a = Spectrum(mz=np.array([100, 200, 300, 400.2]), intensities=np.array([0.5, 0.3, 0.1, 0.05]))
spec_b = Spectrum(mz=np.array([10.2, 20.2, 30.2, 40.2, 78.2]), intensities=np.array([0.5, 0.3, 0.1, 0.05, 3.1]))

# Boundaries were not applied in the previous coce
plot_spectra_mirror(spec_a, spec_b, max_mz=250 , min_mz=60)

niekdejonge:
@lionel42 Thank you! I agree that this should be consistent with plot_spectrum. I will make some small changes to make all tests pass.

niekdejonge:
I could not make changes to your fork directly (at least did not figure out how to do so quickly), so I made a separate PR #610 that included your commit. Thanks again!

Issue #600, opened 2024-01-19
Weird `repair_adduct_based_on_smiles` `derive_ionmode` logic?


Adafede:
In [repair_adduct_based_on_smiles](https://matchms.readthedocs.io/en/latest/_modules/matchms/filtering/metadata_processing/derive_ionmode.html#), there is a line saying
```
if ion_mode not in ("positive", "negative"):
        logger.warning("Ionmode: %s not positive or negative, first run derive_ionmode",
                        ion_mode)
```

So I tried it but [derive_ionmode](https://matchms.readthedocs.io/en/latest/_modules/matchms/filtering/metadata_processing/derive_ionmode.html#), but:

```
ionmode = spectrum.get("ionmode")
    if ionmode in ["positive", "negative"]:
        return spectrum

    adduct = spectrum.get("adduct", None)
    # Harmonize adduct string
    if adduct:
        adduct = _clean_adduct(adduct)

    # Load lists of known adducts
    known_adducts = load_known_adducts()
    # Try completing missing or incorrect ionmodes
    if adduct in list(known_adducts["adduct"]):
        ionmode = known_adducts.loc[known_adducts["adduct"] == adduct, "ionmode"].values[0]
    else:
        ionmode = None 
```

But the actual goal is to infer an adduct we do not have...so...weird? Am I missing something?

niekdejonge:
@Adafede Thanks for making this issue and sorry for the late reply, I missed at first. 

When writing this warning I was under the impression that derive_ionmode also repaired based on the charge. We thrust the ionmode field over the charge field, but if ionmode is missing but charge is given it should still repair ionmode based on the charge field. 

This warning would still be given if charge, ionmode and adduct are missing. We could also change this to searching in both common positive mode and negative mode adducts, to determine both the adduct, the charge and the ionmode from the difference between precursor and parent mass

niekdejonge:
@Adafede Sorry for the delay in implementing the change. It has just been implemented in #614. Thanks again for making the issue.

Adafede:
@niekdejonge Nothing to be sorry about, thank you for implementing the fix! 😉

PR #614, opened 2024-04-05
Derive ionmode from charge

niekdejonge:
Fixes issue mentioned by @Adafede  in #600

Before we did not derive the ionmode from the charge (but only from the adduct).


florian-huber (approved):
Good changes, thanks @niekdejonge !

Issue #616, opened 2024-04-08
Reconsider repairing precursor mz


niekdejonge:
I revisited our implementations for repairing parent mass and precursor mz and think we should change some things. I would suggest to never repair the precursor mz. 

Many of the entries in GNPS entered the parent mass, but also the precursor mz manually. Mistakes are made in this process. We set out to repair the most common mistakes, but I am afraid we are repairing a bit too much. Especially when it comes to the precursor mz, since this should in principle just be an output from the machine. 

The methods I would like to reconsider:
- Repair_parent_mass_is_mol_wt
  - This does not only repair the parent mass, but also the precursor mz. The name is confusing, since it only mentions the parent mass (and not the precursor mz), but my suggested change is not repairing the precursor mz. Instead in these cases the spectrum, should just be removed, since the precursor mz was incorrectly added manually. and we can never be sure what the actual precursor mz was. 
- In repair_precursor_is_parent_mass we repair cases where the parent mass is entered instead of the precursor mz. This happens only 359 times in the entire GNPS library. The problem is that we have to guess the precursor mz afterwards. But the adduct and ionmode have to be correct and we are not sure that this is the case. It is better to just remove these cases.

florian-huber:
I agree with you. We cannot save all crappy data 😄

niekdejonge:
Great, I will make the PRs, we will have to update the manuscript once we get the revisions in.

julianpollmann:
@florian-huber @niekdejonge This issue can be closed, right?

PR #618, opened 2024-04-08
Remove repair precursor

niekdejonge:
Removed repair_precursor_is_parent_mass

As mentioned in https://github.com/matchms/matchms/issues/616 we decided to remove the filters that repair precursor_mz, since the precursor_mz can not be repaired with enough certainty, therefore it is better to remove spectra with manually wrongly added precursor mz.


florian-huber (approved):
Works for me, thanks!

PR #619, opened 2024-04-08
Fix repair parent mass is mol wt

niekdejonge:
Repair parent mass is mol wt should not repair the precursor mz as discussed in #616


florian-huber (approved):
Good changes!

niekdejonge:
There were just some merge conflicts, but it made me notice one other issue. We now removed repair_precursor_mz is parent mass. The cases that were previously repaired by that function are now repaired by replacing the adduct of "[M+H]+" with [M+] which almost has the same effect to the parent mass as changing the precursor mz would have. But, we cannot easily know the difference between these two options. So to prevent accidentally repairing this in the wrong way, I will just not repair towards [M+]+ anymore.

PR #623, opened 2024-04-08
Merge require precursor mz

niekdejonge:
We had two functions that had very similar functionality. require_precursor_mz and require_precursor_mz_below_mz. I have merged the two functions. In addition some assert statements were replaced with warnings, to not break the pipeline. (in line with other filter functions)


hechth:
@niekdejonge please update the changelog, otherwise approved :)

niekdejonge:
@hechth Thanks! I will add the changelog

PR #627, opened 2024-04-17
Added `require_formula` and `require_compound_name` filters

zargham-ahmad:
Added require_formula and require_compound_name filter functions and related tests

Related issue https://github.com/RECETOX/galaxytools/issues/507


hechth (approved):

niekdejonge (approved):
Looks good! Thanks @zargham-ahmad

matchms/filtering/filter_order.py
```diff
@@ -34,6 +34,8 @@
                msfilters.repair_not_matching_annotation,
                msfilters.require_valid_annotation,
                msfilters.require_correct_ionmode,
+               msfilters.require_compound_name,
```
niekdejonge, line 37: @zargham-ahmad I will still add one small thing. There is a test function, with tests for basic requirements for ordering the filters. I will add a test that require_formula should come after derive_formula_from_name. The rational for these tests is, that in the future we might have to reorder some filter and in that case there is less risk of accidental mistakes in the order. I will add it.

niekdejonge: @zargham-ahmad I created a PR to your branch https://github.com/zargham-ahmad/matchms/pull/1 but I did not have the rights to directly merge it into your branch.

zargham-ahmad: @niekdejonge merged, Thank you.

PR #628, opened 2024-04-18
Return processing report

### LINKED ISSUE #626: Pipeline should be able to take in custum filters
SpectrumProcessor is able to handle custom filters, however, pipeline is not. 
This is the case because create_workflow converts the filters to string representations. This works for the matchms filters, but does not work for the custom filters.

niekdejonge:
To fix #626 
Create workflow did not allow for taking in custom filters, since it would convert it to string (instead of just passing the callable).  For matchms filters this doesn't give any issues, since they would be converted back again correctly.


florian-huber (approved):
Looks good to me. Thanks @niekdejonge 
(still requires a bit of linting though)

PR #631, opened 2024-04-19
adding colors for mirrror spectra and tests

lionel42:
Before it was not possible to change the colors of the spectra in the `plot_spectra_mirror` function.

This PR makes it possible, by adding two arguments `color_top` and `color_bottom` .
In case one tries the argument `peak_color` which can be used for `plot_spectrum`, a meaningful error is raised.

Also includes tests.


niekdejonge (approved):
Thanks @lionel42 Looks good to me!

PR #636, opened 2024-05-03
Don't use unstable sorting in similarity measures

tornikeo:
Similar to #596 , ModifiedCosine and other classes also use non-stable version of `np.argsort`. I've encountered this issue once more while implementing [CudaModifiedCosine](https://github.com/pangeAI/cudams/tree/development). 

In this PR, I've changed sorting kind to `mergesort` for other similarity measures too, since this issue is likely to pop up once I've added those too.


florian-huber (approved):
Thanks @tornikeo!
Looks good to me and will be merged.
(and sorry, took me a while...)

Issue #637, opened 2024-05-03
Vectorize similarity calculations


tornikeo:
**Is your feature request related to a problem? Please describe.**
Using Numpy matmul for some vector similarity functions, instead of `@njit`, can speed up processing from 

```
CPU times: user 2.05 s, sys: 19.4 ms, total: 2.07 s
Wall time: 2.41 s
```

to:

```
CPU times: user 47.4 ms, sys: 8.95 ms, total: 56.4 ms
Wall time: 62.7 ms
```

On 1k x 1k spectra, each of 1k size.  Purely on CPU. 

Here's a colab [notebook to replicate the speedup](https://colab.research.google.com/drive/1dYvdcbkUUOubgjPAZTtsfpG5Z7duc1kw?usp=sharing).

**Describe the solution you'd like**
I already have [some functions](https://github.com/PangeAI/cudams/blob/development/cudams/similarity/vector_similarity_functions.py) vectorized over at cudams. **I'd like to move some of them to base matchms repo**. 

If that sounds fine with that proposal, I'll open a PR for it in a bit. Maybe there's some reason to avoid adding vectorized version (maybe somehow lower readability?)

tornikeo:
Closing this as done in #692

PR #638, opened 2024-05-03
Vectorise matrix similarity functions

tornikeo:
Addressing #637. I rewrote the `njitted` functions using plain matmul operations, using `numpy`. This [is around x50 times faster](https://colab.research.google.com/drive/1dYvdcbkUUOubgjPAZTtsfpG5Z7duc1kw?usp=sharing) (2s for 1k pairwise 1k peaks, vs 40ms).


niekdejonge:
@florian-huber Seems like a great change to me, but I am less familiar with this part of the code base. Could you check and merge this PR?

niekdejonge:
@tornikeo please feel free to add yourself to the citation.cff file, btw!

tornikeo:
Oh, wait a sec... Once I remove pytorch matmul from the picture, numpy seems only around 2x faster, not 50x times faster. This is all on CPU. It's weird to have such a difference between cpu-pytorch and numpy.

Still investigating this. Just a 2x acceleration might not be worth more obscure code.

niekdejonge:
Hmm yes, surprising that the difference is that large. Still 2x acceleration can be worth it, I will leave it up to @florian-huber. Let us know what comes out of the investigation :)

tornikeo:
Found the culprit. It's because [numpy only uses BLAS for floaty types](https://github.com/numpy/numpy/issues/14556). There's no integer support for BLAS matmul. Still numpy is slower than torch. With this, we arrive to around 10x speedup.

I've also expanded the [comparison notebook](https://colab.research.google.com/drive/1dYvdcbkUUOubgjPAZTtsfpG5Z7duc1kw?usp=sharing) to compare all three approaches (numpy, torch and njit).

florian-huber:
Hi @tornikeo ,
sorry, I am a bit slow in the moment but will try to look more into this asap.

florian-huber:
Performance boost indeed looks great!
I also tested this for different sizes (number of vectors):
image


Interestingly, I found no noticeable difference between the numpy and the torch implementation. Definitely nothing worth adding torch as a dependency. I would stick with the new numpy implementation as suggested by @tornikeo !

(here is my adaption of your notebook: https://colab.research.google.com/drive/1Ru1SjUzMlF3ejHAX1dbMGRRENyGTRJau?usp=sharing)

florian-huber:
I will first merge this into a local branch (`numpy_replace_njitted`) to have all tests running.

Issue #647, opened 2024-05-15
File export crashed for non utf-8 characters


florian-huber:
I imported recent MoNA data and wanted to store this as `.mgf`.
This gave me the following error:
`UnicodeEncodeError: 'charmap' codec can't encode character '\u2070' in position 102: character maps to <undefined>`

I believe this came from the msp imported spectrum below:

```
Name: methyl 8-hydroxy-4,5,7,10,14,14-hexamethyl-6,17-dioxo-16-oxapentacyclo[13.2.2.0¹,¹³.0²,¹⁰.0⁵,⁹]nonadeca-3,7-diene-9-carboxylate
Synon: $:00in-source
DB#: VF-NPL-QTOF008906
InChIKey: LISLGICORSJAKO-UHFFFAOYSA-N
Precursor_type: [M+H]+
Spectrum_type: MS2
PrecursorMZ: 443.2423
Instrument_type: LC-ESI-QTOF
Instrument: Agilent 6530 Q-TOF
Ion_mode: P
Collision_energy: 40 V
Formula: C26H34O6
MW: 442
ExactMass: 442.235538808
[...]

PR #648, opened 2024-05-15
Skip errors when loading file for msp

niekdejonge:
@florian-huber do you think this solves #647


florian-huber (approved):
Looks good and should do the trick. Thanks!

PR #650, opened 2024-05-16
Remove profile spectra

niekdejonge:
A method to remove likely wrongly converted profile spectra, by checking if there are multiple peaks surrounding the highest peak that have a high intensity.


matchms/filtering/peak_processing/remove_profiled_spectra.py
```diff
@@ -0,0 +1,66 @@
...
+        return spectrum
+
+    # mz_window = spectrum.mz[-1] - spectrum.mz[0] / 1000
+    number_of_high_intensity_surounding_peaks = _get_number_of_high_intensity_surounding_peaks(spectrum.intensities,
+                                                                                               spectrum.mz, mz_window)
+    if number_of_high_intensity_surounding_peaks < 3:
+        return spectrum
+    logger.info(
+        "Spectrum removed because likely profile data."
+        "the number of high intensity fragments next to the highest peak = %s.",
```
florian-huber:
```suggestion
        "Number of high intensity fragments next to the highest peak = %s.",
```

matchms/filtering/peak_processing/remove_profiled_spectra.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Spectra are removed if within the mz_window of 0.5 of the highest peak at least 2 peaks next to the main peak are of
+    intensity > max_intensity/2.
+
+    Reproduced from MZmine.
+    https://github.com/mzmine/mzmine3/blob/master/src/main/java/io/github/mzmine/util/scans/ScanUtils.java#L609
+    """
+    peaks_n = spectrum.mz.shape[0]
+    if peaks_n < 5:
+        return spectrum
```
florian-huber:
```suggestion
    if peaks_n < 3:
        return spectrum
```

florian-huber: I get that this is mostly to speed up the process.
But since the `5` here is a magic number coming without explanation it should be consistent with the later tested criterium.

niekdejonge: I agree. Won't speed it up that much more anyway.

florian-huber (approved):
Nice addition @niekdejonge !

Maybe add a small test that runs the full filter function. But feel free to skip.

Issue #646, opened 2024-05-15
SpectrumProcessor saving of spectra should maybe be one by one


florian-huber:
Currently the `SpectrumProcessor`  will first process all spectra (`self.process_spectrum`) and only at the end write them to file (`save_spectra(processed_spectrums, cleaned_spectra_file)`).

For larger workflows this seems not ideal because it can crash at a very late stage (just happened to me...).
I think it would be better to write each spectrum to the file once processed, at least for formats where this is possible such as `.msp` and `.mgf`.

hechth:
I agree - this is also already possible with the streaming write we implemented for those formats - we should add the json implementation though at some point.

acquayefrank:
It seems that this issue has been resolved by this PR: https://github.com/matchms/matchms/pull/651. Hence, I think it should be closed.

PR #651, opened 2024-05-16
Save mgf msp once per iteration in pipeline

tornikeo:
Addresses issue #646, which is also an issue for me. Added an extra `append` boolean kwarg to `save_spectra`, and conditionally allowed saving processed spectra inside the `process_spectrums` (depends on filetype). MGF and MSP file types now automatically get written "line-by-line". 

I refrained from adding an extra param to `process_spectrums`. I think it's better to leave the function arglist to a minimum.


niekdejonge:
Thanks! For me it is fine as well to not add an save_incrementally in process_spectrums, since I think this it is generally the wanted behaviour. If necessary for some users we can always make it optional in the future.

PR #655, opened 2024-05-18
Fail fast on loading missing mgf

### LINKED ISSUE #652: Importing should raise error if file does not exist
If you run load_from_mgf() on a file that does not exist it does not crash before you actually start to iterate over the generator. 
It would be better to already raise an error before iterating over the generator.

tornikeo:
This is my proposed way of addressing #652 . Since generators do not run any code until we `list(...)` them, it's only possible to `raise` anything if we make the generator code a wrapped function. If anyone has a more elegant solution, I'm all ears :D 

Also added a test-case for this.

### DIFF
--- a/matchms/importing/load_from_mgf.py
+++ b/matchms/importing/load_from_mgf.py
@@ -1,3 +1,4 @@
+import os
 from typing import Generator, TextIO, Union
 from pyteomics.mgf import MGF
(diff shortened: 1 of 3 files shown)


niekdejonge:
Thanks! @tornikeo Looks good to me. I don't know a more elegant solution, so let's keep this solution for now. If someone else has a better solution we can change it in the future

PR #657, opened 2024-05-22
Require matching ionmode and adduct

niekdejonge:


florian-huber (approved):
Looks straightforward and good to me. Thanks @niekdejonge !

PR #659, opened 2024-05-22
Require matching adduct precursor mz parent mass

niekdejonge:


matchms/filtering/default_pipelines.py
```diff
@@ -41,7 +41,8 @@
     (msfilters.derive_annotation_from_compound_name, {"mass_tolerance": 0.1}),
 ]
 REQUIRE_COMPLETE_ANNOTATION = [(msfilters.require_parent_mass_match_smiles, {'mass_tolerance': 0.1}),
-                               msfilters.require_valid_annotation, ]
+                               msfilters.require_valid_annotation,
+                               msfilters.require_matching_adduct_precursor_mz_parent_mass]
```
florian-huber: Generally looks fine. But for the GNPS data that could be fairly strict, right?

florian-huber (approved):
Generally looks good. I was only wondering if this wouldn't be too strict (e.g. on the GNPS data)?

matchms/filtering/default_pipelines.py
niekdejonge: We run the require_matching_adduct_precursor_mz_parent_mass filter after we run some other repair filters. Like for instance filling in the adduct from a common adduct list. So if these other repair filters are activated this is only the final check for the cases where none of the repairs worked.

PR #660, opened 2024-05-23
Allow peaks with many floating types

tornikeo:
A tiny quality-of-life addition: I've relaxed the assertion check for `Fragments.mz` and `Fragments.intensitities` dtypes. Now, any "floaty" type is acceptable - including `np.float16`, `np.float32` and `np.float64`. Added tests for this too.


niekdejonge:
Thanks! @tornikeo 
@florian-huber Is this okay? Or would this be a problem for deep learning applications after?

florian-huber:
@niekdejonge @tornikeo 
no, that shouldn't generally be a problem. So I am fine with the changes!

PR #664, opened 2024-05-27
Add require number of peaks below maximum

niekdejonge:


matchms/filtering/peak_processing/require_number_of_peaks_below_maximum.py
```diff
@@ -0,0 +1,29 @@
+import logging
+from typing import Optional
+from matchms.Spectrum import Spectrum
+
+
+logger = logging.getLogger("matchms")
+
+
+def require_number_of_peaks_below_maximum(spectrum_in: Spectrum,
```
florian-huber:
```suggestion
def require_upper_limit_of_number_of_peaks(spectrum_in: Spectrum,
```

florian-huber: Also not 100% happy with my suggestion. But I first misunderstood your function name and thought you meant only peaks below the max intensity or something like this...

florian-huber (approved):
Looks good @niekdejonge . I was just not so sure about the filter name.

Maybe change to
- `require_maximum_number_of_peaks()` ?
(or `require_upper_limit_of_number_of_peaks()` although that feels too long ...)

matchms/filtering/peak_processing/require_number_of_peaks_below_maximum.py
niekdejonge: @florian-huber How about require_less_than_maximum_number_of_peaks?

PR #665, opened 2024-05-27
Add check filters handle none

niekdejonge:
matchms filters should be able to handle spectra set to None. This test checks if this is the case for all filters.


florian-huber (approved):
Looks all good to me @niekdejonge ! Thanks!

Issue #667, opened 2024-05-29
`is_symmetric` needs more checks


tornikeo:
**Describe the bug**
`is_symmetric` requires a few extra asserts about size. Otherwise it produces confusing errors.

**To Reproduce**

Try running this:

```py
import matchms as mms

refs = tuple(mms.importing.load_from_mgf('tests/testdata/pesticides.mgf'))
ques = refs
mms.similarity.CosineGreedy().matrix(
    references=refs[:1],
    queries=ques[:32],
    is_symmetric=True
)
```

It will break without any reasonable expalantion, with an error:

```py
{
	"name": "IndexError",
	"message": "index 1 is out of bounds for axis 0 with size 1",
	"stack": "---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[7], line 5
      3 refs = tuple(mms.importing.load_from_mgf('tests/testdata/pesticides.mgf'))
      4 ques = refs
----> 5 mms.similarity.CosineGreedy().matrix(
      6     references=refs[:1],
      7     queries=ques[:32],
      8     is_symmetric=True
      9 )

File matchms/matchms/similarity/BaseSimilarity.py:90, in BaseSimilarity.matrix(self, references, queries, array_type, is_symmetric)
     88 if array_type == \"numpy\":
     89     scores_array = np.zeros(shape=(n_rows, n_cols), dtype=self.score_datatype)
---> 90     scores_array[idx_row, idx_col] = scores_data.reshape(-1)
     91     return scores_array
     92 if array_type == \"sparse\":

IndexError: index 1 is out of bounds for axis 0 with size 1"
}
```

**Expected behavior**

assertion error, is_symmetric requires inputs of equal size.

niekdejonge:
Solved with #668

PR #668, opened 2024-05-29
Check for size equality when using is_symmetric

tornikeo:
Addresses #667


niekdejonge:
@tornikeo Yes, that error should be handled better, thanks for adding!

PR #676, opened 2024-06-04
Add repair parent mass from smiles

niekdejonge:
Adds a filter to repair the parent mass from smiles
The parent mass is only repaired if it is not already close to the smiles monoisotopic mass.

Also moves all adduct/parent mass repairing filters after requires valid annotaion.


niekdejonge:
@florian-huber I am already merging this PR without code review, to get a new release ready before the weekend, so I can do a final run for the revisions with the new version.

PR #678, opened 2024-06-19
fix name position

lionel42:
I had the problem with the plot_aigainst function that the labels were not at the correct place when setting min and max mz values , because it assumed the axis starts at 0.

image

This very small fix makes it work by offsetting the labels to min_mz


niekdejonge:
@lionel42 Thank you!

PR #681, opened 2024-06-20
replace losses attribute by property method

florian-huber:
This is a new attempt for what was started in #321 .

Following a suggestion by @hechth , I now added `losses` as a `@property` so that it can still simply be called by `my_spectrum.losses`. Only if we want to adapt the min_mz and/or max-mz values we need to call `my_spectrum.compute_losses(...)`.

- new default for `loss_mz_to` is the precursor_mz


tests/filtering/test_spectrum_processor.py
```diff
@@ -120,10 +119,9 @@ def test_filter_spectrums_report(spectrums):
     assert report.counter_changed_metadata == {'make_charge_int': 2, 'interpret_pepmass': 3, 'derive_ionmode': 3}
     report_df = report.to_dataframe()
     assert np.all(report_df.loc[["require_minimum_number_of_peaks", "interpret_pepmass",
-                                 "add_losses", "correct_charge"]].values == np.array(
+                                 "correct_charge"]].values == np.array(
         [[1, 0, 0],
-         [0, 3, 0],
-         [0, 0, 2],
+         [0, 3, 3],
```
florian-huber: @niekdejonge I didn't get the number is this array or why this changed more than one line after I removed `add_losses`. Can you briefly check if that's correct?

niekdejonge: Hmm I tried it in the main branch locally. And for me the test passes when kept as [0,3,0]. I only changed this test, not the rest of the code. So that is surprising to me as well.

niekdejonge: @florian-huber I don't know what is causing this. I have made the same changes to the test in #683 Do you have an idea what could be the issue?

florian-huber: This is weird. In two ways. 
A) Interpret_pepmass should only change the metadata, and
B) Why does it change 3 spectra anyway? After the `require_minimum...` filter in the test data there should only be 2 spectra left, right?

```
s1 = SpectrumBuilder(). \
    with_metadata({"charge": "+1", "pepmass": 100}). \
    with_mz([10, 20, 30]).with_intensities([0.1, 0.4, 10]).build()
s2 = SpectrumBuilder().with_metadata({"charge": "-1",
                                      "pepmass": 102}). \
    with_mz([10, 20, 30]).with_intensities([0.1, 0.2, 1]).build()
s3 = SpectrumBuilder().with_metadata({"charge": -1,
                                      "pepmass": 104}). \
    with_mz([10, ]).with_intensities([0.1, ]).build()
spectrums = [s1, s2, s3]

processor = SpectrumProcessor(filters=["make_charge_int",
                                       "interpret_pepmass",
                                       "derive_ionmode",
                                       "correct_charge",
                                       ])
processor.parse_and_add_filter(filter_description=("require_minimum_number_of_peaks", {"n_required": 2}))
spectrums, report = processor.process_spectrums(spectrums)
```

florian-huber: Discussion on slack solved this:
- require filter runs after the others (my bet)
- and, losses are still tested in the report --> will be removed.

florian-huber: @niekdejonge fyi: I just adapted this.

matchms/Spectrum.py
```diff
@@ -234,11 +231,42 @@ def metadata(self, value):
 
     @property
     def losses(self) -> Optional[Fragments]:
-        return self._losses.clone() if self._losses is not None else None
+        return self.compute_losses()
 
-    @losses.setter
-    def losses(self, value: Fragments):
-        self._losses = value
+    def compute_losses(
```
niekdejonge, line 236: Maybe we should remove the option to only calculate the losses between loss_mz_from and loss_mz_to. This made sense before, since it would save memory. However, now I think the masking should just happen in any function that actually works with the losses and this should always return all losses. What do you think?

niekdejonge (approved):
Apart from my 1 one comment everything looks good, thanks!

matchms/Spectrum.py
florian-huber: This was not only to save memory, but also too look for losses in a particular range (say 100 to 200 Da).

But I agree, this could easily be done in a filter step after computing the losses.

florian-huber: I would probably keep it for now, because I have no good overview of who is using losses for what...
(we can then later decide to either remove this, or make it a `_compute_losses()` method for internal use only)

niekdejonge: Fine with me

PR #684, opened 2024-06-24
Added excempt label to stale.yml

julianpollmann:
* Mark issues excempt with "future developments" labels
* Increased operations per run to process all issues


florian-huber (approved):
Nice, thanks!

PR #686, opened 2024-06-25
Remove None Spectra from Spectrum lists before exporting

### LINKED ISSUE #672: Handle None spectra when saving
When saving a list of spectra containing None values the saving raises an error. 

Wanted behavior: None values in a list of spectra are skipped.

julianpollmann:
Fixes #672

@niekdejonge could you check if this fulfills your requirements?

### DIFF
--- a/matchms/exporting/metadata_export.py
+++ b/matchms/exporting/metadata_export.py
@@ -3,7 +3,7 @@ import json
 from typing import Any, Dict, List, Optional, Set, Tuple
 import numpy as np
 from ..Spectrum import Spectrum
-from ..utils import filter_empty_spectrums
+from ..utils import filter_empty_spectra
 
 
 def _get_metadata_dict(spectrum: Spectrum, include_fields: Optional[List[str]] = None) -> Dict[str, Any]:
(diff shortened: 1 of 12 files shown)


matchms/exporting/metadata_export.py
```diff
@@ -38,6 +39,7 @@ def export_metadata_as_json(spectrums: List[Spectrum], filename: str,
     identifier:
         Identifier used for naming each spectrum in the output file.
     """
+    spectrums = filter_empty_spectrums(spectrums)
```
florian-huber: Maybe we could make a start here and switch to `spectra`?
```suggestion
    spectra = filter_empty_spectra(spectra)
```

florian-huber: This won't work consistently for all code because we have too many instances of `spectrums`, in particular also in function/method names and parameters names. But I feels weird to have both spectrums and spectra in the same code block.

niekdejonge (approved):
Looks great @julianpollmann Thanks!

matchms/exporting/metadata_export.py
julianpollmann: Okay, I'll change local variables. However I wont touch the method signature of `export_metadata_as_json` to not break dependencies till tackling #687

florian-huber (approved):
Very nice! Thanks a lot @julianpollmann .

florian-huber:
@julianpollmann just realized this was not yet in the Changelog (so I added it there, hope that's fine).

PR #689, opened 2024-06-26
Fix/update actions

julianpollmann:
Updates deprecated action dependencies


julianpollmann:
Twine dependency is updated to ^5.1 to fix KeyError on pypi build. See https://github.com/pypa/twine/issues/977

florian-huber (approved):
Thanks for catching this @julianpollmann !

PR #691, opened 2024-06-28
Update numba and numpy

florian-huber:
- Allow numba up to `<0.60` and numpy up to `<1.27`.
(Helps because Pandas 2.0 requires numpy > 1.25)


niekdejonge (commented):
If the tests pass it would be fine with me. It would be compatible with required versions for MS2Deepscore and MS2Query. (but these versions were never tested, since depending on ms2query).

What surprises me a bit is that the tests passed before as well right? With pandas 2.0 and numpy < 1.25? Or did something come up that resulted in breaking changes?

florian-huber:
Not sure why we didn't see this as an issue earlier. I think it was only a pip warning in the workflow and no exception was triggered.

PR #694, opened 2024-07-04
Added encoding to loading and saving mgf

niekdejonge:
There was an issue with loading mgf libraries saved on linux machines and loaded on windows.


niekdejonge:
@florian-huber @julianpollmann I will already merge (since I will be on holiday after), but just fyi.

PR #719, opened 2024-11-22
Change scipy dependency to fix breakage in spec2vec / gensim

julianpollmann:
Installing matchms (`scipy^1.14.0`) with spec2vec leads to a dependency break because spec2vec depends on gensim, which requires scipy<1.14.0:

`gensim 4.3.3 requires scipy<1.14.0,>=1.7.0, but you have scipy 1.14.1 which is incompatible.`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
 [tool.poetry.dependencies]
 python = ">=3.10,<3.13"
 numpy = ">1.24"
-scipy = "^1.14.1"
+scipy = "<1.14.0"
 pandas = "^2.2.3"
 pillow = "!=9.4.0"
 lxml = "^4.9.3"


hechth:
@julianpollmann is this still up to date or relevant?

julianpollmann:
@hechth @niekdejonge scipy in matchms has been updated to `1.15` (see 9579bea073791be15cb0f8540d7efc1e671d750f), so this is not compatible with `spec2vec` (and `scipy <=1.10.1`, see [12d4eb5](https://github.com/iomega/spec2vec/commit/12d4eb5b3f45839eb711be4fe60c63a7a2ad8a4b)). Gensim is on 1.14.
I guess we should close this and migrate to 1.15 on spec2vec (and Gensim?)
Btw. whats the reason spec2vec is restricted to `<=1.10.1`?

julianpollmann:
Already fixed

Issue #586, opened 2023-12-04
Save empty spectrum list as empty file


niekdejonge:
Currently, empty lists of spectra are just not saved (at least for mgf). 

This is unexpected behavior to me, I would expect an empty file to be generated that when opened results in an empty list again. 

So this would mean just adding a line like:
```python
open(file_name, 'w').close()
```
to each save_spectra function.

This would only make sense if loading empty spectra files results in empty lists (or generators), which I am not sure of.

hechth:
I think it would make sense to either raise an exception or write it to the logger - just silently writing an empty file sounds like a source for errors.
I agree that creating an empty file makes sense for pipelines which for example always expect an output - so I'd go with the logging version and write it to the WARNING level.

niekdejonge:
Adding a warning would make sense to me as well

niekdejonge:
If we add this, the load_from_mgf() function should also return an empty list if an empty file is given (this should reduce the risk for errors)

wasimxyz:
Hi, long time watcher and admirer of matchms! Been wanting to start contributing, and this seems like a good first issue. It's been open for a while and seems fairly trivial, so I'm wondering if it's been completed already?

julianpollmann:
Hey @wasimsandhu, doesn't look like it has been tackled yet. Feel free to contribute!

wasimxyz:
Awesome!

PR #722, opened 2024-12-20
Save empty spectrum list as empty file

wasimxyz:
## Summary

No file is generated when `save_spectra` is called and the `spectra` parameter is an empty list, which can be unexpected behavior. This PR addresses that by generating an empty file and logging a warning. Completes #586.

## Changes

In `save_spectra`:

```python
if not isinstance(spectra, list):
    spectra = [spectra]

if len(spectra) == 0:
    logger.warning("No spectra to save. File will be empty.")
    open(file, "w").close()
    return
```

The diff is large because I reformatted the files in the `exporting` module (out of habit). I can revert these changes if desired.

## Testing

Ran the following code and confirmed that an empty file was created. Unit tests are passing on my machine.

```py
import matchms
spectra = []
matchms.exporting.save_spectra(
    spectra, file="spectra.json", export_style="json"
)
```

### DIFF
--- a/matchms/exporting/__init__.py
+++ b/matchms/exporting/__init__.py
@@ -5,15 +5,11 @@ Functions for exporting mass spectral data
 Individual :meth:`~matchms.Spectrum`, or lists of :meth:`~matchms.Spectrum`
 can be exported to json, mgf, or msp files.
 """
+
 from .save_as_json import save_as_json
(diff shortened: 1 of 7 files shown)


wasimxyz:
@julianpollmann Hi, just bumping this PR to see if there's any update. 🙂

Issue #721, opened 2024-12-20
Use Ruff for formatting


wasimxyz:
It seems as though many files in `matchms` are unformatted or manually formatted (or maybe a custom formatter was being used in a dev's local environment?). Namely, indents for multi-line statements make the code a little hard to read. For example, in `SpectrumProcessor.py`:

```python
    def __init__(self,
                 filters: Iterable[Union[str,
                                         Callable,
                                         FunctionWithParametersType]]):
        self.filters = []
        self.filter_order = [x.__name__ for x in ALL_FILTERS]
        for filter_name in filters:
            self.parse_and_add_filter(filter_name)

    def parse_and_add_filter(self, filter_description: Union[str,
                                                             Callable,
                                                             FunctionWithParametersType,
                                                             ],
                             filter_position: Optional[int] = None):
        """Adds a filter, by parsing the different allowed inputs.
        """
```

[Ruff](https://docs.astral.sh/ruff/) is a Python linter and formatter in one, and it's much faster than Flake8 and Black. Here's what the above snippet looks like after formatting (`Shift + Option + F` in VS Code):

```python
    def __init__(
        self, filters: Iterable[Union[str, Callable, FunctionWithParametersType]]
    ):
        self.filters = []
        self.filter_order = [x.__name__ for x in ALL_FILTERS]
        for filter_name in filters:
            self.parse_and_add_filter(filter_name)

    def parse_and_add_filter(
        self,
        filter_description: Union[
            str,
            Callable,
            FunctionWithParametersType,
        ],
        filter_position: Optional[int] = None,
    ):
        """Adds a filter, by parsing the different allowed inputs.
        """
```

Setup and usage is [fairly straightforward](https://docs.astral.sh/ruff/installation/). If everyone's on board, happy to open a PR and set it up, along with a format/lint GitHub Action for pull requests.

niekdejonge:
@wasimsandhu Thanks for making the issue! Sorry for the late reply, we (at least I) overlooked it before. We are looking into implementing this as a pre-commit hook

PR #743, opened 2025-03-11
Issues/721

acquayefrank:
## Describe your changes
I added pre-commit functionality that runs ruff for linting and pytest.

## Issue ticket number and link
https://github.com/matchms/matchms/issues/721

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] I have added thorough tests to see if it is a core feature.

## Known Issues

Although I tested it, I am unsure if pytest runs smoothly (there was no error, but that's often a good thing).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,8 +34,9 @@ repos:
         name: pytest
         entry: pytest
         language: system
-        types: [python]
-        
+        pass_filenames: false
+        always_run: true
+
 # sets up .pre-commit-ci.yaml to ensure pre-commit dependencies stay up to date
 ci:
   autoupdate_schedule: weekly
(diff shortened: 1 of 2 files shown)


niekdejonge (approved):
@acquayefrank Nice! I didn't know of the concept of pre-comits before. But it sounds good to me. Certainly the linting makes sense to me. 

I am not fully sure if we should add pytest to the pre-commit as well. I can foresee the following issues:
- If you don't have internet some tests don't pass, so this would make it impossible to commit without internet connection. (maybe we should just fix those tests)
- When working on larger PR's I sometimes break larger changes up in multiple commits in logical chunks that do not always pass all tests. (this is probably bad practice, but this is not exceptional.
- The run time of >20 s can make it a bit annoying (especially if it you know it cannot break tests) 

But I don't have a very strong opinion, @julianpollmann is looking into this a bit more detailed as well, so I will leave the decision up to you @julianpollmann

PR #750, opened 2025-03-11
`normalize_intensities` no longer sets Spectrum to None

### LINKED ISSUE #735: `normalize_intensities` sets Spectrum to None
**Describe the bug**
*Not a bug, but unexpected behavior*
I noticed that `matchms:normalize_intensities` will set a `Spectrum` object with all peak intensities 0 to `None`. 

`WARNING:matchms:normalize_intensities:Spectrum with all peak intensities <= 0 was set to None.`

This is unexpected, since the `load_from_mgf` function will create a `Spectrum` object even if all peak intensities are 0. I would rather expect an error during `Spectrum` instantiation if such a condition is present. Having `normalize_intensities` set the `Spectrum` object to `None` is really disruptive, since it is a cleanup function of which I would not expect such an aggressive behavior. Usually, the `Spectrum` objects are in a list, and if there are suddenly `None` types included, this leads to `'NoneType' object has no attribute 'peaks'`.

**To Reproduce**
Assume a .mgf-file such as
```
BEGIN IONS
PEPMASS=1.0
COMPOUND_NAME=Fakeomycin
MSLEVEL=2
1234.12 0
1234.12 0
END IONS
```
```
spectra = list(
    matchms.importing.load_from_mgf(filename)
)
spectra = [matchms.filtering.normalize_intensities(i) for i in spectra]
print(spectra)
```
```
2025-03-07 19:47:59,207:WARNING:matchms:normalize_intensities:Spectrum with all peak intensities <= 0 was set to None.
[None]
```

**Expected behavior**
I would expect the Spectrum object initiation to fail during reading of the mgf file. It would be also okay to simply set the intensities to 0, or set Spectrum.peaks.intensities to an empty list.

 - OS: Ubuntu 22.04
 - MatchMS version 0.28.2

**Suggestions**

I would propose the following changes:

- Include additional validation functions during `Spectrum` initiation (check if there are any ions, check if intensities != 0) and raise an error right away
- Change the behavior of `normalize_intensities()`

Let me know what you think. I have a bit of time coming up an can open a PR if desired.

Knerten0815:
When a spectrum with only negative peak intensities is passed to `normalize_intensities` the peak intensities are all set to 0 and the spectrum is returned. Previously the function returned None in that case. 

Should fix #735

### DIFF
--- a/matchms/filtering/peak_processing/normalize_intensities.py
+++ b/matchms/filtering/peak_processing/normalize_intensities.py
@@ -21,9 +21,8 @@ def normalize_intensities(spectrum_in: SpectrumType) -> SpectrumType:
     max_intensity = np.max(spectrum.peaks.intensities)
 
     if max_intensity <= 0:
-        logger.warning("Peak intensities of spectrum with all peak intensities <= 0 were set to 0.")
-        zero_intensities = np.zeros(spectrum.peaks.intensities.shape)
-        spectrum.peaks = Fragments(mz=spectrum.peaks.mz, intensities=zero_intensities)
+        logger.warning("Peaks of spectrum with all peak intensities <= 0 were deleted.")
+        spectrum.peaks = Fragments(mz=np.array([]), intensities=np.array([]))
         return spectrum
 
     # Normalize peak intensities
--- a/matchms/filtering/peak_processing/normalize_intensities.py
+++ b/matchms/filtering/peak_processing/normalize_intensities.py
@@ -21,8 +21,10 @@ def normalize_intensities(spectrum_in: SpectrumType) -> SpectrumType:
     max_intensity = np.max(spectrum.peaks.intensities)
 
     if max_intensity <= 0:
-        logger.warning("Spectrum with all peak intensities <= 0 was set to None.")
-        return None
+        logger.warning("Peak intensities of spectrum with all peak intensities <= 0 were set to 0.")
+        zero_intensities = np.zeros(spectrum.peaks.intensities.shape)
+        spectrum.peaks = Fragments(mz=spectrum.peaks.mz, intensities=zero_intensities)
+        return spectrum
 
     # Normalize peak intensities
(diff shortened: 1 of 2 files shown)


matchms/filtering/peak_processing/normalize_intensities.py
```diff
@@ -21,8 +21,10 @@ def normalize_intensities(spectrum_in: SpectrumType) -> SpectrumType:
     max_intensity = np.max(spectrum.peaks.intensities)
 
     if max_intensity <= 0:
-        logger.warning("Spectrum with all peak intensities <= 0 was set to None.")
-        return None
+        logger.warning("Peak intensities of spectrum with all peak intensities <= 0 were set to 0.")
```
niekdejonge: @Knerten0815 Great, this implements the fix correctly. Having intensities of 0 is a bit weird and unexpected. So I am afraid it might break some other similarity functions. So as an alternative it might be better to just return a Spectrum without any peaks.

Knerten0815: Ok, thanks for the review! I updated the code so that there are no more m/z values and intensities in `spectrum.peaks`.

matchms/filtering/peak_processing/normalize_intensities.py
```diff
@@ -21,8 +21,9 @@ def normalize_intensities(spectrum_in: SpectrumType) -> SpectrumType:
     max_intensity = np.max(spectrum.peaks.intensities)
 
     if max_intensity <= 0:
-        logger.warning("Spectrum with all peak intensities <= 0 was set to None.")
-        return None
+        logger.warning("Peaks of spectrum with all peak intensities <= 0 were deleted.")
+        spectrum.peaks = Fragments(mz=np.array([]), intensities=np.array([]))
+        return spectrum
```
niekdejonge, line 26: Looks good to me!

niekdejonge:
@Knerten0815 Good work! I will merge it!

PR #757, opened 2025-03-12
Implemented first version of mzSpecLib export

### LINKED ISSUE #752: Implement mzSpecLib export
Start by implementing an mzSpecLib export from the internal matchms standard.

hechth:
This PR implements a first version of the mzSpecLib export, mapping several attributes. Note that this is far from complete.

It implements unit tests and provides a clear way of how this can be extended to map more attributes.


matchms/exporting/save_as_mzspeclib.py
```diff
@@ -0,0 +1,201 @@
...
+    'inchi': 'MS:1003403|InChI',
+    'inchikey': 'MS:1002894|InChIKey'
+}
+
+SPECTRUM_ATTRIBUTES = {
+    'compound_name': 'MS:1003061|library spectrum name',
+    'precursor_mz': 'MS:1003208|experimental precursor monoisotopic m/z',
+    'scans': 'MS:1003057|scan number',
+    'charge': 'MS:1000041|charge state',
+    'ionmode': 'MS:1000465|scan polarity'
```
niekdejonge, line 19: You could maybe add retention_time, which is stored as "retention_time" by add_retention_time filter. (but maybe this is not easy, since we don't support the s/min labeling, in that case ignore this comment.)

matchms/exporting/__init__.py
```diff
@@ -9,7 +9,14 @@
 from .save_as_json import save_as_json
 from .save_as_mgf import save_as_mgf
 from .save_as_msp import save_as_msp
+from .save_as_mzspeclib import save_as_mzspeclib
```
niekdejonge, line 12: We could add save_as_mzspeclib to the wrapper function save_spectra as well.

matchms/exporting/save_as_mzspeclib.py
```diff
@@ -0,0 +1,201 @@
...
+    
+    for attr in spectrum_attributes:
+        if attr in STANDARDIZED_SPECTRUM_ATTRIBUTES:
+            _write_spectrum_attribute_with_unit(file, spectrum, attr_counter, attr)
+            attr_counter += 1
+        elif attr not in SPECTRUM_ATTRIBUTES and attr not in ANALYTE_ATTRIBUTES:
+            _write_other_spectrum_attribute(file, spectrum, attr_counter, attr)
+            attr_counter += 1
+
+    print(f'MS:1003059|number of peaks={len(spectrum.peaks)}', file=file)
```
niekdejonge, line 98: Should this print statement still be converted to adding the number of peaks to the file?

niekdejonge: Never mind, I wasn't aware that it was possible to print to a file. But as long as that doesn't give performance issues, it is of course fine!

niekdejonge (approved):
Looks good @hechth I left some small comments, but feel free to merge! 

Did you check that spectra saved by matchms can be loaded by mzspeclib as well? 
And do you think it is worth adding a loader as well?

hechth:
@niekdejonge I started on a loader during the workshop but didn't get too far and didn't have the time since then to work on it. So if we can merge this PR, it would be a good first step.

matchms/exporting/save_as_mzspeclib.py
hechth: Yeah we don't know which unit it is in and I'd rather not write it out than write it out wrong

PR #769, opened 2025-03-17
Refactor base similarity

niekdejonge:
The following changes were made:
- **Removed array_type parameter.** BaseSimilarity.matrix() now always returns a numpy matrix. And actually computes a dense matrix instead of computing sparse and converting to a dense matrix.
- **Implement FilterScoreByValue**, this replaces the BaseSimilarity.keep_score method, allowing for more flexibility in filtering during compute (and before storing). Before it was only filtering out 0 values.
- **Rewritten BaseSimilarity.matrix()**, the matrix was computed as a sparse array, leading to a lot higher memory footprint. Because it uses lists and it needs to score both indexes in addition to the scores. Now it is computed as a numpy matrix. 
- **Added two new similarity score computation methods** (starting from dense and starting from sparse) that do apply the FilterScoreByValue filter, during the compute. This makes sure that any score that is does not pass the filter is not stored. However, this adds additional memory, since a new index has to be build and therefore is only worth it if less than 1/12th of the scores are stored. (see section below)
- **Separate Scores and Basesimilarity**. The score computation was heavily intertwined. It happened in Scores, BaseSimilarity, calculate_scores and Pipeline. Especially the fact that Scores had a calculation function was confusing to me. I have now more cleanly separated the logic. All score_computation methods are implemented in BaseSimilarity and the rest just calls BaseSimilarity. 

To do after initial check of right implementation:

- [ ] Add tests for new BaseSimilarity methods and calculate scores method and ScoreFilters.
- [ ] Optional: Add optimized methods for all different CosineScores implementations.
- [ ] If we stick with FilterScoreByValue during similarity compute, update Pipeline to use this and replace _masking_functions = ["filter_by_range"]
- [ ] Make the tests in test-scoring_method_consistency actually check stuff and add check for is_symmetric
- [ ] When mask and filters are None sparse array should still run, by just running matrix followed by a conversion. This should be implemented once we settle on a scores representation, because the temporary COO_matrix does not yet support building from dense matrix. 

Why I applied these changes:
The old implementation of BaseSimilarity had the option to output matrixes or sparse arrays. This was a bit confusing and required all post functions to handle both output types. Now it just always outputs the matrix type.  

Before the matrix function did apply the keep_score filter, while the sparse array did not. This is inconsistent and confusing. The applying of keep_score seems like a more memory efficient method, since you don’t require to store the scores below a threshold, however, it also requires us to keep track of the indexes of the sparse array in a list. This is most of the time not speed nor memory efficient, since it requires rewriting of the indexes to a list. 

**Why applying score filters right away is often not memory efficient**
Numpy is about 4 times more memory efficient than a list. (because of fixed types). And if we want to store only the scores that are not zero, we need both indexes and the score, so 3 lists. This means that storing the filtered scores is 12 times less memory efficient. 
In the case when you already start from a sparse array, the moment that the memory footprint is less than 12 times, the scores are anyway not a significant part of memory anyway. Since the "mask" is already stored, which is 2/3 of the memory. So in the most extreme case where 0 scores pass the filter, this saves 1/3 of memory. While in any other case the memory footprint is 12x what it would be without filtering. This to me seems like a trade off that is never beneficial to the filtering during compute. 
When doing the first compute, which always requires computing a dense matrix. This is less obvious, still the ratio is 12x, but in this case it can actually happen that the difference is 12x. E.g. when computing mass differences with max diff = 1.0. Or when we only want the top 10 hits per row. In this case it might make sense, especially for the very large matrixes to support filtering during the compute. But it is good to realize that it is only worth it if you keep less than 1/12th of the scores. 

**Decisions to still make**
- I have now implemented methods for filtering during compute. However, this is only efficient if it removes more than 11/12th of the scores, otherwise a more memory efficient implementation is just computing all scores first followed by doing a filtering step. 1. We could therefore also just remove the (new) methods that allow filtering during and only allow filtering on the Scores objects, or just do normal compute followed by filtering within BaseSimilarity. 2. We could do a quick dummy run to determine the optimal run settings. If the matrix is large enough first a few random pairs are computed, if the filter filters out >11/12th the filter is directly applied otherwise it is applied after.
- What will be the intended way of using BaseSimilarity? Always using the Scores object as input? Should only BaseSimilarity.calculate_scores be a public method? Or also the specific implementations? 
- Should we completely remove calculate_scores?


matchms/similarity/BaseSimilarity.py
```diff
@@ -39,28 +43,113 @@ def pair(self, reference: SpectrumType, query: SpectrumType) -> float:
...
+            else:
+                new_scores = self.matrix_with_filter(scores.references, scores.queries,
+                                                     is_symmetric=scores.is_symmetric, score_filters=filters)
+
+        if isinstance(new_scores, COOMatrix):
+            scores.scores.add_sparse_data(new_scores.row,
+                                          new_scores.column,
+                                          new_scores.scores,
+                                          name,
+                                          join_type="left")
```
florian-huber:
```suggestion
                                          join_type=join_type)
```

matchms/calculate_scores.py
```diff
@@ -5,7 +5,6 @@
 
 def calculate_scores(references: ReferencesType, queries: QueriesType,
```
florian-huber: I think we should not have a `calculate_scores()` function AND a `BaseSimilarity.calculate_scores()` method with different implementations. Could we not have a function `calculate_scores()` which is then used in the BaseSimilarity class?

florian-huber (approved):
I think we should avoid having two `calculate_scores()` implementations (see comments). For the rest this looks like a very good start, thanks a lot @niekdejonge !

matchms/calculate_scores.py
niekdejonge: Good point. I have removed calculate_scores as a method and converted it to a function. I renamed the original calculate_scores function to create_scores_object_and_calculate_scores. I am fine with removing this in the end, but it was currently used in many tests, so I kept it for now.

matchms/similarity/FingerprintSimilarity.py
```diff
@@ -104,9 +105,8 @@ def pair(self, reference: SpectrumType, query: SpectrumType) -> float:
 
         raise NotImplementedError
 
-    def matrix(self, references: List[SpectrumType], queries: List[SpectrumType],
-               array_type: str = "numpy",
-               is_symmetric: bool = False) -> np.array:
+    def _matrix_without_mask_without_filter(self, references: List[Spectrum], queries: List[Spectrum],
```
florian-huber: This method is not used anywhere. Is there a `matrix()` method missing?

florian-huber: OK, sorry, now I got the changes in the `BaseSimilarity` class.

matchms/similarity/BaseSimilarity.py
```diff
@@ -13,17 +15,33 @@ class BaseSimilarity:
...
 
+    def __init__(self, score_filters: Tuple[FilterScoreByValue, ...]):
+        """
+
+        Attributes
+        ----------
+        score_filters:
+            Tuple of filters that should be applied to each score before scoring. If you want to run without filtering,
+            it is best to run matrix(). Filters can also be run after computing a matrix first, however, for strict
+            filtering this implementation can be more memory efficient. Only use matrix_with_filter if you expect to
```
florian-huber: There is no more `matrix_with_filter`.

julianpollmann:
Hey @florian-huber @niekdejonge,
while looking through BaseSimilarity a few questions came to my mind:

- Does [_matrix_with_mask](https://github.com/matchms/matchms/blob/refactor_base_similarity/matchms/similarity/BaseSimilarity.py#L150-L158) depend on [self.score_filters](https://github.com/matchms/matchms/blob/refactor_base_similarity/matchms/similarity/BaseSimilarity.py#L41) being set? If _score_filters_ is an _empty tuple ()_ no scores are added at all and _[sim_matrix](https://github.com/matchms/matchms/blob/refactor_base_similarity/matchms/similarity/BaseSimilarity.py#L150)_ with zeroes is returned
- Should we check `if mask_indices is None` (=will only be True, if is None; COOIndex might have no Items) or `if mask_indices` (=will check also if COOIndex have Items). See this [SO ](https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false)Question

florian-huber:
@julianpollmann I would go for `if x is None:`. Python will also work with `if x:` but to me, when reading the code, it is always unclear if `None` would be included.

julianpollmann:
@florian-huber should we then add a check to guarantee that COOIndex has items?

niekdejonge:
@julianpollmann Regarding: "If score_filters is an empty tuple () no scores are added at all" That is indeed unintended. When no filters are set, every score should just be added. I thought that the code would already work like that, since np.all([]) returns True. But I could certainly be wrong here, in that case that should indeed be changed. It might make sense to also add a test that empty filters are handled correctly (If I did not add this yet).

Regarding "Should we check if mask_indices is None". You mean in the matrix method right? This is indeed intended to only capture None. If the COOindex is empty it should still run _matrix_with_mask(), which will result in a matrix with only zeros, which is the intended behaviour. This situation can happen when running a pipeline with filters. So for instance when first selecting only precursor mz difference <0.1 Da. For some datasets no pair has this difference <0.1 , in that case the expected result is an empty matrix, since no pairs passed the first check. Of course in that case it would be more memory efficient to run sparse_array(), but that logic is up to the user, or handled in the calculate_scores function. 

Does that answer the questions you had?

niekdejonge:
This has now been replaced by #864

PR #774, opened 2025-03-20
Fix/speedup filters

julianpollmann:
- This adds the parameter to all filters to clone the spectrum in the filter, defaulting to True in the filter.
- SpectrumProcessor will clone each spectrum just once and disables cloning in each filter.

- ProcessingReport relies in cloning of spectrum inside the filter, therefore the param `create_report` is added to `process_spectra` is added, defaulting to False.

- Adds better type hinting for the filters and improved docstrings.

### DIFF
--- a/matchms/filtering/SpectrumProcessor.py
+++ b/matchms/filtering/SpectrumProcessor.py
@@ -152,8 +152,18 @@ class SpectrumProcessor:
         """
         if processing_report is not None:
             processing_report.counter_number_processed += 1
+
+        clone = processing_report is not None
+
         for filter_func in self.filters:
-            spectrum_out = filter_func(spectrum, clone=False)
+            method_params = inspect.signature(filter_func).parameters
+
+            if clone and "clone" not in method_params:
+                logger.error("Processing report is set to True, but the filter function %s does not support cloning.", filter_func.__name__)
+
+            # Ensures that filter function can be used, even if it does not support cloning
+            spectrum_out = filter_func(spectrum, **({"clone": clone} if "clone" in method_params else {}))
+
             if processing_report is not None:
                 processing_report.add_to_report(spectrum, spectrum_out, filter_func.__name__)
             if spectrum_out is None:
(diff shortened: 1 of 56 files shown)


matchms/filtering/SpectrumProcessor.py
```diff
@@ -152,8 +152,18 @@ def process_spectrum(self, spectrum,
...
         if processing_report is not None:
             processing_report.counter_number_processed += 1
+
+        clone = processing_report is not None
+
         for filter_func in self.filters:
-            spectrum_out = filter_func(spectrum)
+            method_params = inspect.signature(filter_func).parameters
+
+            if clone and "clone" not in method_params:
```
niekdejonge: If a filter functions doesn't support clone, but does return None (in the case of require filters). This would log an error right? Which is maybe not intended? 

I am not sure how to solve it though...

niekdejonge (approved):
Great work @julianpollmann. Nice that you right away made typehinting more complete and completed the docstrings. It is a bit annoying that the processing report requires the cloning... I forgot about that. But not really something we can do anything about easily now. So I would merge for now. If we in the future still make the transition to Filter classes, we can consider making each filter return True or False to tell the Spectrum Processor if the spectrum was actually changed.

matchms/filtering/SpectrumProcessor.py
julianpollmann: Hey,
yes you're right. If ProcessingReport is True, require filters would trigger this error. We could exclude them from this check:
```
if not filter_func.__name__.startswith("require_"):
    if clone and "clone" not in method_params:
        logger.error("Processing report is set to True, but the filter function %s does not support cloning.", filter_func.__name__)
```
and remove cloning from [require_correct_ionmode ](https://github.com/matchms/matchms/blob/fix/speedup-filters/matchms/filtering/metadata_processing/require_correct_ionmode.py#L35) and [require_parent_mass_match_smiles](https://github.com/matchms/matchms/blob/fix/speedup-filters/matchms/filtering/metadata_processing/require_parent_mass_match_smiles.py#L35).

What do you think? Or might that affect the result of the filters?

On the long term it might be beneficial to have a "global" filter config and a per filter config.

julianpollmann: @niekdejonge I removed cloning from require filters and disabled the logging error for these cases: https://github.com/matchms/matchms/pull/774/commits/22b8204b2f50b4c975df3e284ea1fc81b1c3a9b1
IF ok, I would merge

niekdejonge: Sorry I missed your previous reply. That sounds good to me! It makes me more and more think that classes would have made a lot of sense for the filters. But for now this is a good solution, so I think it is ready to be merged!

Issue #747, opened 2025-03-11
pip install -e .[dev] doesn't work


niekdejonge:
There are multiple cases where pip install -e .[dev] does not seem to install the dev dependencies

niekdejonge:
Apparently the update to the readme I implemented does not fix it. In general I just don't get poetry to work...
What does work is pip install -e . 
However installing dev doesn't work. 
So I will just do pip install -e . followed by manually installing all dev dependencies. 

@julianpollmann or @hechth or @florian-huber 
I have never used poetry before the workshop, can someone else look into this? 
I get all kind of weird issues when I try to install using Poetry, so I feel like switching back to setup.py at the moment.

julianpollmann:
This also affects the test dev job in CI Build workflow, but only on Windows systems.
Downgrading to poetry 1.8 seems not to solve this issue.

niekdejonge:
Yes, I am not sure what is going on. Do you know if poetry has any benefits over pip? Otherwise I would propose to just switch back to a setup.py, since I have never had issues with that. 

@hechth Do you have downstream functionality that requires poetry? Or any ideas how this can be resolved?

julianpollmann:
Poetry is right now used as build tool for releasing matchms to pypi. Therefore we should keep it.
The issue relates to [poetry-bumpversion](https://github.com/matchms/matchms/blob/master/pyproject.toml#L59) forcing a downgrade to poetry 1.8.4 (Although it should be compatible with poetry 2.x). On windows poetry cannot downgrade itself, therefore the issue.
I commented on an issue on [poetry-bumpversion](https://github.com/monim67/poetry-bumpversion/issues/14), but we might remove that dependency for now?!

niekdejonge:
Yes I think we could just remove this. I believe the only thing poetry-bumpversion is used for now is updating __version__ at two locations at the same time. So if we have to do this manually this should not be a big deal.

PR #777, opened 2025-03-21
Bump twine to fix dependency issues when using poetry

julianpollmann:
Dependency issues between poetry-bumpversion and twine did result in poetry installation fail on Windows, due to downgrading of poetry.
Bumping twine to 6.x solves this issue.

- [x] Test with pip dev install
- [x] poetry editable install
- [x] Update Docs


julianpollmann:
This fixes #747

niekdejonge:
@julianpollmann Great, thanks for fixing into this annoying issue!

PR #779, opened 2025-03-24
Refactor multiple scores

niekdejonge:
It was often confusing that some scores return multiple values.
The use case most of the times is filtering on these scores, but after the filtering it is not necessary. 

Here a proposed solution: 
Add a parameter to BaseSimilarity (and subclasses): only_store_main_score. If this is True all matrix methods just return a single score (the main one). If False it will return all scores, e.g. including matches. If this setting is set to True both scores are still calculated and can be used with the FilterScoreByValue class (by setting score_name="nr_of_matches"), but after filtering it is not stored.

To do (after confirmation that this is a good fix) 
- [ ] Add call to super.__init__ for SimilarityScores
- [ ] Add tests testing this mechanic. 
- [ ] Add a check that filter on anything else cannot be combined if only one score is returned for the current filter after matrix compute method. This requires some fixes

**Note, some merge conflicts need to be resolved, but the check if this makes sense can still be made**

### DIFF
--- a/matchms/similarity/BaseSimilarity.py
+++ b/matchms/similarity/BaseSimilarity.py
@@ -123,7 +123,7 @@ class BaseSimilarity:
                      references: np.ndarray[SpectrumType], queries: np.ndarray[SpectrumType],
                      is_symmetric: bool = False
                      ) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


niekdejonge:
We decided to remove the multiscore functionality altogether, so this will not be necessary anymore

PR #781, opened 2025-03-25
Added a few reference spectra

LucaCappelletti94:
As per issue #778, I am adding a handful of reference spectra I am using for tests myself. These are:

* Aspirin
* Cocaine
* Glucose
* Hydroxy-cholesterol
* Phenylalanine
* Salicin

I may add a few more in the future.

PR #786, opened 2025-03-27
Added reference spectra

florian-huber:
Comes from #781 by @LucaCappelletti94


florian-huber:
Looks good to me now. Thanks @LucaCappelletti94 !

PR #788, opened 2025-03-31
Update BaseSimilarity.py

florian-huber:
- Editing docstrings to make API a bit clearer.
- add docstrings and make indentations more consistent

### DIFF
--- a/matchms/similarity/BaseSimilarity.py
+++ b/matchms/similarity/BaseSimilarity.py
@@ -9,15 +9,17 @@ from matchms.typing import SpectrumType
 
 
 class BaseSimilarity:
-    """Similarity function base class.
-    When building a custom similarity measure, inherit from this class and implement
-    the desired methods.
+    """
+    Base class for similarity functions.
+
+    To implement a custom similarity measure, subclass BaseSimilarity and implement
+    the `pair` method, which calculates the similarity between two spectra.
 
     Attributes
     ----------
     is_commutative:
-       Whether similarity function is commutative, which means that the order of spectra
-       does not matter (similarity(A, B) == similarity(B, A)). Default is True.
+        Indicates whether the similarity function is commutative (i.e. similarity(A, B) == similarity(B, A)).
+        Defaults to True.
     score_datatype:
         Data type for the score output, e.g. "float" or [("score", "float"), ("matches", "int")].
         If multiple data types are set, the main score should be set to "score" (used as default for filtering).
@@ -28,42 +30,46 @@ class BaseSimilarity:
 
     def __init__(self, score_filters: Tuple[FilterScoreByValue, ...]):
         """
-
-        Attributes
+        Parameters
         ----------
         score_filters:
-            Tuple of filters that should be applied to each score before scoring. If you want to run without filtering,
-            it is best to run matrix(). Filters can also be run after computing a matrix first, however, for strict
-            filtering this implementation can be more memory efficient. Only use matrix_with_filter if you expect to
-            filter out more than 90% of your data. Otherwise, it is more memory efficient to first run matrix, followed by
-            filtering the matrix.
+            A tuple of filter objects to apply to each similarity score.
+            - If you do not wish to apply any filtering, pass an empty tuple and use the dense matrix method (`matrix()`).
+            - For cases where you expect to filter out more than 90% of the computed scores, consider using
+              the sparse array methods (see `sparse_array()`).
         """
         self.score_filters = score_filters
 
     @abstractmethod
     def pair(self, reference: SpectrumType, query: SpectrumType) -> np.ndarray:
-        """Method to calculate the similarity for one input pair.
+        """
+        Compute the similarity score for a single pair of spectra.
 
         Parameters
         ----------
         reference
-            Single reference spectrum.
+            A single reference spectrum.
         query
-            Single query spectrum.
+            A single query spectrum.
 
         Returns
-            The similarity score as numpy array (using self.score_datatype). For instance returning
-            np.asarray(score, dtype=self.score_datatype)
+            The similarity score as numpy array (with dtype given by self.score_datatype).
+            For example: return np.asarray(score, dtype=self.score_datatype)
         """
         raise NotImplementedError
 
-    def matrix(self, references: Sequence[SpectrumType],
+    def matrix(self,
+               references: Sequence[SpectrumType],
                queries: Sequence[SpectrumType],
                is_symmetric: bool = False,
-               mask_indices: COOIndex = None,
-               ) -> np.ndarray:
+               mask_indices: COOIndex = None
+              ) -> np.ndarray:
         """
         Compute a dense similarity matrix for all pairs of reference and query spectra.
+
+        Use this method when filtering is either required (with a provided mask) or when working
+        with moderately sized datasets. If no filtering is needed and you don't have a mask, simply
+        pass an empty tuple for `score_filters` and use this method.
 
         Parameters
         ----------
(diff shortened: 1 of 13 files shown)


matchms/similarity/BaseSimilarity.py
```diff
@@ -28,87 +30,126 @@ class BaseSimilarity:
...
+            queries: Sequence[SpectrumType],
+            is_symmetric: bool = False,
+            mask_indices: COOIndex = None) -> np.ndarray:
         """
         Compute a dense similarity matrix for all pairs of reference and query spectra.
 
+        Use this method when filtering is either required (with a provided mask) or when working
+        with moderately sized datasets. If no filtering is needed and you don't have a mask, simply
+        pass an empty tuple for `score_filters` and use this method.
+
```
niekdejonge:
```suggestion
        Use this method when you expect most scores to pass the filtering. For instance if you don't apply any filtering. But also when you expect filtering to filter out less than 90% of all calculated scores. Filters can be added when creating the filter class object.
```

niekdejonge: This was a bit confusing to me, would this capture the same?

matchms/similarity/BaseSimilarity.py
```diff
@@ -28,87 +30,126 @@ class BaseSimilarity:
...
-            Indices to calculate scores for the rest is set to 0.
         is_symmetric:
-            Indicates if the similarity matrix is symmetric (e.g., for all-vs-all comparisons).
-            When True, only the upper triangle of the matrix is computed and then mirrored,
-            which can reduce computation time.
+            If True, indicates that the similarity matrix is symmetric (i.e., references and queries are the same).
+            Only the upper triangle is computed and then mirrored. Defaults to False.
+        mask_indices:
+            A COOIndex instance specifying which pairs to compute.
+            If provided, only the specified index pairs will be computed (others remain zero).
```
niekdejonge, line 86:
```suggestion
            If provided, only the specified index pairs will be computed (others remain zero). This is helpful when a previous score already filters out many of the pairs, reducing computation time and memory footprint. 
```

matchms/similarity/BaseSimilarity.py
florian-huber: Yes, should work, too. Only it is not clear what "filter class object" is I think.

florian-huber: We should be very careful here because the word "filter" is also use for spectrum processing in matchms, so better make this as clear as possible.

matchms/similarity/BaseSimilarity.py
```diff
@@ -28,87 +30,127 @@ class BaseSimilarity:
...
+            self,
+            references: Sequence[SpectrumType],
+            queries: Sequence[SpectrumType],
+            mask_indices: COOIndex = None,
+            is_symmetric = False) -> COOMatrix:
+        """
+        Compute a sparse array (in COO format) of similarity scores.
+
+        Use this method if you expect heavy filtering (i.e. many scores are dropped) or
+        if you want to compute scores only for a selected set of index pairs.
```
niekdejonge:
```suggestion
        if you want to compute scores only for a selected set of index pairs. By using sparse_array() this can reduce the memory footprint if many scores are dropped. 
```

matchms/similarity/BaseSimilarity.py
```diff
@@ -28,87 +30,127 @@ class BaseSimilarity:
...
         sim_matrix = self._matrix_without_mask_without_filter(references, queries, is_symmetric=is_symmetric)
         for score_filter in self.score_filters:
             sim_matrix = score_filter.filter_matrix(sim_matrix)
         return sim_matrix
 
-    def _matrix_without_mask_without_filter(self,
-                                            references: Sequence[SpectrumType], queries: Sequence[SpectrumType],
-                                            is_symmetric: bool = False
-                                            ) -> np.ndarray:
+    def _matrix_without_mask_without_filter(
```
niekdejonge, line 177: We could also merge this function with _matrix_without_mask_with_filter. Would that make more sense? Or would that merge to much logic in one function?

matchms/similarity/BaseSimilarity.py
```diff
@@ -142,11 +184,28 @@ def _matrix_without_mask_without_filter(self,
...
-                          ) -> np.ndarray:
+    def _matrix_with_mask(
+            self,
+            references: Sequence[SpectrumType], queries: Sequence[SpectrumType],
+            mask_indices: COOIndex,
+            is_symmetric: bool = False) -> np.ndarray:
+        """
+        Compute a dense similarity matrix using a provided mask.
+
+        Only the (row, column) pairs specified in mask_indices are computed.
```
niekdejonge:
```suggestion
        Only the (row, column) pairs specified in mask_indices are computed. All pairs not in mask_indices are set to 0. 
```

matchms/similarity/BaseSimilarity.py
```diff
@@ -231,13 +296,19 @@ def _sparse_array_with_mask_without_filter(self, references: Sequence[SpectrumTy
             scores[i] = self.pair(references[i_row], queries[i_col])
         return COOMatrix(row_idx=mask_indices.idx_row, column_idx=mask_indices.idx_col, scores=scores, scores_dtype=self.score_datatype)
 
-    def _sparse_array_with_filter(self, references: Sequence[SpectrumType], queries: Sequence[SpectrumType],
-                                 mask_indices: COOIndex) -> COOMatrix:
-        """Uses a mask to compute only the required scores and filters scores that do not pass the filter.
+    def _sparse_array_with_filter(
```
niekdejonge: Maybe we should name this to _sparse_array_with_filter_with_mask, to follow consistent naming.

matchms/similarity/BaseSimilarity.py
```diff
@@ -231,13 +296,19 @@ def _sparse_array_with_mask_without_filter(self, references: Sequence[SpectrumTy
...
+        """
+        Compute a sparse similarity matrix (COO format) using a mask and applying score filters.
 
-        This method most of the time does not make sense. It is only worth it if you want to store less than 1/12th
-        of the computed scores. This is not the case most of the time, so sparse_array is most of the time the most
-        memory efficient.
+        Only the (row, column) pairs specified in mask_indices are computed, and only scores that pass
+        all filters are stored.
+
+        Please note: If the mask_indices contain more than 1/12th of all indices, this method could become less memory efficient
```
niekdejonge, line 384: This is slightly different. If the score filter removes less than 1/12th of all the pairs (in the mask_indiced), it is less memory efficient than just computing and storing all pairs without filtering and applying the filter after. 

It might make more sense to actually re-implement this method, to just run _sparse_array_with_mask_without_filter followed by filtering out based on the score filters. 
This saves quite some memory, since no lists are needed and the mask_indices are just reused instead of added again to a list.

julianpollmann:
Hey @florian-huber @niekdejonge I changed the default value of BaseSimilarity's score_filters to None instead of an empty tuple.

niekdejonge:
@julianpollmann please make sure that the changing the score filters to None does not break the code. I wrote it now to expect an empty tuple. For an empty tuple a loop like: for score_filter in self.score_filters in _matrix_without_mask_with_filter will just skip the filtering, while None will result in an error. 

However, I agree that it should be accepted to set filters to None. 
I would propose, internally still storing the filters as a tuple (also empty if None), so this logic of looping over filters remains easy. 
In __init__ we could add:         self.filters = filters if filters is not None else ()

matchms/similarity/BaseSimilarity.py
julianpollmann: I would keep them seperated.

matchms/similarity/BaseSimilarity.py
julianpollmann: Should we then also rename `_matrix_with_mask` to `_matrix_with_filter_with_mask` to make it more consistent?

julianpollmann:
@niekdejonge I set `score_filters` to None in init and use a tuple internally.

matchms/similarity/BaseSimilarity.py
niekdejonge: Yes indeed! 

And maybe also fix the ordering, while we are at it. Now we start with mask most of the time, but a few cases start with filter instead. 

So to change is:
_sparse_array_with_filter_without_mask -> _sparse_array_without_mask_with_filter
_sparse_array_with_filter -> _sparse_array_with_mask_with_filter
_matrix_with_mask -> _matrix_with_mask_with_filter

julianpollmann: Done

PR #790, opened 2025-04-08
Add ruff and pre-commit hook

julianpollmann:
This does following:

- remove isort, pylint, prospector and black
- add ruff as linter and formatter
- add a pre-commit hook to automatically lint/format modified files before a commit

To ensure a smooth transition, some linting rules of ruff are disabled to mimic the medium profile of prospector. These need to be enabled later.


julianpollmann:
Tests will fail due to Tests for EmbeddingSimilarity #792

PR #807, opened 2025-06-06
Fix msp import set key as numpy scalar instead of value

### LINKED ISSUE #806: matchms 0.29.0 and 0.30.0 reads and writes some fields in peak_comments as numpy scalars.
**Describe the bug**
I am unsure if this happens to all readers and writers, but I noticed when reading a `.msp` file that the peak_comments key is a Numpy scalar. This is not necessarily a problem, but it trickles down when writing `.mgf` files. You end up with outputs containing  `np.float64(136.03947)` as a key. This causes problems later when reading the `.mgf` file.

**To Reproduce**
Steps to reproduce the behaviour:
1. Install matchms 0.29.0  or 0.30.0
2. Download the test file from [here](https://github.com/acquayefrank/galaxytools/blob/master/tools/matchms/test-data/similarity/RECETOX_Exposome_pesticides_HR_MS_20220323.msp)
3. Run the code below, and you will notice the keys are Numpy scalars, e.g. `np.float64(89.03882).`
``` python
from matchms.importing import load_from_msp

spectra = load_from_msp("RECETOX_Exposome_pesticides_HR_MS_20220323.msp")

spectra_list = list(spectra)

for spectrum in spectra_list:
    for key in ["PEAK_COMMENTS", "peak_comments"]:
        if key in spectrum.metadata:
            print(spectrum.metadata[key])
            break
```
3.  Run the code below to save the spectra:

``` python
from matchms.exporting import save_as_mgf


save_as_mgf(spectra_list, "res.mgf")
```
4. Open the saved file, and you will see a line like this
<pre>
PEAK_COMMENTS={np.float64(112.07591): 'Theoretical m/z 112.075687, Mass diff 0 (1.99 ppm), SMILES O=C(C=CC)N(C)C, Annotation [C6H11NO-H]+, Rule of HR True', np.float64(127.01563): 'Theoretical m/z 127.01547, Mass diff 0 (1.26 ppm), SMILES O=P(O)(OC)OC, Annotation [C2H7O4P+H]+, Rule of HR True', np.float64(193.02605): 'Theoretical m/z 193.026035, Mass diff 0 (0.08 ppm), SMILES O=CC=C(OP(=O)(OC)OC)C, Annotation [C6H11O5P-H]+, Rule of HR True', np.float64(238.08437): 'Theoretical m/z 238.083891, Mass diff 0 (2.01 ppm), SMILES O=C(C=C(OP(=O)(OC)OC)C)N(C)C, Annotation [C8H16NO5P+H]+, Rule of HR True'}  </pre>

5. Try reading the file:
``` python
from matchms.importing import load_from_mgf

read_spectra = list(load_from_mgf("res.mgf"))
```
and you will end up with this error
<pre>
ValueError: malformed node or string on line 1: 
</pre>

**Expected behaviour**
when writing to the file, I should have  115.05463 as the key, not np.float64(115.05463).

**Screenshots**
NA

 - OS:  WSL - Ubuntu
 - Version 24.04.2 LTS

**Additional context**
This was running in the galaxy instance. But I also tested it outside the galaxy instance.

julianpollmann:
Should fix #806 for msp.
Needs testing with other formats.

### DIFF
--- a/matchms/importing/load_from_msp.py
+++ b/matchms/importing/load_from_msp.py
@@ -133,54 +133,59 @@ def parse_metadata(rline: str, params: dict):
     """Reads metadata contained in line into params dict.
 
     The complexity of this function stems from the fact that MSP allows for many different formats of metadata.
+
+    Parameters
+    ----------
+    rline: str
+        The line of the read MSP file that contains metadata.
+    params : dict
+        The params that the key value pairs of the metadata line will be added to.
     """
-    matches = []
     splitted_line = rline.split(":", 1)
+    if len(splitted_line) != 2:
+        return
 
-    if splitted_line[0].lower() == "comments" and "=" in splitted_line[1]:
-        # This pattern check for different formats.
-        # The first checks for compound="caffeine", the second and third "compound=caffeine and the last
-        # for parent_mass=12.1
-        # The second and third almost match the same pattern, but the second is there to ensure tha the pattern
-        # "SMILES=CC(O)C(O)=O" is recognized as 'smiles': 'CC(O)C(O)=O' instead of 'smiles=cc(o)c(o)': 'O'
-        # The third pattern is needed since some keys like DB# should be stored, which is not recognized by the \w regex
-        # Cases like "Smiles=" will not be stores (since len(match) = 1) The reason that we did not change the * to +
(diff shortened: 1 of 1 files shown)


acquayefrank:
@julianpollmann, will there be a Pypi release for this?

julianpollmann:
@acquayefrank yes, we might release a patch version.

PR #810, opened 2025-06-13
Release 0.30.1

julianpollmann:


hechth:
This is really bizarre - this tiny change shouldn't cause so many problems, many of the tests are failing. Isn't this maybe a more general problem?

hechth:
@julianpollmann and @florian-huber seems like the problem in the build action was fixed in the dependencies: https://github.com/matchms/matchms/actions/runs/15818694787/job/44583971878

PR #814, opened 2025-06-26
Fixed metadata export

### LINKED ISSUE #813: metadata export is broken
The metadata export doesn't export the metadata and the tests didn't catch this behaviour.

hechth:
Closes #813

Somehow this slipped away, the metadata export function didn't work, I changed some of the assigned datatypes in the array and fixed the tests. Also, the keys now get sorted, so that the returned array should be always the same


hechth:
link to CI build action: https://github.com/matchms/matchms/actions/runs/15902505802

matchms/exporting/metadata_export.py
```diff
@@ -90,8 +90,8 @@ def export_metadata_as_csv(
 
 
 def _subset_metadata(
-    include_fields: List[str], metadata: np.array, columns: Set[str]
-) -> Tuple[np.array, Set[str]]:
+    include_fields: list[str], metadata: np.array, columns: list[str]
```
julianpollmann, line 95: other type hints in this file are using `typing`'s _List_, _Dict_ ,.... .I guess we should switch to built-in types _list_ for the other methods to make it more consistent. But maybe in a different PR?

julianpollmann (commented):
Looks good. Tests for `export_metadata_as_csv` were failing on windows. A newline was added after each row. I pushed a fix.

Btw. What's the use case of [appending the metadata (a+)](https://github.com/matchms/matchms/blob/hechth/issue813/matchms/exporting/metadata_export.py#L87) and not just writing (w) it?

niekdejonge (approved):
Looks good to me as well. I didn't realize we had a metadata export, so good to know! :)

matchms/exporting/metadata_export.py
niekdejonge: Using type hints like this is only supported for python 3.10+. But I think other things already break compatibility with 3.9, so I don't think it is okay indeed to switch towards this other form of type hinting.

julianpollmann: Yes, matchms requires ">=3.10,<3.14". I'll quickly add this.

PR #825, opened 2025-07-24
Add processing report to pipeline again

niekdejonge:


niekdejonge:
@julianpollmann Could you quickly check that this is an okay change to make. I will already merge, since I can imagine you are on holidays and might therefore miss this, but if you foresee any issues, please let me know and we can revert the changes.

julianpollmann:
Hey @niekdejonge, as I remember we removed that from the pipeline as default since it requires cloning spectra which will slow down filtering. I can look in that tomorrow or monday.

julianpollmann (approved):

PR #827, opened 2025-08-15
FlashSimilarity for faster flash entropy, cosine, modified cosine computation

florian-huber:
New fast similarity score computation inspired by the Flash Entropy from Li and Fiehn (2023).

This is based on first creating a m/z index (`_LibraryIndex`) that contains ALL fragments of the dataset ordered by m/z. This allows much faster computation than, for instance, `CosineGreedy` or `ModifiedCosine`.

In the current implementation it is possible to chose differnt score types:
- `score_type` can be `spectral entropy` or `cosine`
- `matching_mode` can be `fragment`, `neutral_loss` or `hybrid` (hybrid being a combination of both fragments and neutral losses)

The computation of modified cosine scores is then a combination of `cosine` with `hybrid` matching mode.

I quickly compared `CosineGreedy` with `FlashSimilarity` (cosine mode) and blink (see #829 ). This shows that FlashSimilarity is not scaling as well as Blink for the presented task (all-vs-all comparisons), but is MUCH faster than CosineGreedy, by about 100x. In addition, and unlike Blink, FlashSimilarity is no approximation but computes the exact same scores as CosineGreedy.

image

`FlashSimlarity` now also allows to compute the modified cosine scores (exact same as with current `ModifiedCosine`) about 100x faster.
image


matchms/similarity/FlashSimilarity.py
```diff
@@ -0,0 +1,921 @@
...
+                                    dtype=self.dtype)
+        return cleaned, (None if pmz is None else float(pmz))
+
+    def pair(self, reference: SpectrumType, query: SpectrumType) -> np.ndarray:
+        """
+        Compute Flash similarity for a single (reference, query) pair.
+        Uses the same preprocessing and scoring logic as the matrix path, but builds a tiny
+        1-spectrum library from the query.
+        
+        Careful: This is not the fast intended use; better .matrix() instead.
```
julianpollmann, line 127: Should we include this as a logged info()?

julianpollmann: logger = logging.getLogger("matchms")

logger.warning("This is not the fast intended use; better .matrix() instead.")

julianpollmann (approved):

PR #829, opened 2025-08-25
Add BLINK score implementation

florian-huber:
This is to have fast cosine estimates in matchms and is based on [Harwood et al. (2023)](https://www.nature.com/articles/s41598-023-40496-9), but in an implementation that differs from the [original GitHub implementation](https://github.com/biorack/blink/blob/main/blink/matrix_manipulation.py).

In terms of runtime both implementations are close. The original seems to still be a bit faster, but I considered that to not be worth spending too much time on (at least not to me right now).
image

One consequence of the different implementation is that the rounding is not the same for the tolerance windows. The original BLINK, say with a tolerance of `0.01` was an approximation of the `CosineGreedy` with tolerance < 0.01 (which is why the authors used `tolerance=0.0099`. The implementation in this PR, however, sticks to the same tolerance so it would approximate CosineGreedy with a `tolerance=0.01`.


matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+        When array_type='sparse', drop scores < sparse_score_min. Default 0.0.
+    """
+
+    is_commutative = True
+    score_datatype = [("score", np.float32), ("matches", "int")]
+
+    def __init__(
+        self,
+        tolerance: float = 0.01,
+        bin_width: float = 0.001,
```
julianpollmann, line 113: not sure if we should set 0.0001 as default?

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+        prepped_refs = [self._prep_spectrum(s) for s in references]
+        prepped_qrys = [self._prep_spectrum(s) for s in queries]
+
+        # Early exit if any side empty
+        n_ref = len(prepped_refs)
+        n_qry = len(prepped_qrys)
+        if n_ref == 0 or n_qry == 0:
+            if array_type == "numpy":
+                return np.zeros((n_ref, n_qry), dtype=np.float32)
+            return coo_matrix((n_ref, n_qry), dtype=np.float32)
```
julianpollmann: maybe we could use recent _coo_array_ api?
`return coo_array((n_ref, n_qry), dtype=np.float32)`
See https://docs.scipy.org/doc/scipy-1.16.2/reference/sparse.migration_to_sparray.html#migration-to-sparray

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+
+        # Collect global bin range
+        all_bins_list = [b for (b, _, _) in prepped_refs if b.size] + [b for (b, _, _) in prepped_qrys if b.size]
+        if not all_bins_list:
+            if array_type == "numpy":
+                return np.zeros((n_ref, n_qry), dtype=np.float32)
+            return coo_matrix((n_ref, n_qry), dtype=np.float32)
```
julianpollmann: same here

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+            batch = prepped_qrys[j0:j1]
+            IQ_blur = self._build_blurred_csr(batch, n_rows, offset, self._R)  # blurred queries (rows x B)
+
+            # scores_batch = I_ref.T @ IQ_blur  -> shape (n_ref, B)
+            scores_batch_sparse = I_ref.T @ IQ_blur
+
+            if array_type == "numpy":
+                S[:, j0:j1] = scores_batch_sparse.toarray()
+            else:
+                scores_batch_sparse = scores_batch_sparse.tocoo()
```
julianpollmann: using array api:
scores_batch_sparse = scores_batch_sparse.asformat("coo")

https://docs.scipy.org/doc/scipy-1.16.2/reference/generated/scipy.sparse.csr_array.asformat.html#scipy.sparse.csr_array.asformat

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+                rows = np.concatenate(sparse_rows) if len(sparse_rows) > 1 else sparse_rows[0]
+                cols = np.concatenate(sparse_cols) if len(sparse_cols) > 1 else sparse_cols[0]
+                data = np.concatenate(sparse_data) if len(sparse_data) > 1 else sparse_data[0]
+            else:
+                rows = np.array([], dtype=np.int32)
+                cols = np.array([], dtype=np.int32)
+                data = np.array([], dtype=np.float32)
+            if self.clip_to_one and data.size:
+                np.minimum(data, 1.0, out=data)
+            return coo_matrix((data, (rows, cols)), shape=(n_ref, n_qry), dtype=np.float32)
```
julianpollmann: same here

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+        row_indices = []
+        data = []
+        for j, (bins, vals, _cnts) in enumerate(prepped_list):
+            if bins.size == 0:
+                continue
+            row_indices.append(bins + offset)
+            col_indices.append(np.full(bins.size, j, dtype=np.int32))
+            data.append(vals.astype(np.float32, copy=False))
+        if not row_indices:
+            return csr_matrix((n_rows, len(prepped_list)), dtype=np.float32)
```
julianpollmann: same here

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+        rows = np.concatenate(row_indices)
+        cols = np.concatenate(col_indices)
+        dat = np.concatenate(data)
+        return csr_matrix((dat, (rows, cols)), shape=(n_rows, len(prepped_list)), dtype=np.float32)
```
julianpollmann: same here

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+            base_rows = bins + offset
+            rows_exp, data_exp = self._expand_column_blur(base_rows, vals, R, n_rows)
+            if rows_exp.size == 0:
+                continue
+            rows_all.append(rows_exp)
+            cols_all.append(np.full(rows_exp.size, j, dtype=np.int32))
+            data_all.append(data_exp)
+
+        if not rows_all:
+            return csr_matrix((n_rows, len(prepped_list)), dtype=np.float32)
```
julianpollmann: same here

matchms/similarity/BlinkCosine.py
```diff
@@ -0,0 +1,544 @@
...
+
+        rows = np.concatenate(rows_all)
+        cols = np.concatenate(cols_all)
+        dat = np.concatenate(data_all)
+        # COO -> CSR automatically sums duplicates (needed when blur windows overlap)
+        return csr_matrix((dat, (rows, cols)), shape=(n_rows, len(prepped_list)), dtype=np.float32)
```
julianpollmann: same here

julianpollmann (approved):

matchms/similarity/BlinkCosine.py
florian-huber: done

matchms/similarity/BlinkCosine.py
florian-huber: done

matchms/similarity/BlinkCosine.py
florian-huber: This was set to `0.001` because that was the value used in the original article (and it worked well in my local tests).

julianpollmann (approved):

florian-huber:
Thanks a lot for reviewing @julianpollmann !

Issue #823, opened 2025-07-24
Pubchempy smiles lookup is broken.


niekdejonge:
Pubchempy smiles lookup is broken. 

This is not a matchms issue, but an issue in pubchempy: https://github.com/mcs07/PubChemPy/issues/101
It does not really break functionality, since inchi is still correctly retrieved. And the derive smiles from inchi filter still retrieves the smiles.

niekdejonge:
A temporary fix with #824 by changing the test to test for inchi. It is not a real issue, since smiles can be retrieved from inchi. The real fix has to be made in pubchempy

julianpollmann:
@niekdejonge, yes I'm aware of that. There is a PR ([104](https://github.com/mcs07/PubChemPy/pull/104)) on pubchempy which resolves that, however not merged till now.

julianpollmann:
@niekdejonge @florian-huber the corresponding PubChempy code has been [updated](https://github.com/mcs07/PubChemPy/commits/main/), but there is no new release (and the last release is from 2017). I emailed the author and asked for an update. Should we wait or just add the github repo as dependency?
`pubchempy= { git = "https://github.com/mcs07/PubChemPy.git" }`

niekdejonge:
@julianpollmann Ah good point, I did not notice that before. If using the github repo as dependency works well, I think that is the best solution. Is there a way to do any versioning in that case?

julianpollmann:
We could pin to the current commit till a new Version is released.

niekdejonge:
Sounds good to me!

PR #831, opened 2025-08-26
Remove pillow + bump rdkit to 2025.3.5

julianpollmann:


julianpollmann:
Fixes #823
Makes #817 obsolete

@niekdejonge @florian-huber any restrictions on the rdkit Version?

niekdejonge:
@julianpollmann Not that I can remember from the top of my head. I noticed you increased the minimum rdkit to a higher version. If this is required for the current matchms please go ahead. If not, I would recommend keeping it at the 2024 setting to give others implementing matchms more flexibility for their dependensies (e.g. if they need older rdkit or have other dependencies not compatible with 2025 version)

Adafede:
@julianpollmann, @niekdejonge  
Reading the changelog: https://github.com/matchms/matchms/releases/tag/0.31.0 is confusing as actually rdkit was not bumped to 2025? 😕

florian-huber:
Oh. I guess that's my fault. I made the mistake of using the auto-generate release notes. Which, as it appears, are only simply using the PR titles :(

Adafede:
> Oh. I guess that's my fault. I made the mistake of using the auto-generate release notes. Which, as it appears, are only simply using the PR titles :(

Not a big problem, and already fixed, wow 🚀

Issue #832, opened 2025-08-26
Add filter to remove MS2 peaks > precursor-m/z


florian-huber:
Most tests show that removing the precursor signal from the MS2 signals helps to get better similarity scores (see, for instance [Li et al. (2021)](https://www.nature.com/articles/s41592-021-01331-z)). `matchms` already has the `remove_peaks_around_precursor_mz` filter.

But, others also added the removal of peaks higher than the precursor-m/z, such as [Li et al. (2023)](https://pmc.ncbi.nlm.nih.gov/articles/PMC11511675/). For this we should also add a `remove_peaks_above_precursor_mz`.

PR #833, opened 2025-08-26
New filter to remove peaks around and above precursor m/z

florian-huber:
- Added new filter to address #832 that will remove all peaks > precursor-m/z + offset_to_precursor.
- Added this filter to the default pipeline `CLEAN_PEAKS` (since I expect that this is usually desired for most scores)

- also did some minor changes to `remove_peaks_around_precursor_mz` (though I also would propose to later drop that --> see #835)


matchms/filtering/peak_processing/remove_peaks_around_precursor_mz.py
```diff
@@ -31,14 +31,18 @@ def remove_peaks_around_precursor_mz(
...
+    if not isinstance(precursor_mz, (float, int)):
+        raise ValueError(
+            "Expected 'precursor_mz' to be a scalar number.",
+            "Consider applying 'add_precursor_mz' filter first."
+            )
     assert mz_tolerance >= 0, "mz_tolerance must be a positive scalar."
 
     mzs, intensities = spectrum.peaks.mz, spectrum.peaks.intensities
+    # TODO: the following line is not very robust (e.g., float precision issues)
     peaks_to_remove = (np.abs(precursor_mz - mzs) <= mz_tolerance) & (mzs != precursor_mz)
```
julianpollmann, line 46: Maybe we could use np.isclose over here? Not quite sure, which tolerance to set.
peaks_to_remove = (np.abs(precursor_mz - mzs) <= mz_tolerance) & (~np.isclose(mzs, precursor_mz, atol=1e-4))

julianpollmann (approved):

matchms/filtering/peak_processing/remove_peaks_around_precursor_mz.py
florian-huber: Good idea, we could decide on this in #835

florian-huber:
Thanks for reviewing @julianpollmann !

PR #852, opened 2026-01-21
Allow for other tools than `poetry`

Adafede:
Dear matchms team,

I recently tried to bump dependencies of matchms to support python 3.14 locally but it seems currently impossible because of `pynndescent`...

While doing so, I realized the current `pyproject.toml` contains `poetry`-specific markers which I rarely use.
Therefore, I tried to make it more generic and bump some dependencies that could be without managing to reach 3.14 but maybe soon?

Feel free to cherry pick anything useful! 😉

PR #853, opened 2026-01-21
ruff

Adafede:
Dear matchms team,

While working on #852, I finished running `ruff` as suggested in the `README` and was not expecting it to change so many files but well...here they are :sweat_smile:


julianpollmann:
Hey @Adafede thanks for your work! To avoid jeopardizing other work, we want to postpone the reformat of the entire codebase right now. If you have small feature commits, could you simply format those files? That way, we can better distinguish between commits. Thanks!

Adafede:
Oh, feel free to simply close...I did run it on the whole codebase without other changes so nothing big

florian-huber:
This will be picked up again in #896 ! 
(will close this one, though)

PR #854, opened 2026-01-21
Do not use TKinter in tests

Adafede:
Should fix  remaining failing test in #853

### DIFF
--- a/tests/plotting/test_spectrum_plots.py
+++ b/tests/plotting/test_spectrum_plots.py
@@ -2,11 +2,11 @@ import matplotlib
 import numpy as np
 import pytest
 
-
-matplotlib.use("Agg")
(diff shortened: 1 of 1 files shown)


tests/plotting/test_spectrum_plots.py
```diff
@@ -1,5 +1,9 @@
+import matplotlib
 import numpy as np
 import pytest
+
+
+matplotlib.use("Agg")
```
julianpollmann: Thanks for the fix!

Could you move this line after imports?
Otherwise `from matplotlib import pyplot as plt` will still import TKinter and tests are failing on Windows.

Adafede: Done 🙂

PR #862, opened 2026-03-02
Exact modified hungarian

### LINKED ISSUE #861: Exact modified cosine
At this time, the library provides a class called "ModifiedCosine". This class would me more appropriately called "ModifiedGreedyCosine", since it is an approximation and does not solve an assignment problem. It would therefore be desirable to also have an actually exact "ModifiedCosine", so to provide users with a choice and have a reasonable ground truth benchmark. Currently [doing such benchmarks](https://github.com/earth-metabolome-initiative/mass-spectrometry-benchmarks), I will share them once I am more confident in their quality.

1. Rename ModifiedCosine to something that makes it clear it is an approximation, like ModifiedGreedyCosine
2. Implement the actual ModifiedCosine

I can make a PR adding it when I have time.

LucaCappelletti94:
This PR splits the existing (approximate) ModifiedCosine implementation into two explicitly named variants:

- ModifiedCosineGreedy (approximate matching, same behavior as the old ModifiedCosine)
- ModifiedCosineHungarian (new exact matching via Hungarian assignment)

### User-facing changes

- Removed matchms.similarity.ModifiedCosine
- Added matchms.similarity.ModifiedCosineGreedy
- Added matchms.similarity.ModifiedCosineHungarian
- Pipeline score step modifiedcosine is no longer accepted; use:
    - modifiedcosinegreedy
    - modifiedcosinehungarian

### Reference exactness

- Added reference-based tests for ModifiedCosineHungarian.
- The expected exact scores in those reference tests were determined using the Rust implementation we use at EMI.

Closes #861


tests/similarity/test_modified_cosine_hungarian_reference.py
```diff
@@ -0,0 +1,76 @@
+import pytest
+from matchms.reference_spectra import aspirin, cocaine, glucose, hydroxy_cholesterol, phenylalanine, salicin
+from matchms.similarity import ModifiedCosineHungarian
+
+
+SCORE_ABS_TOL = 3e-8
+
+REFERENCE_PAIRS = [
+    {"left": "aspirin", "right": "aspirin", "score": 1.000000000000000, "matches": 29},
+    {"left": "aspirin", "right": "cocaine", "score": 0.002140425769044, "matches": 2},
```
florian-huber, line 10: Good to finally have a test for aspirin vs. cocaine.

florian-huber (approved):
Thanks a lot @LucaCappelletti94 
This is a really nice addition to matchms!

Since it includes a minor breaking change (ModifiedCosine --> ModifiedCosineGreedy) I will also do a new release.

PR #864, opened 2026-03-03
Change score compute for matchms 1.0

niekdejonge:
Large refactoring of Scores. 
We removed Scores and the multiple Data Types. Now instead BaseSimilarity computes scores by using matrix or sparse_array, returning or a numpy matrix or a scipy coo_array. We don't have Scores objects that store multiple score types as a stacked sparse array, instead each score is stored separately. 

Filtering on scores now works by creating a mask, which can be used to do efficient compute when passing to BaseSimilarity.matrix or.sparse_array. 

An example run would look like this: 

```python
precursor_mz_match = PrecursorMzMatch()
scores = precursor_mz_match.matrix(spectra, spectra, is_symmetric=False)
scores_mask = ScoresMask.from_matrix(scores, "==", True)
cosine_greedy = CosineGreedy()
cosine_scores = cosine_greedy.sparse_array(spectra, spectra, is_symmetric=False, mask_indices=scores_mask)
```
So scores are not stacked anymore and just store one score type at a time. But if the user wants it can still keep the intermediate scores, but if memory is an issue, after computing the mask intermediate scores can also easily be removed from memory or stored on the disk. 

Still to do:
- Implement issymetric + mask for matrix and sparse implementation (probably just change mask to be symmectri)
- let Pipeline automatically/smartly choose between sparse or matrix, now it is just always a matrix that is computed. It should switch to sparse if mask.coverge < 0.5. But this requires handling of pipeline.scores being coo_array or matrix


matchms/similarity/BaseSimilarity.py
```diff
@@ -1,168 +1,238 @@
...
+            if len(references) != len(queries):
+                raise ValueError(
+                    f"Found unequal number of spectra {len(references)} and {len(queries)} while `is_symmetric` is True"
+                )
+
+            # Compute pairwise similarities for symmetric case
+            for i_ref, reference in enumerate(tqdm(references, "Calculating similarities")):
+                for i_query, query in enumerate(queries[i_ref:], start=i_ref):  # Compute only upper triangle
                     score = self.pair(reference, query)
-                    if self.keep_score(score):
```
florian-huber, line 84: This is an essential part of the sparse logic. Many sparse computations will only make sense if this is defined and used, so we should not simply drop this. Say you wanted to only keep scores for mod.cosine if matches > 2, or scores > 0.2, then the number of stored elements would be reduced drastically (which is intended).

matchms/similarity/BaseSimilarity.py
```diff
@@ -1,168 +1,238 @@
...
-        idx_col = []
-        scores = []
-        # Wrap the outer loop with tqdm to track progress
-        for i_ref, reference in enumerate(tqdm(references[:n_rows],
-                                               desc="Calculating similarities",
-                                               disable=not progress_bar)):
-            if is_symmetric and self.is_commutative:
-                for i_query, query in enumerate(queries[i_ref:n_cols], start=i_ref):
+        sim_matrix = np.zeros((len(references), len(queries)), dtype=self.score_datatype)
+        if is_symmetric:
```
florian-huber, line 138: Only is correct if the score is commutative, so must be `if is_symmetric and self.is_commutative:

niekdejonge: Non_commutative scores would be for instance precursor mz difference? But, yes good point we can of course add this again

florian-huber:
This was taken as a starting point for #878 and has now become redundant.

Issue #866, opened 2026-03-04
Make modified cosine variants consistent when precursor delta is within tolerance


LucaCappelletti94:
As per chat with @Adafede, when `abs(precursor_mz_ref - precursor_mz_query) <= tolerance`, modified cosine variants should return the same score (within numeric epsilon) as the non-modified variants, with stable behavior at the exact tolerance boundary.

This would mean basically dispatching a call to the non-modified variants in the `Modified*` classes when we find ourselves in the aforementioned situation.

Adafede:
@LucaCappelletti94 Weirdly merging did not close the issue?

PR #867, opened 2026-03-04
Fixed modified cosine behavior for precursor deltas within tolerance

LucaCappelletti94:
This PR fixes issue #866 where `ModifiedCosineGreedy` and `ModifiedCosineHungarian` could behave differently from regular cosine even when precursor m/z differences were within tolerance (including boundary cases). We now compute the precursor shift once and, when `|precursor_mz_ref - precursor_mz_query| <= tolerance`, delegate directly to `CosineGreedy`/`CosineHungarian` so behavior is consistent with the expected unshifted case; dedicated tests were added for zero, positive-boundary, and negative-boundary deltas to confirm parity.

### DIFF
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -139,6 +139,9 @@ jobs:
       - name: Create conda environment
         uses: conda-incubator/setup-miniconda@v3
         with:
+          miniforge-variant: Miniforge3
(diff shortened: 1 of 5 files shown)


florian-huber (approved):
Nice!
That is a corner case we haven't looked at yet. Thanks @LucaCappelletti94 !

PR #868, opened 2026-03-05
Add ComputeScores

niekdejonge:
The Pipeline class is quite complex at the moment, where a lot of functionality is mixed. I have now created a separate class ComputeScores. Which does nothing new, but it is just all the score compute related steps of Pipeline, which I think can best be separated out like this. So no breaking changes here.


florian-huber:
Has meanwhile be implemented in a related by differnt manner, see #909

Issue #869, opened 2026-03-06
sketching future matchms API


florian-huber:
We have been working on multiple open ends for a while (including during the matchms workshop in 2025).

We are slowly approaching some common ground here, so this issue is (A) for me to get things clearer ;) and (B) for discussing open questions and checking overlooked issues and inconsistencies.

The new matchms design ("matchms 1.0") will include:
- `SpectraCollection` - think of this as the pandas DataFrame within matchms. This (and not `Spectrum`!) will become the central data type to work with, which makes sense because most people work with MANY spectra and not single ones.

At the same time, we will simplify or remove elements that are perceived as difficult/unintuitive to handle:
- `Scores` was designed to be able to do everything possible. Which is why it fails to work for most people. Classical case of overthinking a problem (I will take the blame here ... :) )
- `Pipeline` is conceptually fine, I guess, but the benefit seems not big enough when compared to the somewhat cumbersome handling.

florian-huber:
(1) Let's start with some design flaws.
Currently, we try to achieve a 2x performance boost for score computations for symmetric score matrices. This is fine, but the handling of it is weird (see #773, which now lives on here).
It is connected to some (pre-)historic parts in matchms which stem from the fact that we were often working from library-search-like situations and therefore named the inputs `references` and `queries`.
To me, that doesn't make sense, so we should simplify this. I suggest a workflow like this (independent of how we name that function!):

```python
scores = calculate_scores(spectra_1, spectra_2)  # default = not-symmetric case

# or
scores = calculate_scores(spectra_1)  # here spectra_2 = None, so obviously that will be symmetric
```

florian-huber:
(2) Too much of a beast: `Scores`

We tried to achieve too many things with that class. That meant it was in the end built around what later turned out to be corner cases (--> computing multiple scores in a row to reduce the number of interesting pairs more and more while keeping all relevant results in a maximally memory efficient way).

That led to serious API-design flaws (which we discussed intensively at the matchms workshop).
- retrieving the different stored scores feels weird because we have to guess or lookup their internal storage names like `CosineGreedy_scores`.
- Most of the time, we are not in such a corner case and only care about the final scores of interest. So, no reason to make such a fuss about keeping all previous computations (such as precursor-m/z matches = True/False) in the same object!

We will, therefore, remove `Scores` or replace it by a lightweight class with intuitive API.
(RIP dear sparsestack, one less library I'll have to maintain...)

[this takes up #766 ]

[...]

PR #870, opened 2026-03-06
Refactor Scores, BaseSimilarity, FlashSimilarity

florian-huber:
1) Completely replaces the former `Scores` class!
The new `Scores` class is only a container for the computed scores. As discussed/proposed in #869 it no longer contains the score computation logic, and it no longer keeps large stacks of different scores!
What it does, however, is to "hide" the underlying complexity to the user so that it is more intuitive to work with Scores objects.

```python
scores = CosineGreedy().matrix(spectra_1)  # this will now assume a symmetric all-vs-all

# Slicing
scores["score"][0, :]   # will return one column of Cosine scores
scores[0, :]["score"]   # will do the same

scores[2, :]  # will return a dictionary like {"score": [.....], "matches": [.....]} containing the selected row

scores["matches"]  # will return a Scores object with only the field "matches"
scores["matches"][:, :]  # will return a dense matrix with all matches values
```

2) `BaseSimilarity` class was refactored and now has a variant `BaseSimilarityWithSparse`
This was required to adjust to the new `Scores` class, but also because the sparse implementation had some issues. 
Key methods that remain are `.pair()` and .`matrix()`, but instead of using weird flags in the latter we now have a dedicated additional `.sparse_matrix()` method with more functionalities.
Moreover, there is a fundamental implementation difference between our various scores. Some are "`pair()`"-focuses, so they actually only implement the pair method and then matrix() and sparse_matrix() have to be derived from that.
(e.g. CosineHungarian etc.).
Many other methods have a particular implementation of the `matrix()` method for higher efficiency, such as PrecursorMzMatch or the Fingerprint based scores. Sometimes this makes it highly undesirable to fallback to a sparse implementation around the .pair() method! 

3) Throughout the code, the `references, queries` terminology was changed to `spectra_1, spectra_2`. For the score computations is now also possible to only have one list of spectra, so instead of `modified_cosine.matrix(my_spectra, my_spectra, is_symmetric=True)` it simply becomes `modified_cosine.matrix(my_spectra)`.

4) `FlashSimilarity` was split into `FlashEntropy` and `FlashCosine` with the latter returning both scores and the number of matched peaks.

5) `ParentmassMatch` and `PrecursorMzMatch` are still present, but internally are simply wrappers of the more general `MetadataMatch` .

6) The `Pipeline` class and the `SimilarityNetwork` class had to be completely rewritten to work with the new Scores class and similarity scores changes.

7) `add_fingerprint` was removed! This function was used to add molecular fingerprints to Spectrum objects, which was poor design. It was therefore removed. Instead, the `Fingerprints` class should be used for this, which has now been linked to [`chemap`](https://github.com/matchms/chemap) to avoid having to do maintenance on two packages. This also gives many more options for molecular fingerprint types and settings.

8) Old leftovers were removed. Some, mostly long-outdated, parts were removed (e.g., long-deprecated methods).


matchms/Scores.py
```diff
@@ -1,497 +1,397 @@
...
-        """
-        if name is None and len(self.score_names) > 1 and sort is True:
-            raise IndexError("For sorting, score must be specified")
+@dataclass(frozen=True)
+class ScoresField:
+    """View on one score field.
+    
+    Supports array-like access and comparison operations to create boolean masks.
+    """
+    _scores: "Scores"
```
julianpollmann: use Scores Type

matchms/Scores.py
```diff
@@ -1,497 +1,397 @@
...
+
+    def to_coo(self, field: Optional[str] = None) -> coo_array:
+        field = self._resolve_field(field)
+        value = self._data[field]
+        if self.is_sparse:
+            return value
+        row, col = np.nonzero(value)
+        return coo_array((value[row, col], (row, col)), shape=value.shape)
+
+    def filter(self, mask) -> "Scores":
```
julianpollmann: use Scores Type

matchms/Scores.py
```diff
@@ -1,497 +1,397 @@
...
-        Construct similarity function from its serialized form.
+    def __eq__(self, other):
+        """Element-wise comparison for scalar Scores."""
+        return self._compare_scalar(other, "__eq__")
+
+    def __ne__(self, other):
+        """Element-wise comparison for scalar Scores."""
+        return self._compare_scalar(other, "__ne__")
+
+    def _filter_with_scores_mask(self, mask: ScoresMask) -> "Scores":
```
julianpollmann: use Scores Type

matchms/Scores.py
```diff
@@ -1,497 +1,397 @@
...
+        if mask.shape != self.shape:
+            raise ValueError(f"Mask has shape {mask.shape}, expected {self.shape}.")
+
+        if self.is_sparse and mask.is_sparse:
+            return self._filter_sparse_with_sparse_mask(mask)
+
+        return self._filter_with_dense_mask(mask.to_dense())
+
+    def _filter_sparse_with_sparse_mask(self, mask: ScoresMask) -> "Scores":
```
julianpollmann: use Scores Type

matchms/Scores.py
```diff
@@ -1,497 +1,397 @@
...
+                dtype=bool,
+            )
+            filtered[field] = coo_array(
+                (coo.data[keep], (coo.row[keep], coo.col[keep])),
+                shape=self.shape,
+            )
+
+        return Scores(filtered)
+
+    def _filter_with_dense_mask(self, mask: np.ndarray) -> "Scores":
```
julianpollmann: use Scores Type

florian-huber:
Closing this, is not reopened as #878

PR #871, opened 2026-03-06
Refactor pipeline

niekdejonge:
In #868 Pipeline was already cleaned and split up somewhat.

What makes Pipeline a bit confusing (to me) is that it allows for so many different inputs. In tutorials we have different styles like "str", Callable and than or with a dict or without. 

The reason for this is that we want to be able to save and load as yaml. 
My take here is that we don't really need to allow for loading from text if it is not from a yaml file and that it can be split more for better code readability. 

Making a pipeline would become something like: 

```python
pipeline = Pipeline(similarity_methods_and_masks: ComputeScores([CosineSore(), MaskSettings(">", 4))
        query_filters: SpectrumProcessor(the_filters)
        reference_filters: SpectrumProcessor(the_filters)) 
```

And if you want to save or load from yaml you just run: 
```python
# save 
pipeline.save_as_yaml(yaml_file_name)
#load
pipeline = Pipeline.from_yaml(yaml_file)
```
The loading and saving from yaml handles all the text conversions and allows only text (not callables etc) 
And the creation by users in for instance jupyter notebooks should not use tests, but just classes and filters directly. 

Not sure if we want this, since it would lead to breaking changes. This is just a draft PR to sketch what that would look like, not yet tested if it fully functions. We can clean up the code (and remove bugs) when we decide we want to move that direction.

### DIFF
--- a/matchms/Pipeline.py
+++ b/matchms/Pipeline.py
@@ -2,61 +2,19 @@ import logging
 import os
 from collections import OrderedDict
 from datetime import datetime
-from typing import Callable, Iterable, List, Optional, Sequence, Union
+from typing import List, Optional, Union
 from deprecated import deprecated
 from matchms.filtering.filter_order import ALL_FILTERS
-from matchms.filtering.SpectrumProcessor import FunctionWithParametersType, SpectrumProcessor
+from matchms.filtering.SpectrumProcessor import SpectrumProcessor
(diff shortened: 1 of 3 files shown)


florian-huber:
I took the Pipeline code from here as starting point for #870

florian-huber:
This is now (also by taking parts from this PR here) part of the new development branch.

PR #875, opened 2026-03-11
Linear cosine

LucaCappelletti94:
- Adds `LinearCosine`, a linear-time cosine similarity from the SIRIUS framework. After a one-time peak merge that enforces well-separated m/z gaps (> 2× tolerance), scoring runs in `O(n+m)` via a two-pointer sweep, versus O(n³) for `CosineHungarian`.
- Adds a `pytest-benchmark` suite comparing `matrix()` performance of `CosineHungarian`, `CosineGreedy`, `LinearCosine`, and `FlashSimilarity` on synthetic spectra (sizes 50 / 100 / 200).

## Algorithm

The algorithm has two phases, both implemented as Numba-jitted functions for decent performance.

**Phase 1: Merge close peaks** (`sirius_merge_close_peaks`):
Processes peaks in descending intensity order. Each unconsumed peak adopts its own m/z and sums the intensities of all unconsumed neighbors within a 2× tolerance window. The result is guaranteed to have consecutive m/z gaps > 2× tolerance, meaning each peak can match at most one partner in the other spectrum. Ask @Adafede why this is a good idea in practice. @oolonek might have a different opinion.

**Phase 2: Two-pointer cosine** (`linear_cosine_score`):
Walks both sorted spectra. Matching peaks (|Δm/z| ≤ tolerance) contribute `mz^mz_power · intensity^intensity_power` products to the cosine numerator. Single pass, O(n+m).

The `matrix()` override pre-merges all spectra once (N+M calls) instead of redundantly merging during each of the N×M pairwise comparisons, and exploits `is_symmetric=True` to halve work for all-vs-all comparisons.

## Testing

Reference data in `tests/linear_cosine_reference.json` covers **6 real compounds** (aspirin, cocaine, glucose, hydroxy-cholesterol, phenylalanine, salicin) across **4 parameter sets** (varying tolerance, mz_power, intensity_power), obtained from the Rust impl.

## Benchmarks

Added `pytest-benchmark` as a dev dependency with `addopts = "--benchmark-disable"` so normal test runs are unaffected. Run with:

```bash
uv run pytest tests/similarity/test_bench_matrix.py --benchmark-enable -v
```

Results (single-threaded, `is_symmetric=True`, synthetic spectra with ~18 overlapping peaks per pair):

| Function | 50 spectra | 100 spectra | 200 spectra |
|---|---|---|---|
| **FlashSimilarity** | 4.9 ms | 11.9 ms | 33 ms |
| **LinearCosine** | 7.2 ms | 27 ms | 108 ms |
| **CosineGreedy** | 42 ms | 166 ms | 666 ms |
| **CosineHungarian** | 114 ms | 451 ms | 1,777 ms |

I was expecting the new LinearCosine to be faster than FlashSimilarity, so maybe there is space for improvement.


matchms/similarity/__init__.py
```diff
@@ -20,13 +20,15 @@
...
 """
+
 from .BinnedEmbeddingSimilarity import BinnedEmbeddingSimilarity
 from .BlinkCosine import BlinkCosine
 from .CosineGreedy import CosineGreedy
 from .CosineHungarian import CosineHungarian
 from .FingerprintSimilarity import FingerprintSimilarity
 from .FlashSimilarity import FlashSimilarity
 from .IntersectMz import IntersectMz
+from .LinearCosine import LinearCosine
```
florian-huber: I think that `CosineLinear` would be better here to be more consistent with `CosineGreedy` and `CosineHungarian`.

matchms/similarity/linear_cosine_functions.py
```diff
@@ -0,0 +1,160 @@
...
+        (M, 2) array of merged peaks sorted by ascending m/z.
+    """
+    n = spec.shape[0]
+    if n == 0:
+        return spec.copy()
+
+    merge_window = 2.0 * mz_tolerance
+
+    # Sort indices by ascending intensity and iterate in reverse (descending)
+    order = np.argsort(spec[:, 1])
```
florian-huber: I could imagine that this *might* lead to equal-intensity tie issues, not 100% sure though...

florian-huber: We could also check later and fix if necessary (I could draw this PR to a local branch to try out some stuff...).

matchms/similarity/__init__.py
florian-huber:
```suggestion
from .CosineLinear import CosineLinear
```

matchms/similarity/__init__.py
```diff
@@ -44,6 +46,7 @@
     "FingerprintSimilarity",
     "FlashSimilarity",
     "IntersectMz",
+    "LinearCosine",
```
florian-huber:
```suggestion
    "CosineLinear",
```

matchms/similarity/__init__.py
```diff
@@ -62,12 +65,22 @@ def get_similarity_function_by_name(similarity_function_name: str):
...
+    functions = [
+        BinnedEmbeddingSimilarity,
+        BlinkCosine,
+        ModifiedCosineGreedy,
+        CosineGreedy,
+        CosineHungarian,
+        FingerprintSimilarity,
+        FlashSimilarity,
+        IntersectMz,
+        LinearCosine,
```
florian-huber:
```suggestion
        CosineLinear,
```

florian-huber:
@LucaCappelletti94 thanks a lot for this PR!

I think we should change the name to `CosineLinear`, then I could already merge it.
After merging, I plan to try two things:
- check if equal-intensity ties could be an issue
- Switch to a numba-fied double for-loop in `matrix()` to check if this changes anything about the benchmarking.
(I wouldn't spend too much time on that since we will soon have larger changes to the codebase anyway, see #878 )

LucaCappelletti94:
Ping?

florian-huber (approved):
Looking good! Thanks a lot @LucaCappelletti94 !
Besides the new `CosineLinear`, also the new benchmarking option is a very nice addition to matchms!

PR #878, opened 2026-03-13
Refactor Scores, BaseSimilarity and restructure various scores

florian-huber:
[moved this from #870 (because changing the base branch was no longer possible)]

1) Completely replaces the former `Scores` class!
The new `Scores` class is only a container for the computed scores. As discussed/proposed in #869 it no longer contains the score computation logic, and it no longer keeps large stacks of different scores!
What it does, however, is to "hide" the underlying complexity to the user so that it is more intuitive to work with Scores objects.

```python
scores = CosineGreedy().matrix(spectra_1)  # this will now assume a symmetric all-vs-all

# Slicing
scores["score"][0, :]   # will return one column of Cosine scores
scores[0, :]["score"]   # will do the same

scores[2, :]  # will return a dictionary like {"score": [.....], "matches": [.....]} containing the selected row

scores["matches"]  # will return a Scores object with only the field "matches"
scores["matches"][:, :]  # will return a dense matrix with all matches values
```

2) `BaseSimilarity` class was refactored and now has a variant `BaseSimilarityWithSparse`
This was required to adjust to the new `Scores` class, but also because the sparse implementation had some issues. 
Key methods that remain are `.pair()` and .`matrix()`, but instead of using weird flags in the latter we now have a dedicated additional `.sparse_matrix()` method with more functionalities.
Moreover, there is a fundamental implementation difference between our various scores. Some are "`pair()`"-focuses, so they actually only implement the pair method and then matrix() and sparse_matrix() have to be derived from that.
(e.g. CosineHungarian etc.).
Many other methods have a particular implementation of the `matrix()` method for higher efficiency, such as PrecursorMzMatch or the Fingerprint based scores. Sometimes this makes it highly undesirable to fallback to a sparse implementation around the .pair() method! 

3) Throughout the code, the `references, queries` terminology was changed to `spectra_1, spectra_2`. For the score computations is now also possible to only have one list of spectra, so instead of `modified_cosine.matrix(my_spectra, my_spectra, is_symmetric=True)` it simply becomes `modified_cosine.matrix(my_spectra)`.

4) `FlashSimilarity` was split into `FlashEntropy` and `FlashCosine` with the latter returning both scores and the number of matched peaks.

5) `ParentmassMatch` and `PrecursorMzMatch` are still present, but internally are simply wrappers of the more general `MetadataMatch` .

6) The `Pipeline` class and the `SimilarityNetwork` class had to be completely rewritten to work with the new Scores class and similarity scores changes.

7) `add_fingerprint` was removed! This function was used to add molecular fingerprints to Spectrum objects, which was poor design. It was therefore removed. Instead, the `Fingerprints` class should be used for this, which has now been linked to [`chemap`](https://github.com/matchms/chemap) to avoid having to do maintenance on two packages. This also gives many more options for molecular fingerprint types and settings.

8) Old leftovers were removed. Some, mostly long-outdated, parts were removed (e.g., long-deprecated methods).


matchms/networking/networking_functions.py
```diff
@@ -1,100 +1,168 @@
...
     similars_idx = {}
     similars_scores = {}
-    for i, spec in enumerate(scores.references):
-        spec_id = spec.get(identifier_key)
-        _, c, v = scores.scores[i, :, score_name]
-        idx = np.argsort(v)[::-1][:top_n]
+
+    for i in range(matrix.shape[0]):
+        values = matrix[i, :]
+        order = np.argsort(values)[::-1]
```
julianpollmann: Note sure, but we just need to sort top_n over here?
Maybe change to `order = np.argpartition(values, -top_n, axis=1)[:, -k:]`
Then the actual sorting: `order = order[np.argsort(values[order])[::-1]]`

matchms/networking/networking_functions.py
```diff
@@ -1,100 +1,168 @@
...
     similars_idx = {}
     similars_scores = {}
-    for i, spec in enumerate(scores.queries):
-        spec_id = spec.get(identifier_key)
-        r, _, v = scores.scores[:, i, score_name]
-        idx = np.argsort(v)[::-1]
+
+    for j in range(matrix.shape[1]):
+        values = matrix[:, j]
+        order = np.argsort(values)[::-1]
```
julianpollmann: same here?

matchms/similarity/BaseSimilarity.py
```diff
@@ -1,168 +1,518 @@
+# Subclassing guide:
```
julianpollmann, line 1: Should we leave this over here? Maybe move to Class Docstring?

matchms/similarity/BaseSimilarity.py
```diff
@@ -1,168 +1,518 @@
...
-from sparsestack import StackedSparseArray
+import numpy.typing as npt
+from scipy.sparse import coo_array
 from tqdm import tqdm
-from matchms.typing import SpectrumType
+from matchms.Scores import Scores
+from matchms.typing import ScoreFilter, SpectrumType
 
 
 class BaseSimilarity:
```
julianpollmann: class BaseSimilarity(ABC):

matchms/similarity/BaseSimilarity.py
```diff
@@ -1,168 +1,518 @@
+# Subclassing guide:
+# - implement pair()
+# - optionally define score_datatype and score_fields
+# - optionally overwrite keep_score() for default sparse filtering
+# - optionally overwrite matrix() for performance optimizations
+# - for scores that also provide a sparse score compuation use BaseSimilarityWithSparse
+#  and optionally overwrite sparse_matrix() for performance optimizations
+# - users can also pass score_filter=... to sparse_matrix()
+
 from abc import abstractmethod
```
julianpollmann: from abc import ABC, abstractmethod

matchms/similarity/Cosine.py
```diff
@@ -0,0 +1,130 @@
...
+        if self.use_hungarian:
+            cosine = CosineHungarian(
+                tolerance=self.tolerance,
+                intensity_power=self.intensity_power,
+            )
+        else:
+            cosine = CosineGreedy(
+                tolerance=self.tolerance,
+                intensity_power=self.intensity_power,
+            )
```
julianpollmann, line 72: We could log (info), which Cosine variant will be used. Especially, since pair uses CosineGreedy and matrix uses FlashCosine by default.
If not logger = logging.getLogger("matchms") can be removed

matchms/similarity/FingerprintSimilarity.py
```diff
@@ -1,167 +1,253 @@
...
+        raise NotImplementedError(
+            "FingerprintSimilarity.pair() is not supported. "
+            "Use matrix(...) with spectra or Fingerprints objects instead."
+        )
+
+    def matrix(
+        self,
+        spectra_1: Optional[Sequence[SpectrumType]] = None,
+        spectra_2: Optional[Sequence[SpectrumType]] = None,
+        *,
```
julianpollmann: This * could break compatibility when expecting a generic BaseSimilarity Type. Do we need it?

julianpollmann:
@florian-huber should we adjust all score dtypes (at least floats) to be np.float64 or np.float32? [BaseSimilarity](https://github.com/matchms/matchms/blob/refactor_scores_v2/matchms/similarity/BaseSimilarity.py#L43) and some scores have double precision while others e.g., [BlinkCosine](https://github.com/matchms/matchms/blob/refactor_scores_v2/matchms/similarity/BlinkCosine.py#L112) have single precision.

matchms/similarity/ModifiedCosine.py
```diff
@@ -0,0 +1,136 @@
...
+                spectra_2=spectra_2,
+                score_fields=score_fields,
+                progress_bar=progress_bar,
+            )
+        modcos = FlashCosine(
+            matching_mode="hybrid",
+            tolerance=self.tolerance,
+            intensity_power=self.intensity_power,
+            )
+
```
julianpollmann, line 129: We could log which cosine was used.

matchms/Fingerprints.py
```diff
@@ -63,394 +87,324 @@ class Fingerprints:
...
+        return len(self._inchikeys)
 
     @property
     def to_dataframe(self) -> pd.DataFrame:
+        """Return fingerprints as a pandas DataFrame indexed by InChIKey."""
+        if self._fingerprints is None:
+            return pd.DataFrame(index=[])
+
+        if sp.issparse(self._fingerprints):
+            fingerprints = list(self._fingerprints)
```
julianpollmann: fingerprints = [row.toarray().flatten() for row in self._fingerprints]

matchms/Pipeline.py
```diff
@@ -1,413 +1,457 @@
...
+    def __init__(
+        self,
+        workflow: OrderedDict,
+        progress_bar: bool = True,
+        logging_level: str = "WARNING",
+        logging_file: Optional[str] = None,
+    ):
+        self._spectra_1: List[SpectrumType] = []
+        self._spectra_2: Optional[List[SpectrumType]] = None
         self.is_symmetric = False
```
julianpollmann: Should we remove is_symmetric over here aswell or keep it for backward compatibility?

matchms/Pipeline.py
```diff
@@ -1,413 +1,457 @@
...
+                cleaned_spectra_file=cleaned_spectra_1_file,
+                create_report=create_report,
             )
             self.write_to_logfile(str(report))
-            if cleaned_reference_file is not None:
-                self.write_to_logfile(f"--- Reference spectra written to {cleaned_reference_file} ---")
+            if cleaned_spectra_1_file is not None:
+                self.write_to_logfile(f"--- Spectra_1 written to {cleaned_spectra_1_file} ---")
+
+        if self.is_symmetric:
```
julianpollmann: Could be removed?

julianpollmann (approved):

matchms/networking/networking_functions.py
florian-huber: Should be faster indeed. I would have to use `np.argpartition(values.reshape(1, -1), -top_n, axis=1)[:, -top_n:][::-1]` I think.

florian-huber: Ah. Problem will only be that diagonal is removed there after, so it should be a +1 added.

matchms/similarity/FingerprintSimilarity.py
florian-huber: done

matchms/networking/networking_functions.py
florian-huber: Done. And thanks for the hint, argpartition is indeed often faster!

matchms/similarity/BaseSimilarity.py
florian-huber: Done. Changes some tests accordingly.

matchms/Fingerprints.py
florian-huber: I kept that part (to allow a sparse format) but realized I could skip the if/else here.

matchms/Pipeline.py
florian-huber: True. Can go.

matchms/Pipeline.py
florian-huber: I removed it now.

matchms/similarity/BaseSimilarity.py
florian-huber: I would postpone this.

matchms/similarity/ModifiedCosine.py
florian-huber: Good idea! But for now, I would postpone this.

florian-huber:
Thanks a lot for reviewing this huge pile of code @julianpollmann !

(I postponed two smaller things and made an issue for the float32 vs float64 question, the rest was changed based on your comments)

Issue #873, opened 2026-03-11
matchms v1.0 planning


florian-huber:
Matchms 1.0 has been in the making for quite some time. And in discussions for even longer...
Here now a list with current items planned for that future release (no worries, we will do a beta first ;-) )

# Key items:
- [x] Introduce a powerful central new data type: `SpectraCollection`
- [x] Replace old `Scores`-monster with more lightweight `Scores` class that only does one thing: being a container for computed scores.
- [x] Refactor `BaseSimilarity` to work with the new `Scores` class and be more flexible (and efficient) for future score designs.
- [x] Refactor all available scores to the new logic and design.
- [x] Make `ParentmassMatch` and `PrecusorMzMatch` wrappers of a more general MetadataMatch
- [x] Introduce *one* central similarity method for `Cosine` and `ModifiedCosine`. All other methods remain, but this will make it easier for users to pick the most obvious choice.

## Nice-to-haves:
- [x] Molecular fingerprints are now still being added to each spectrum object (which is poor design). Compute fingerprints on the fly when needed, or (better) use something like the available `Fingerprint` class and build the fingerprint computations from there. Consider also the use or link to `chemap` to reduce code redundancy.

For current progress see #870

florian-huber:
# matchms < 1.0

See #874 

Add deprecated warnings for
- [ ] `Scores` class in its current form
- [x] `add_fingerprint()`

Issue #688, opened 2024-06-26
Design and implement a Spectra / Dataset class


florian-huber:
Currently large MS/MS datasets are handled as lists of `Spectrum` objects.

We might want to consider an own `Spectra` (or differently named) dataset class with more functionality than the simple lists we now use. This could include:

- Pandas-like easy access to full dataset information (number of spectra, basic statistics etc.)
- Dataset-level functions such as full-dataset filters (for instance: remove duplicates, further metadata harmonization ...)

florian-huber:
Simple example of code I see in many of my scripts:
```python
metadata = {
    "inchikey": [],
    "inchi": [],
    "smiles": [],
    "instrument_type": [],
    "precursor_mz": [],
    "parent_mass": [],
    "spectrum_id": [],
    "collision_energy": [],
    "ionmode": [],
}

for s in tqdm(spectrums):
    for key in metadata.keys():
        metadata[key].append(s.get(key))
        
```

hechth:
I would refer to the implementation of a spectral library how it is done in mzSpecLib - if we consider this as the main source of spectra (which currently is the case, as matchms doesn't extract spectra from other data).

florian-huber:
--> https://github.com/HUPO-PSI/mzSpecLib

florian-huber:
Just brainstorming here:

```
default_fields = ["inchikey", "inchi", ....]


def spectra_loader(Pipeline) --> Spectra:

class Spectra:

    def __init__(self, spectra: list, fields=default_fields):
        self._spectra = spectra  # static?
        self.fingerprints = None
        self._metadata = self.collect_metadata()  # shape(len(spectra), len(fields)

    def collect_metadata(self, fields: list):
          pass

    # Getter / Setter for metadata & spectra

    def sort_values(self, ...):
          self._metadata.sort...  
          self.spectra   # sort

    def query(self)....
          self._metadata.query....

    def get_compound_annotations() -> pd.DataFrame (inchikey, smiles, inchi, parent mass):
        """All unique inchikeys with respective most common smiles and inchi..."""

    def get_unique_compound_spectra(self) --> list(Spectrum):

    def sort(by="smiles", type="occurrence"):

    # export -> mgf, json...
```

niekdejonge:
In another project I did write SpectrumSet classes. This was mostly to make it easier to work on a molecule/inchikey level. 

For me it resolved: 
[...]

PR #879, opened 2026-03-19
Add spectra collection

julianpollmann:
First draft of a SpectraCollection as discussed in #869, #873 and #688 when working with multiple Spectra.
This encapsulates Fragments (as csr_array) and Metadata (as DataFrame) and enables:
* faster peak processing due to sparse implementation
* synchronized Fragment and ;etadata access (e.g., removing/filtering Spectra will remove both Fragment and Metadata entry)
* sorting over either Fragment attributes (e.g., intensity sum) or Metadata keyword entries/columns
* filtering of many Spectra using masks for Fragment attributes or Metadata entries
* computing Spectra hashes and remove duplicates
* displaying minimal Fragment statistics (Peak counts, Intensity Sums and Shannon Entropy based on Intensities)
* Pandas-like logic for filtering, sorting and inplace Operations and slicing

### Working with SpectraCollection:
```
spectra = list(load_from_mgf("..."))
spec_coll = SpectraCollection(spectra)

sliced_spectra = spec_coll[:10]
...
```

### Sorting
SpectraCollection can be sorted based on Fragment attributes:
`spec_coll.sort(by="sum", on="fragments")`
or by Metadata:
`spec_coll.sort(by="retention_time", on="metadata", inplace=False)`

### Filtering
SpectraCollection can be filtered based on MetaData entries:
`spec_coll.filter(coll.metadata["ms_level"] == 2)`
By Fragment attributes:
`spec_coll.filter(coll.fragments.sum() > 500)`
or by boolean masks
```
mask = [True, False, ...]
spec_coll.filter(mask)
```

Still TBD:

- [ ] Advanced filtering pipeline/interface to filter many Spectra efficiently
- [ ] Adjust Importing Spectra to directly import into SpectraCollection (right now SpectraCollection gets a list of Spectrum Objects)


florian-huber:
I started working with this on some actual data. So I will here simply note some ideas/observations:

## Major
- I think we should not keep the fragment data as plain CSR in `_fragments`, but instead create a `FragmentCollection` class to take care of the fragment data handling
- more mz stability would be better. Right now I get back float64 mz values (even though `Spectrum` itself uses float32)
```python
mz_initial = 101.111178 

bin_int = spectra.mz_to_bin(mz_initial)
mz_from_bin = spectra.bin_to_mz(bin_int)
bin_int, mz_from_bin, mz_from_bin
```
BTW: that also is a good argument for why CSR is probably "exact enough", because float32 seemed to be fine so far and the binning here stays in that range.

## Minor
- `__repr__()` should give back something more meaningful, and not call it list

julianpollmann:
@florian-huber ok, thanks for the feedback.
I'll replace FragmentProxy with a FragmentCollection Class.

Note quite sure, but Spectrum Class doesn't specify the precision right now (can be float32 and float64) depending on the input, e.g.:
```
spectra = list(load_from_mgf("tests/testdata/pesticides.mgf"))
spectra[0].mz.dtype -> dtype('float64')
mzs_orig = [spec.mz for spec in spectra]

Casting mz_from_bin to float32 will lead to different mzs:
bin_size: float = 0.000001
def mz_to_bin(mz: np.ndarray | float) -> np.ndarray:
    return np.floor(mz / bin_size).astype(np.int64)

# Adding half the bin size here to treat spectra at the lower and upper bin ends equally
def bin_to_mz(bin_idx: np.ndarray | int) -> np.ndarray:
    return (bin_idx * bin_size) + (bin_size / 2)

def bin_to_mz_float32(bin_idx: np.ndarray | int) -> np.ndarray:
    return ((bin_idx * bin_size) + (bin_size / 2)).astype(np.float32)

bins = mz_to_bin(mzs_orig[0])
mz_float64 = bin_to_mz(bins)
mz_float32 = bin_to_mz_float32(bins)

mzs_orig[0][0] -> 70.786774 # float64
bins[0]        -> 70786774 # int64
mz_float64[0]  -> 70.7867745 # float64
mz_float32[0]  -> 70.78677 # float32
```

florian-huber:
Ah, OK, so we should somewhere set the mz precision I guess.
I am in parallel trying some refactoring ideas in #881 (just fyi to avoid double work @julianpollmann ).

florian-huber (approved):

florian-huber:
Thanks a lot @julianpollmann 
As you pointed out, there are a lot of things to be done (filtering pipeline, importing, but also scores adjustments). I would, however, do this in a separate PR and already merge this one!

PR #891, opened 2026-05-11
Ignore mass shifts < tolerance for Flash-style computations

florian-huber:
Recently, a small change was added to `ModifiedCosineGreedy` and `ModifiedCosineHungarian` to ignore precursor-m/z shifts when they are `< tolerance`  because that would often lead to undesired results.

This is now also implemented for the `FlashCosine` similarity implementation.


florian-huber:
Failing test seems not related to this code change.

PR #899, opened 2026-05-28
Add support for Python 3.14

BioGeek:
Adds Python 3.14 to the supported versions for matchms.

`pynndescent` [now supports Python 3.14](https://github.com/lmcinnes/pynndescent/commit/42fd38964368446eb0e26346d29f3066a449a39e) so this is no longer a blocker for `matchms`. The next blocker turned out to be `sparsestack==0.7.0`, which pins `numba>=0.61.0,<0.62.0` and `numba` 0.61.x ships no Python 3.14 wheels (and `llvmlite` 0.44.0 fails to build from source on 3.14 due to the distutils.spawn(dry_run=...) removal).

  Changes

  - pyproject.toml
    - Widened `requires-python to >=3.10,<3.15` and added the 3.14 trove classifier.
    - Bumped `scipy>=1.15.3,<1.16` → `scipy>=1.15.3,<1.17`. scipy 1.15.x has no 3.14 wheels; 3.14 support starts in 1.16.
    - Added `[tool.uv.sources]` pointing sparsestack at [BioGeek/sparsestack@python3.14 ](https://github.com/BioGeek/sparsestack/tree/python3.14) as a temporary workaround. I will revert this block once https://github.com/matchms/sparsestack/pull/31 is merged and a new sparsestack release is published on PyPI.
  - uv.lock: refreshed. Notable upgrades on the 3.14 fork: `numba` 0.61.2 → 0.65.1, `llvmlite` 0.44.0 → 0.47.0, `scipy` 1.15.3 → 1.16.3.

PR #901, opened 2026-06-01
Add support for Python 3.14 (targeting `development` branch)

BioGeek:
## Summary

Adds Python 3.14 to the supported versions on the `development` branch.

This is an alternative to #899 (which targets `master` and needs a temporary `[tool.uv.sources]` override for sparsestack while matchms/sparsestack#31 lands). Since `development` has already dropped sparsestack, no workaround is needed here — the three-line change is enough.

## Changes

- `pyproject.toml`
  - `requires-python = ">=3.11,<3.14"` → `">=3.11,<3.15"` (the `# blocked because of pynndescent` comment is removed — pynndescent 0.6.0's own constraints are lax: `numba>=0.55`, `llvmlite>=0.38`; the historical blocker was sparsestack pinning `numba<0.62`, which `development` no longer depends on).
  - Added `Programming Language :: Python :: 3.14` classifier.
  - `scipy>=1.15.3,<1.16` → `scipy>=1.15.3,<1.17` (scipy 1.15.x has no 3.14 wheels; support starts in 1.16).
- `uv.lock`: refreshed. Notable upgrades on the 3.14 fork: `numba 0.61.2 → 0.65.1`, `llvmlite 0.44.0 → 0.47.0`, `scipy 1.15.3 → 1.16.3`.

## Test plan

- [ ] `uv sync --group dev` succeeds on Python 3.14 (verified locally on 3.14.3)
- [ ] `uv run pytest` passes locally
- [ ] CI passes on 3.11, 3.12, 3.13, and 3.14

cc @florian-huber — per your comment on #899, opening this against `development` so we don't need the sparsestack fork override.


BioGeek:
Marking this as draft — CI exposed a second blocker.

The `development` branch dropped sparsestack (good), but added [`chemap`](https://github.com/matchms/chemap) as a dependency, and chemap 0.3.6 has `requires-python = ">=3.11,<3.14"`. So the build fails with:

```
chemap requires Python <3.14,>=3.11, so it will not be installable for Python >=3.14,<3.15
```

chemap is imported at module level in `matchms/Fingerprints.py` and `matchms/similarity/FingerprintSimilarity.py`, so it can't be made conditional via a marker.

I've opened matchms/chemap#16 to bump chemap's Python cap to `<3.15`. All of chemap's runtime deps already ship 3.14 wheels, so the cap looks stale rather than a real constraint.

Once that lands and a new chemap is released on PyPI, I'll refresh the lockfile here and take this PR out of draft. /cc @florian-huber

florian-huber:
Hi @BioGeek ,

seems like a longer road through the dependency trees 😄 .
I now did new releases of `chemap` (and `sparsestack`) with your changes.
But now tests here seem to fail due to `scikit-fingerprints` which is indeed not yet running with Python 3.14 (https://github.com/MLCIL/scikit-fingerprints).

But since using molecular fingerprints in matchms is only required for a subset of the functionalities, we could consider making the entire `chemap`--> `scikit-fingerprints` dependency branch optional.

(On the scikit-fingerprint side there is an issue for this, too --> https://github.com/MLCIL/scikit-fingerprints/issues/496)

florian-huber:
Outside the `development` branch, the switch to Python 3.14 seems to work, though --> was now released as `matchms 0.33.1`

BioGeek:
Thanks!

The `scikit-fingerprints` issue https://github.com/MLCIL/scikit-fingerprints/issues/496 mentions that it is blocked by the `cloudpickle` which is vendored inside `Joblib`. 

Joblib recently unvedored cloudpickle https://github.com/joblib/joblib/pull/1775 and I created a PR against cloudpickle to add Python 3.14 support https://github.com/cloudpipe/cloudpickle/pull/594.

So once those two dependencies create new releases,  `scikit-fingerprints` should support Python 3.14 as well. I created a draft PR https://github.com/MLCIL/scikit-fingerprints/pull/620  with temporary workarounds in the meantime.

PR #917, opened 2026-06-22
Add overload statements

niekdejonge:
Adds overloads so type hinting works well when getting items from SpectrumCollections.


florian-huber:
I would close this since it will be solved differently in #909

Issue #921, opened 2026-06-29
mzml loading ignores most metadata


niekdejonge:
I never realized this was the case, since I mostly worked with mgf. But I now had some mzml files and noticed that only a few metadata fields are actually parsed. These are "charge", "scan_number","title", "precursor_mz", "scan_start_time", "retention_time". All the rest of the metadata is silently ignored. 

For the short term, it would be good to add a warning, that mzml reading is a prototype and is not fully supported by matchms. 

Or better if we implement a real mzml loader. At quick glance this seems quite doable, since it is a very well structured datatype and for each accession of metadata there seems to be a human readable name given. 

@florian-huber Did you know this was the case for mzml and is there something I am missing that makes it challenging to implement a mzml loader that loads all available metadata, instead of a small subset? 

Also related to this issue: #708

niekdejonge:
One quick fix to me seems to just skip the parse_mzml_mzxml_metadata method and to just give all the metadata loaded by pyteomics (with the mz and intensities removed) as metadata. 

So the code would look like: 

```python
from pathlib import Path
from typing import Generator, Union
import numpy as np
from matchms.importing.parsing_utils import sort_by_mz
from matchms.Spectrum import Spectrum
from pyteomics.mzml import read


def load_from_mzml(
    filename: Union[str, Path], ms_level: int = 2, metadata_harmonization: bool = True
) -> Generator[Spectrum, None, None]:
    """Load spectrum(s) from mzml file.

    This function will create ~matchms.Spectrum for every spectrum of desired
    ms_level found in a given MzML file. For more extensive parsing options consider
    using pyteomics or pymzml packages.

    Example:

    .. code-block:: python

        from matchms.importing import load_from_mzml

        file_mzml = "testdata.mzml"
        spectra = list(load_from_mzml(file_mzml))

    Parameters
    ----------
    filename:
        Filename for mzml file to import.
    ms_level:
        Specify which ms level to import. Default is 2.
    metadata_harmonization : bool, optional
        Set to False if metadata harmonization to default keys is not desired.
        The default is True.
    """
    if isinstance(filename, Path):
        filename = str(filename)  # pyteomics does not support pathlib.Path

    with read(filename, dtype=dict) as reader:
        for pyteomics_spectrum in reader:
            if "ms level" in pyteomics_spectrum and pyteomics_spectrum["ms level"] == ms_level:
                mz = np.asarray(pyteomics_spectrum.pop("m/z array"), dtype="float")
[...]

PR #922, opened 2026-06-30
Fix mzml loader

niekdejonge:
Solving #921 

The current mzml loader ignores all metadata that is not one of the 6 specified fields. 
It seemed somewhat easy to parse the data that the pyteomics reader, so that is my attempt here. 

Implemented the following:
1: mzml has hierarchical fields, this doesn't match well with matchms. So I wrote a script that flattens this hierarchy. It will just connect the different hierarchy levels with a "-". There are some cases where multiple entries are allowed. Like multiple precursor m/z, but these are (as far as I know) edge cases matchms doesn't support. 
2. Previously some parsing happened of converting to matchms key names. This happened in a separate method, now I moved it to the standard key conversions list, which also fits better with the general way this is handled in matchms.

So I think it still has all the old functionality now, with in addition the non standardized metadata keys also being loaded. 

@florian-huber What do you think. Do you know anyone that is familiar with using mzml files that could do a quick stress test? I now just tested on some files I had here and for those it worked fine. 

@hechth If I remember correctly, I think you worked with mzml files right? Would this work for you, or would this break your pipelines?

### DIFF
--- a/matchms/filtering/metadata_processing/correct_charge.py
+++ b/matchms/filtering/metadata_processing/correct_charge.py
@@ -50,11 +50,6 @@ def correct_charge(spectrum_in: SpectrumType, clone: Optional[bool] = True) -> O
     elif charge == 0 and ionmode == "negative":
         charge = -1
         logger.info("Guessed charge to -1 based on negative ionmode")
-    if charge == 0:
-        if spectrum.get("polarity") == "-":
-            charge = -1
-        if spectrum.get("polarity") == "+":
-            charge = 1
     # Correct charge when in conflict with ionmode (trust ionmode more!)
     if np.sign(charge) == 1 and ionmode == "negative":
         logger.info("Changed sign of given charge: %s to match negative ionmode", charge)
--- a/matchms/filtering/metadata_processing/correct_charge.py
+++ b/matchms/filtering/metadata_processing/correct_charge.py
@@ -50,7 +50,11 @@ def correct_charge(spectrum_in: SpectrumType, clone: Optional[bool] = True) -> O
     elif charge == 0 and ionmode == "negative":
         charge = -1
         logger.info("Guessed charge to -1 based on negative ionmode")
-
+    if charge == 0:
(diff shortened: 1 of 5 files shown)


matchms/filtering/metadata_processing/correct_charge.py
```diff
@@ -50,7 +50,11 @@ def correct_charge(spectrum_in: SpectrumType, clone: Optional[bool] = True) -> O
     elif charge == 0 and ionmode == "negative":
         charge = -1
         logger.info("Guessed charge to -1 based on negative ionmode")
-
+    if charge == 0:
+        if spectrum.get("polarity") == "-":
+            charge = -1
+        if spectrum.get("polarity") == "+":
+            charge = 1
```
niekdejonge: I now moved it here, since I think this is more in style with the rest of matchms. But it would lead to new behavior of the loader. Since before deriving the charge from the polarity happened directly in the loader. We can of course keep it in the loader as well. I now went for the more matchms style, but moving it into the loader to keep it consistent is also fine with me.

matchms/importing/load_from_mzml.py
```diff
@@ -40,12 +40,53 @@ def load_from_mzml(
...
+    elif isinstance(value, list):
+        if len(value) != 1:
+            raise ValueError("Expected only 1 value, for any mzml with count higher than 1, matchms has no support")
+        for k, v in value[0].items():
+            yield from parse_key_value(k, v, False)
+    else:
+        yield key, value
+
+
+def derive_charge_from_polarity(flattend_metadata):
```
niekdejonge, line 76: This is added here now, since it was before also handled during file loading. But I think it would be more consistent with the rest of matchms if this step happens in the correct_charge filter. But I am afraid I might break someones pipeline in that case, so I placed it here, to keep it backwards compatible.

florian-huber (approved):
Looks very good, thanks a lot @niekdejonge .

One question though: do you have any good additional testdata to expand the tests a bit?

PR #924, opened 2026-07-14
Update from biotools

bh2025project5:
Auto-generated PR from biotools ID matchms.


florian-huber:
This created a `README.md` while we use a `README.rst`, so this will not be merged.

PR #955, opened 2026-09-11
Use float64 by default for BlinkCosine scores

stacyshcherbakova:
BlinkCosine was the only similarity score returning float32, which led to mixed float dtypes when stacking scores in one Scores object. Scores are now computed and returned as float64 by default, consistent with the other cosine scores. A new `dtype` parameter (same pattern as FlashSimilarity) allows opting into float32 to reduce memory usage.


florian-huber (approved):
This looks good. Thanks a lot for the pull request @stacyshcherbakova !

PR #957, opened 2026-09-16
Add `ModifiedCosineLinear` from SIRIUS

LucaCappelletti94:
This PR adds `ModifiedCosineLinear` implemented as a descendant of `CosineLinear`, the linear-time exact modified cosine that originated in SIRIUS.

It requires the same preprocessing of the non-modified variant where it merges close peaks first so that every peak has at most one direct and one precursor-shifted partner, finds both partner sets with a two-pointer sweep each, and then (and here is where the magic happens) **resolves the conflicts between them with a small dynamic program on the resulting paths**.

The bulk of the PR lines are from `tests/modified_cosine_linear_reference.json` fixture, generated from the https://github.com/earth-metabolome-initiative/mass-spectrometry-traits impl, which has been differentially fuzzed against the hungarian variant.


LucaCappelletti94:
The ruff failure seems to be pre-existing, I guess it should be bumped and we should do a PR updating either its rules or the points where it rants about.

florian-huber:
@LucaCappelletti94 ruff rules have been updated (and code has been linted) in the `development` branch. I guess we can skip this effort here for the "old" matchms part.

LucaCappelletti94:
> @LucaCappelletti94 ruff rules have been updated (and code has been linted) in the `development` branch. I guess we can skip this effort here for the "old" matchms part.

Should I therefore refactor and point the PR to "development"?

LucaCappelletti94:
Rebased

LucaCappelletti94:
It appears that the `make doctest` is red both here and on development.

Nearly all of it seems to be `readthedocs/conf.py` setting `inherited-members` to `True` while `MetadataCollection` subclasses `pandas.DataFrame`, so sphinx executes the pandas docstring examples that expect `df` and `pd` bound by earlier pandas docstrings.

Other than those, there are some defects reported in the actual library classes and methods, but none introduced by this PR. I will send a PR to try and address them later on.

florian-huber:
Nice!
And don't worry about the doctest, that will be taken care of elsewhere (we anyway will have to refactor quite some documentation parts for version 1.0).

florian-huber (approved):
Looks very good! Thanks @LucaCappelletti94 

(I might later move the json, and the one from the linear cosine to the testdata folder, but other than that, I think no changed needed!)

